PR #5, opened 2017-02-21
adding definitions

jbkoh:
Definitions mainly come from ASHRAE Dictionary (https://www.ashrae.org/resources--publications/free-resources/ashrae-terminology)
I received the dictionary from ASHRAE and a coordinator confirmed that we can use it in Brick definitions. I currently have the file and only can share it personally (not in this thread) or you can request to ASHRAE directly.

Below is not included in this PR but a discussion point:
I currently add a few essential definitions first, but at scale, I think we need something similar to the structure where Tags consist a Tagset for readability. With a discussion with a few people, I decided to provide definitions to TagSets only (not Tags.) However, providing definitions to all TagSets in texts is not scalable and not helpful.  For example, for human readability, we can tell users to refer "Exhaust Air" and "Temperature" to understand "Exhaust Air Temperature"


jbkoh:
Agreed. I just pushed the notebook to my repo after cleaning it.

gtfierro:
Augmented this PR with a1ad19a9eaac14c560887771e7df1d8b1245ba3d to avoid future issues with storing iPython state

Thanks for the PR!

PR #9, opened 2017-03-24
Changing base URI from buildsys.org -> briskschema.org/schema/1.0

jbkoh:


src/TagSets.csv
```diff
@@ -14,7 +14,7 @@ Equipment>Energy Storage,Thermal Energy Storage,The temporary storage or removal
...
-Equipment>HVAC,Compressor,"(1) device for mechanically increasing the pressure of a gas. (2) often described as being either open, hermetic, or semihermetic to describe how the compressor and motor drive is situated in relation to the gas or vapor being compressed. Types include centrifugal, axial flow, reciprocating, rotary screw, rotary vane, scroll, or diaphragm. 1. device for mechanically increasing the pressure of a gas. 2.�specific machine, with or without accessories, for compressing refrigerant vapor.",ASHRAE Dictionary,,,,,,,,,,,,,,,
+Equipment>HVAC,Compressor,"(1) device for mechanically increasing the pressure of a gas. (2) often described as being either open, hermetic, or semihermetic to describe how the compressor and motor drive is situated in relation to the gas or vapor being compressed. Types include centrifugal, axial flow, reciprocating, rotary screw, rotary vane, scroll, or diaphragm. 1. device for mechanically increasing the pressure of a gas. 2. specific machine, with or without accessories, for compressing refrigerant vapor.",ASHRAE Dictionary,,,,,,,,,,,,,,,
```
gtfierro: Can we more this into a separate PR?

src/TagSets.csv
```diff
@@ -24,7 +24,7 @@ Equipment>HVAC,Fan,"Any device with two or more blades or vanes attached to a ro
...
 Equipment>HVAC,Fume Hood,"A fume-collection device mounted over a work space, table, or shelf and serving to conduct unwanted gases away from the area enclosed.",ASHRAE Dictionary,,,,,,,,,,,,,,,
 Equipment>HVAC,Pump,"Machine for imparting energy to a fluid, causing it to do work, drawing a fluid into itself through an entrance port, and forcing the fluid out through an exhaust port.",ASHRAE Dictionary,,,,,,,,,,,,,,,
-Equipment>HVAC,Space Heater,"A heater used to warm the air in an enclosed area, such as a room or office","space heater. (n.d.) Collins English Dictionary � Complete and Unabridged, 12th Edition 2014. (1991, 1994, 1998, 2000, 2003, 2006, 2007, 2009, 2011, 2014). Retrieved February 20 2017 from http://www.thefreedictionary.com/space+heater",,,,,,,O,X,Jason,,,,,,
+Equipment>HVAC,Space Heater,"A heater used to warm the air in an enclosed area, such as a room or office","space heater. (n.d.) Collins English Dictionary  Complete and Unabridged, 12th Edition 2014. (1991, 1994, 1998, 2000, 2003, 2006, 2007, 2009, 2011, 2014). Retrieved February 20 2017 from http://www.thefreedictionary.com/space+heater",,,,,,,O,X,Jason,,,,,,
```
gtfierro: Can we more this into a separate PR?

jbkoh: What do you mean by a separate PR? With more commits later?

gtfierro:
Looks good.  I added a few comments, but there's also the following before I merge this:

- [ ] Make sure the latest schema files are hosted on the proposed URL (brickschema.org/schema/1.0)
- [ ] Update the representative building files so that they are also using the new URL
- [ ] Fix the typo in the issue title

src/TagSets.csv
gtfierro: These changes are not related to changing the Brick URI, so they should go under a separate pull request

jbkoh:
Actually I need to hold this PR regardless with the above points. Produced TTL files cannot be parsed.

src/TagSets.csv
jbkoh: These are separated by different commits. Should one PR contain one commit?

gtfierro: Nah, its fine to keep them as one. They're both unicode/consistency corrections, right?

jbkoh: Yeah they are both consistency corrections, but I need to close this PR to make some changes more. Will send a discussion to the group. Thanks for the review! Your review will be contained in the coming PR.

jbkoh:
I also think the version should be 1.1 other than 1.0 (as the paper's version is 1.0) How do you think?

gtfierro:
Its not officially in the collaboration docs, but I'd like to use semver if possible. Can you have the version be 1.0.1?

jbkoh:
@gtfierro Hi Gabe, I update it to use 1.0.1. Could you review and accept the PR? I will be able to update the website today too.

gtfierro:
Acceptance of the PR is predicated on the example buildings being updated. It is important that the example buildings reflect correct and up-to-date usage of Brick, otherwise they are not very helpful

PR #38, opened 2018-04-06
Fix the version in the URLs + adding test code.

jbkoh:
The current release version 1.0.2 uses 1.0.1 inside its URLs. I fix it and also parameterize versions in the generation code.

Also added a small test code.


gtfierro:
This looks good. One minor suggestion is maybe the test script should catch the exceptions and print out what the possible cause of the error is? e.g. "Failed: Brick.ttl parse"

jbkoh:
Hi Gabe, thanks for the comments.

Now the test script catches and prints the parsing error. Then it exits as the following tests would be dependent on the schema files.

I will merge later today if there is no further comment.

gtfierro:
Does your script include printouts for other errors?

PR #44, opened 2018-05-07
Thermostat HVAC Mode/Status Points

gtfierro:
Many off-the-shelf thermostats expose these points (mode is typically Off/Heat-only/Cool-only/Auto, state is Off/Heat/Cool). I found it necessary to describe these points in Brick so I'm pushing these back. I was able to describe other common thermostat points using pre-existing classes in Brick


gtfierro:
@jbkoh I wasn't sure if I used your `BuildBrick.py` file correctly so let me know what you think. I have a few more classes I'd like to add in subsequent PRs

jbkoh:
Hi @gtfierro, I see that those are from relatively modern thermostats, which should be common these days. (I remember the thermostat in the CIEE office had this too.) Thanks for bringing this issue.

### Some comments

- I see that you use a new Tag, ``State``. Why do you use it instead of ``Status``? We specifically define ``Status`` as ``State, position, or condition of an item. (adopted from ASHRAE dictionary) Technically, this value should not be arbitrarily written by users or other systems as it is an observation.``. We could reuse it if there is no specific reason. Or even more specifically, we can use ``Mode Status`` as you've already done for ``Mode Command`` as well as other TagSets like ``VAV Operating Mode Status``. (But I will use State in the rest of the review to be compliant with the current status.)

- ``Thermostat HVAC State`` is ambiguous of what it really means in the system. Does Cool-only mean a VAV is cooling, all the associated VAVs are cooling or in the associated VAVs are cooling in average? Can we be more specific here? If it only observes a VAV, we already have ``VAV Operating Mode Status``. However, if the meaning is inherently ambiguous (e.g., the definition of cooling mode may be too complicated to define with a few relationships), I think it's fine to use ``HVAC State`` directly.

- Furthermore, even if we can specify HVAC into VAV or else, what it actually observes is unclear as we do not have Point-Equip relationship for it. We might associate specific VAVs that a thermostat is observing using something like ``measures`` similar to SSN approach, which would be an extension of our ``Resource`` discussion. Even though exact behaviors are modeled by ``controls`` with other points, higher-level relationships might be useful for easier querying.

- Do we need ``Thermostat`` in the name? ``HVAC State`` can be standalone IMHO and we can relate it to a thermostat using ``hasPoint`` relationships. I know we have a convention attaching equipment name as a prefix, but not sure it is useful. This may lead to a new discussion.

- (I said VAVs all over the comments, but it can be anything else of course.)


### Schema Compile
I will add the instruction to run the scripts. Basically, you need to update the ``config.json`` if version update is needed, and execute ``./build.sh``.

Btw, I like the idea of distinguishing two commits into the source commit and generated files commit. I will do that from now on too.

Thanks.

gtfierro:
> I see that you use a new Tag, State. Why do you use it instead of Status?
I'm ok with moving it to Status instead of State. `Mode Status` isn't quite right because `Thermostat HVAC Mode` is the operating policy of the thermostat and `Thermostat HVAC Status` would be what the thermostat is currently doing.

> Thermostat HVAC State is ambiguous of what it really means in the system.
I'm trying to define points used for off-the-shelf thermostats used in RTU-based buildings, so there are no VAVs. Cool-only is a common mode on such thermostats that restricts the thermostat's operation to only cooling or being off (i.e. no heating; there is no minimum temperature setpoint). To me, `HVAC State` is even more ambiguous than `Thermostat State` (which is maybe what we want to use? Or `Thermostat RTU state`?). This I think is relevant to the point you raised about whether or not the equipment name belongs in the point name. Upon consideration I do like the idea of keeping the equipment name out of the point name, but I'd like the point name to be more descriptive: `Thermostat Operation Mode` and `Thermostat Operation Status` perhaps.

Let me know when the instructions are updated. I don't want to bump the version until we have a few more changes in place (I have some more TagSets to push)

jbkoh:
I just added the compile instruction to the [README.md](https://github.com/BuildSysUniformMetadata/Brick).

I guess my question should be how much a thermostat is involved in the control of RTUs. If it is actively involved, Thermostat HVAC Status looks good because it represents what the thermostat is doing. However, in our buildings with VAVs, a thermostat is a collection of some inputs that are used by VAVs, and it alone cannot determine whether a VAV is on or not. In this case, HVAC status is actually the status of the VAV. I guess my confusion comes from this point. If the thermostat in your case can represent the operation mode alone, I agree with either ``Thermostat HVAC Status`` or ``Thersmotat Operation Mode Status``.

jbkoh:
Btw, I perfer having this kind of discussions in Issues first for more reachability.

gtfierro:
@jbkoh I decided on `Thermostat HVAC Mode Command` and `Thermostat HVAC Mode Status` as the points. I re-generated the files and pushed them for approval.

Sorry that we ended up having a conversation in the PR instead of in an RFC. Its definitely better that larger changes happen through RFCs before they are PRs.

jbkoh:
LGTM

PR #48, opened 2018-08-23
Add a test feature + fix a bug like CWS_Chilled_Water_Systems...

jbkoh:
The commit message is self-explanatory for the test feature.

The problem was that Equipment TagSet was added if it is not a part of a Point TagSet even though it "usesEquipment" in the ``TagSets.csv`` file. However, in the check, it did not compare synonyms so it didn't notice "CWS" is already there in "Chilled_Water_Systems_Enable_Command". Thus the solution needs to exploit the synonym information.
  
The current solution is hard-coded because it is difficult to exploit synonyms in the current code structure. I manually created the synonym dictionary inside the code and check. If we want to keep this generation flow, we need to rewrite the code to systematically exploit the CSV file instead of iterating them. If we change our methods to systematic generation, we don't need this fix at that time. Thus, I decided to apply a hot but temporal fix. This does not affect the rest of it, so this should be an okay-solution for now.

This is reported by https://groups.google.com/forum/#!topic/brickschema/vvNw9HgzZh8


jbkoh:
Current TagSet changes:
```
Removed: ['HWS_Enable_Hot_Water_System_Outside_Air_Temperature_Setpoint', 'HWS_Disable_Hot_Water_System_Outside_Air_Temperature_Setpoint', 'DHWS_Domestic_Hot_Water_System_Start_Stop_Command', 'HWS_Hot_Water_System_Enable_Command', 'DHWS_Domestic_Hot_Water_System_Enable_Command', 'CWS_Chilled_Water_System_Enable_Command']
Added: ['HVAC_Mode']
```

I think HVAC_Mode was added in the previous commit by Gabe, but somehow wasn't included in the ttl file at that time.

jbkoh:
@gtfierro any comments?

gtfierro:
lgtm!

jbkoh:
Okay, I will fix it in the next update (if I don't forget...)

Thanks for the review!

PR #56, opened 2019-08-09
introduce brick:Class as the top class

jbkoh:
(also augment gitignore for Python)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,127 @@
 Brick_expanded.ttl
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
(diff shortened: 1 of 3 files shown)


JoelBender:
What is/are the problem/s that this addresses?

jbkoh:
Hi Joel, we removed ``brick:TagSet`` class which was the top class of all the TagSets before. We are replacing it with ``brick:Class``. It would help to retrieve all the Brick classes when there are other information models exist in the same endpoint. For example,
```sparql
select ?class where {
  ?class a/rdfs:subClassOf* brick:Class. 
  # ?class a brick:Class.equivalent to the above with an inference engine.
}
```

gtfierro:
Technically, the query above would just need to have `?class a/rdfs:subClassOf brick:Class` because the `subClassOf` relationship is transitive.

In the future, please separate edits from the `.gitignore` from other edits, but thanks for separating the ontology editing and generation into separate commits

PR #58, opened 2019-08-15
Remove examples from the schema file

jbkoh:
Previously, example entities are hardcoded inside Brick.ttl because generate_brick.py does the generation and testing inference together. I separate the inference testing functionality into ``tests/test_inference.py``. And now we can add more testing functionality into the directory too.


generate_brick.py
```diff
@@ -1,16 +1,13 @@
...
 
-BRICK = Namespace("https://brickschema.org/schema/1.1.0/Brick#")
-TAG = Namespace("https://brickschema.org/schema/1.1.0/BrickTag#")
-BLDG = Namespace("https://brickschema.org/schema/1.1.0/ExampleBuilding#")
-OWL = Namespace("http://www.w3.org/2002/07/owl#")
-RDF = Namespace("http://www.w3.org/1999/02/22-rdf-syntax-ns#")
-RDFS = Namespace("http://www.w3.org/2000/01/rdf-schema#")
+BRICK = Namespace("https://brickschema.org/schema/{0}/Brick#".format(BRICK_VERSION))
+TAG = Namespace("https://brickschema.org/schema/{0}/BrickTag#".format(BRICK_VERSION))
+BLDG = Namespace("https://brickschema.org/schema/{0}/ExampleBuilding#".format(BRICK_VERSION))
```
gtfierro: You can remove the `BLDG` namespace

gtfierro:
Thanks for pulling the test code out of there. Next steps should eventually be to wrap the tests in an actual Python test framework and hook that up to Travis CI

generate_brick.py
jbkoh: done.

jbkoh:
right, I am writing another testing script. Let's collect them inside ``tests/`` dir and then configure a CI framework.

gtfierro:
:shipit:

PR #60, opened 2019-08-25
Fix Tags being a Brick Class in the previous PR

jbkoh:
We introduced brick:Class in https://github.com/BrickSchema/Brick/pull/58. However, it accidentally changed Tags to be Brick Classes too. This PR fixes it.


gtfierro:
I think the fix is good, but maybe a better implementation would be to take `Tag` out of the `roots` dictionary in `generate_brick.py` and declare it separately to be an `owl:Class`. That way, we don't have the special case inside the `define_rootclasses` method. What do you think?

gtfierro:
LGTM! :shipit:

PR #62, opened 2019-08-26
Fix a bug: non-substances being Substances

jbkoh:
- Bug1: Remove Luminance and Enthalpy being subclasses of Substance, which are Quantities.
- Bug2: Wrong tags as ``brick"Location brick:hasTag tag:Loc`` as ``brick:Equipment brick:hasTag tag:Equip``.

### DIFF
--- a/Brick.ttl
+++ b/Brick.ttl
@@ -439,11 +439,6 @@ brick:Chilled_Water_Differential_Pressure_Sensor a owl:Class ;
     rdfs:label "Chilled Water Differential Pressure Sensor" ;
     rdfs:subClassOf brick:Differential_Pressure_Sensor ;
     owl:equivalentClass [ owl:intersectionOf ( [ a owl:Restriction ;
-                        owl:hasValue brick:Pressure ;
-                        owl:onProperty brick:measures ] [ a owl:Restriction ;
-                        owl:hasValue brick:Water ;
-                        owl:onProperty brick:measures ] ) ],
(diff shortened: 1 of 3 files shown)


gtfierro:
Should these be quantities rather than substances?

jbkoh:
Luminance is a Quantity and Light would be a Substance
(https://en.wikipedia.org/wiki/Luminance)
Enthalpy as well; "Enthapy is a property of thermodynamic system" (https://en.wikipedia.org/wiki/Enthalpy)

You defined them in both substances.py and quantities.py.

gtfierro:
Ok, lets move them into quantities, and then introduce Light as a substance.

---

Changing `Loc` and `Equip` tags should go into its own PR.

jbkoh:
That's a separate PR but didn't file it. I will submit that first and then accept this.

jbkoh:
It's hard to separate them out for now. I just added the change in the PR request too.

gtfierro:
Its not that hard to separate them out. If you're not comfortable editing the branches in git, you can just create two new branches and copy/paste the changes there. Its important to keep these changes isolated

jbkoh:
I created a separate branch for the later one but because the commit history follows, it shows up in this PR anyway. Otherwise, I have to create new commits separately for those two, which just takes time. I will do it when I find the time, if you don't like.

gtfierro:
Looks great! Thanks for taking the time to separate the two changes

jbkoh:
You are welcome. Thanks for the suggestions!

PR #64, opened 2019-08-28
Fix wrong tags (Loc and Equip)

jbkoh:
A bug fix: Loc and Equip -> Location and Equipment for Tags


jbkoh:
You are right. I only checked Loc. I just updated Equip to Equipment as well.

PR #65, opened 2019-08-28
Add a new test script to check if the tags inference is correct in the hierarchy

jbkoh:
This tests whether all the classes tagging rules are correctly defined or not. If there is any entity associated with Tags associated for a Class, the reasoner should be able to identify the entity is an instance of all the superclasses of the Class too but nothing else.

### The basic logic
To do that, I [instantiate entities](https://github.com/jbkoh/Brick/blob/new-hierarchytest/tests/test_hierarchy_inference.py#L73) (named as class_name + ``_0``) and [associate them with the Tags](https://github.com/jbkoh/Brick/blob/new-hierarchytest/tests/test_hierarchy_inference.py#L74) [defined as restrictions for the classes](https://github.com/jbkoh/Brick/blob/new-hierarchytest/tests/test_hierarchy_inference.py#L54). Then, I run [owlrl reasoner](https://github.com/jbkoh/Brick/blob/new-hierarchytest/tests/test_hierarchy_inference.py#L77) and query what classes are inferred for the instances. [A naive query to get all the superclasses](https://github.com/jbkoh/Brick/blob/new-hierarchytest/tests/test_hierarchy_inference.py#L102) are used to extract "true parents', which [are compared to](https://github.com/jbkoh/Brick/blob/new-hierarchytest/tests/test_hierarchy_inference.py#L114) the inferred parents.
### An example:
- Inference result: https://drive.google.com/open?id=1lrBiCQS3O8RnoGaaVI8Spzbq3_wgDEGs
- Test result: https://drive.google.com/open?id=18Wi-eYnBM51DCs4u793CK8RbLpCnfddZ

As described in the README, reasoning takes a long time and you may reuse an inferred result as explained there. In the above example, one has to copy paste Inference result into the right location as in README.

This shows an interesting result, for which I will create a separate issue, but please take a look at the code in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,3 +220,9 @@ define_subclasses(definitions, BRICK.Equipment)
 ```
 
 For now, the code is the documentation. Look at `equipment.py`, `point.py`, etc for examples and how to add to each of the class hierarchies.
+
+
+## How to Test?
+1. ``python tests/test_inference.py``: A basic test if Classes can be inferred from associated Tags.
+2. ``python tests/test_hierarchy_inference.py``: It tests whether all the instances associated with the right tags are correctly inferred through the hierarchy. For example, an instance with Tags of ``tag:Temperature`` and ``tag:Sensor`` should be inferred as ``brick:Temperature_Sensor``, ``brick:Sensor``, and ``brick:Point``.
+    - This test involves a time-consuming reasoning process, which took 4.5 hours in my desktop. In case you have the result graph at ``tests/test_hierarchy_inference.ttl``, you can reuse the file with an option like ``python tests/test_hierarchy_inference.py --reuse-inference``.
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # Brick
 
+[![Build Status](https://travis-ci.org/BrickSchema/Brick.svg?branch=master)](https://travis-ci.org/BrickSchema/Brick)
+
 Brick is an open-source, BSD-licensed development effort to create a uniform schema for representing metadata in buildings. Brick has three components:
 
 * An RDF class hierarchy describing the various building subsystems and the entities and equipment therein
--- a/README.md
+++ b/README.md
@@ -220,9 +220,3 @@ define_subclasses(definitions, BRICK.Equipment)
 ```
 
 For now, the code is the documentation. Look at `equipment.py`, `point.py`, etc for examples and how to add to each of the class hierarchies.
-
-
-## How to Test?
-1. ``python tests/test_inference.py``: A basic test if Classes can be inferred from associated Tags.
-2. ``python tests/test_hierarchy_inference.py``: It tests whether all the instances associated with the right tags are correctly inferred through the hierarchy. For example, an instance with Tags of ``tag:Temperature`` and ``tag:Sensor`` should be inferred as ``brick:Temperature_Sensor``, ``brick:Sensor``, and ``brick:Point``.
-    - This test involves a time-consuming reasoning process, which took 4.5 hours in my desktop. In case you have the result graph at ``tests/test_hierarchy_inference.ttl``, you can reuse the file with an option like ``python tests/test_hierarchy_inference.py --reuse-inference``.
--- a/README.md
+++ b/README.md
@@ -26,6 +26,13 @@ If you have an issue with Brick's coverage, utility or usability, or any other B
 
 See [CONTRIBUTING.md](https://github.com/BrickSchema/Brick/blob/master/CONTRIBUTING.md)
 
+## Tests
+
+Tests go in the `tests/` directory and should be implemented using [pytest](https://pytest.readthedocs.io/en/latest/getting-started.html#getstarted).
+[`tests/test_inference.py`](https://github.com/BrickSchema/Brick/blob/master/tests/test_inference.py) is a good example.
+
+Run tests by executing `pytest` in the top-level directory of this repository.
+
 ## Ontology Implementation
 
 ### Complexity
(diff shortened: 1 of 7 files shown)


tests/test_hierarchy_inference.py
```diff
@@ -0,0 +1,130 @@
...
+    ?class rdfs:subClassOf+ brick:Class.
+}
+"""
+
+if args.reuse_inference:
+    expanded_g = Graph()
+    expanded_g.parse(inference_file, format='turtle')
+else:
+    # create instances and associate them with related Tags.
+    qstr = q_prefix + """
```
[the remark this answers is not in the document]
jbkoh: @gtfierro I augmented the comments in the recent commits. Please check.

README.md
```diff
@@ -211,3 +211,9 @@ define_subclasses(definitions, BRICK.Equipment)
 ```
 
 For now, the code is the documentation. Look at `equipment.py`, `point.py`, etc for examples and how to add to each of the class hierarchies.
+
+
+## How to Test?
+1. ``python tests/test_inference.py``: A basic test if Classes can be inferred from associated Tags.
+2. ``python tests/test_hierarchy_inference.py``: It tests whether all the instances associated with the right tags are correctly inferred through the hierarchy. For example, an instance with Tags of ``tag:Temperature`` and ``tag:Sensor`` should be inferred as ``brick:Temperature_Sensor``, ``brick:Sensor``, and ``brick:Point``.
+    - This test involves a time-consuming reasoning process, which took 4.5 hours in my desktop. In case you have the result graph at ``tests/test_hierarchy_inference.ttl``, you can reuse the file with an option like ``python tests/test_hierarchy_inference.py --reuse-inference``.
```
gtfierro: Can you try using `owlready2` as the reasoner? Its much faster and shouldn't need the flag for re-using the inference

tests/test_hierarchy_inference.py
```diff
@@ -0,0 +1,149 @@
...
+TAG = Namespace("https://brickschema.org/schema/{0}/BrickTag#".format(BRICK_VERSION))
+SKOS = Namespace("http://www.w3.org/2004/02/skos/core#")
+entity_postfix = '_0'
+
+q_prefix = """
+prefix brick: <https://brickschema.org/schema/1.1.0/Brick#>
+prefix owl: <http://www.w3.org/2002/07/owl#>
+"""
+
+<<<<<<< HEAD
```
gtfierro: Merge conflict residue here

README.md
gtfierro: I think these instructions are misleading. Tests are run using `pytest`, not by executing the scripts directly. If we want to have a list of tests, we can have a subsection under `## Tests` in the file.

tests/test_hierarchy_inference.py
```diff
@@ -0,0 +1,149 @@
+import argparse
+import json
+from copy import deepcopy
+from collections import defaultdict
+
+from tqdm import tqdm
+import arrow
```
gtfierro: Can we use the regular `time` library from stdlib? We're not using any special features of arrow

tests/test_hierarchy_inference.py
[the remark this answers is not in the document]
gtfierro: Looks a lot better, thanks!

README.md
jbkoh: - Yeah, the README file should be updated. I will fix it soon.
- I tested owlready2 but the reasoning completeness is different than owlrl's. It wasn't able to properly infer superclasses. I think owlready2 is faster because its reasoning logic is simpler.

jbkoh: - Updated README file.
- Please verify my observation of the difference between owlready2 and owlrl.

gtfierro: When the owlrl inference finishes, I'll let you know :)

I can't find a good definition of the differences between them, but I've been developing using owlready2 and its worked great for me so far. Do you have an example of inference that it doesn't perform?

gtfierro:
Looks like you're missing an `import time` in your test file as well

jbkoh:
> Looks like you're missing an `import time` in your test file as well

sorry about that :( I just fixed it.

README.md
jbkoh: owlready2 cannot infer an entity's parent classes. For example, given
```turtle
ex:space_0    brick:hasTag    tag:Space.
ex:space_0    brick:hasTag    tag:Location.
```
, owlready2 can infer like

```turtle
ex:space_0    a    brick:Space.
```
but not 
```turtle
ex:space_0    a    brick:Space, brick:Location.
```

Basically, I need RDFS inference in the test.

gtfierro: Ok, I can reproduce this. I agree we need to use owlrl to do the inference properly for now. It makes the build time longer, which is unfortunate. To mitigate that, I was thinking that we commit the fully materialized ontology after running it through the reasoner.

This doesn't actually save us time though, because we'll ultimately have to run it for each of the tests independently (requiring different runs of the reasoner) *or* write the tests so that they can all operate on the same graph (which makes tests harder to write).

instead, if we organize test files by tests that operate on the same graph -- as in, the set of triples added at the beginning of the test file -- then we can somewhat amortize the cost of running the reasoner.

What do you think?

gtfierro:
It looks like there's a stray `pbd.trace()` in the file as well. I think that's the last error we can fix. The build will fail when you push your last change, but that's because Brick is failing tests

jbkoh:
Fixed. There were other strays. I don't know why I pushed them... Thanks for all the discussions and the review. I will merge this one and fix the schema with you.

gtfierro:
Thanks for merging, but in the future, can we wait to merge until we've actually verified that the travis server can correctly run the tests? There is still a `time.get()` in there, so the tests are broken. We also had a Changes Requested marked on this PR which still has not been resolved. In the future, I think it would benefit the project if we can resolve all of the conversations before merging the PR

PR #70, opened 2019-09-06
Fix tests

gtfierro:
- use of non-existent `time.get()` method
- default to owlrl reasoner in `test_inference.py`, which will probably require fixing the tests to match the actual results


jbkoh:
it's super weird as I fixed this one and pushed at some point. Sorry I mixed it up and thanks for the fix.

gtfierro:
Again, please do not merge in the fixes until we can verify that they are done. Did you notice that the build failed because TravisCI timed out? This branch was not ready to be merged

PR #72, opened 2019-09-06
Re-doing the fix-tests merge

gtfierro:
TODOs:

- [x] get travis to work for more than 10 minutes (needs constant output)
- [x] verify that tests are correct, even if they fail

### DIFF
--- a/tests/test_no_inference.py
+++ b/tests/test_no_inference.py
@@ -15,78 +15,78 @@ DCTERMS = Namespace("http://purl.org/dc/terms#")
 SDO = Namespace("http://schema.org#")
 A = RDF.type
 
-G = rdflib.Graph()
-G.parse('Brick.ttl', format='turtle')
-G.bind('rdf', RDF)
-G.bind('owl', OWL)
-G.bind('rdfs', RDFS)
-G.bind('skos', SKOS)
-G.bind('brick', BRICK)
-G.bind('tag', TAG)
-G.bind('bldg', BLDG)
+g = rdflib.Graph()
+g.parse('Brick.ttl', format='turtle')
+g.bind('rdf', RDF)
+g.bind('owl', OWL)
+g.bind('rdfs', RDFS)
+g.bind('skos', SKOS)
+g.bind('brick', BRICK)
+g.bind('tag', TAG)
+g.bind('bldg', BLDG)
 
 # Create instances
-G.add( (BLDG.Coil_1, A, BRICK.Heating_Coil) )
-G.add( (BLDG.AHU1, A, BRICK.Air_Handler_Unit) )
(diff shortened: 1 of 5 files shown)


jbkoh:
Hi Gabe, I think this is a good approach. Is it ready for a review?

gtfierro:
The TODO items on this issue aren't finished yet, so its not ready for review

gtfierro:
This is getting ready to merge.

I've added a [new reasoner method](https://github.com/BrickSchema/Brick/blob/7e567f6a0ab5e86215775b75dd7b991636c7f13f/tests/util/reasoner.py#L67-L132) for the tests to use. It does the inverse edges, tag + substance inference and then finishes off with `owlrl`'s RDFS inference. The result is much faster than full OWLRL inference (about 1.5 minutes on my machine) and so far seems to produce pretty good results.

Obviously, we want to make sure its fully OWLRL compliant eventually, but this might be enough for us to get the automated testing started until we get a more performant reasoner.

@jbkoh thoughts? I still have some cleaning up to do on the code so please don't merge this yet.

gtfierro:
According to Travis, only the `test_hierarchical_inference` test is failing now, which we expected. I'll merge this on approval from @jbkoh

jbkoh:
Sorry I didn't pay attention to this thread. I will take a look now.

jbkoh:
https://github.com/BrickSchema/Brick/blob/7e567f6a0ab5e86215775b75dd7b991636c7f13f/tests/util/reasoner.py#L43
infer inverse relationships can be done simple with a single query.
```
query = """
INSERT {
?o ?invprop ?prop
} WHERE {
?s ?prop ?o.
?prop owl:inverseOf ?invprop.
}
"""
```

tests/test_no_inference.py
```diff
@@ -0,0 +1,105 @@
...
+
+BRICK = Namespace("https://brickschema.org/schema/{0}/Brick#".format(BRICK_VERSION))
+TAG = Namespace("https://brickschema.org/schema/{0}/BrickTag#".format(BRICK_VERSION))
+BLDG = Namespace("https://brickschema.org/schema/{0}/ExampleBuilding#".format(BRICK_VERSION))
+SKOS = Namespace("http://www.w3.org/2004/02/skos/core#")
+DCTERMS = Namespace("http://purl.org/dc/terms#")
+SDO = Namespace("http://schema.org#")
+A = RDF.type
+
+G = rdflib.Graph()
```
jbkoh: Can we standardize the variable name between ``G`` and ``g``? I prefer ``g`` but don't mind using ``G`` too but let's use just one.

gtfierro: Changed everything to `g`!

gtfierro:
Fixed the inverse edge inference as you suggested. I think we're all good to merge now! Thanks for the feedback

PR #73, opened 2019-09-10
Move Average_Supply_Air_Flow_Sensor

JoelBender:
Class is in the wrong part of the tree.


jbkoh:
Hi Joel, thanks for the fix!

Could you run ``python generate_brick.py`` and push ``Brick.ttl`` in a separate commit too?

JoelBender:
It bothers me that the only difference is this...
```
 brick:Average_Supply_Air_Flow_Sensor a owl:Class ;
     rdfs:label "Average Supply Air Flow Sensor" ;
-    rdfs:subClassOf brick:Discharge_Air_Flow_Sensor ;
+    rdfs:subClassOf brick:Supply_Air_Flow_Sensor ;
```
...the rest of the diff is because there is no canonical order for the restrictions.  But I'm not letting that slow me down! :smile:

Issue #53, opened 2019-06-29
Air is implicit in Location in common usage


jbkoh:
E.g., Outside_Temperature_Sensor, Zone_Temperature_Sensor.

Even though it would be explicit if they have "Air" as a tag, people will commonly miss it as Location tags commonly involve Air when they are associated with Points. I would like to investigate if it's possible to have conditional rules. I.e., When Zone && Sensor, Air is inferred.

This is not an urgent issue though.

gtfierro:
@jbkoh  What do you think about removing the `Zone_Temperature_Sensor` class, and introducing `Zone_Air_Temperature_Sensor` as a subclass of `Air_Temperature_Sensor`? We already don't have an `Outside_Temperature_Sensor`

jbkoh:
Sorry that I didn't reply to this one before. Your suggestion is the most straightforward solution but I would like to discuss if there is any other solution here before concluding. The concept of Zone being measured seems prevailing, and people would use it without ``Air`` significantly including the existing systems.

I think there are two cases to solve this ``Zone_Temperature_Sensor`` (a.k.a. ZNT below) problem.
1. We can remove it and just do nothing. We could prevent users from using ZNT with a checker. The tradeoff is usability.
2. We can remove it but provide an inference rule for those externally. E.g., Zone + Temperature + Sensor -> Zone_Air_Temperature_Sensor outside the standard but as a helper function.
3. Include it with ``brick:measures tag:Location``, assuming ``Location`` involves ``Air``. The risk is its actual substance is implicit in the definition of ``Location``.

I understand 1. would be the most unambiguous. But what would be the risk of 3.?

gtfierro:
"Zone" temperature makes it sound like the whole zone's temperature is being measured, which isn't necessarily true. I think Zone Temperature Sensor colloquially refers to a temperature sensor in a room, and not a supply/discharge air temperature sensor.

I like removing it the best, but if we provide a Zone+Temperature+Sensor tag mapping, then we might as well include "Air" there. Future tooling can suggest additional tags that take your given tag set and turn it into one that Brick recognizes.

If we include a `brick:measures tag:Zone` or `tag:Room`, that might be interesting, but we already have a relationship for location. Might be more straightforward to use `brick:hasLocation ?x . ?x rdf:type brick:Location` as a restriction rather than `brick:Measures`. 

I still think removing the class is the best approach for now

jbkoh:
[...]

PR #75, opened 2019-09-14
Sensor class restructuring

gtfierro:
Addressing issues raised in #53 . 

Changes so far

- moved `Zone_Temperature_Sensor` under `Zone_Air_Temperature_Sensor` and moved/renamed the related classes (commit https://github.com/BrickSchema/Brick/commit/2d06e5d68bc3bbb41c01e38f76d421a430e481fb)
- moved `Building_Static_Pressure_{Sensor, Setpoint}` to `Building_Air_Static_Pressure_{Sensor, Setpoint}`
- moved `Zone_Humidity_Sensor` to `Zone_Air_Humidity_Sensor`

### DIFF
--- a/Brick.ttl
+++ b/Brick.ttl
@@ -228,22 +228,13 @@ brick:Building a owl:Class ;
                         owl:hasValue tag:Site ;
                         owl:onProperty brick:hasTag ] ) ] .
 
-brick:Building_Meter a owl:Class ;
-    rdfs:label "Building Meter" ;
-    rdfs:subClassOf brick:Meter ;
-    owl:equivalentClass [ owl:intersectionOf ( [ a owl:Restriction ;
-                        owl:hasValue tag:Meter ;
-                        owl:onProperty brick:hasTag ] [ a owl:Restriction ;
-                        owl:hasValue tag:Equipment ;
-                        owl:onProperty brick:hasTag ] [ a owl:Restriction ;
-                        owl:hasValue tag:Building ;
-                        owl:onProperty brick:hasTag ] ) ] .
-
-brick:Building_Static_Pressure_Sensor a owl:Class ;
-    rdfs:label "Building Static Pressure Sensor" ;
+brick:Building_Air_Static_Pressure_Sensor a owl:Class ;
+    rdfs:label "Building Air Static Pressure Sensor" ;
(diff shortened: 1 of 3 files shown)


jbkoh:
I think ``Building_Meter`` is the only one left ambiguous. It's Equipment but the same logic applies. Shall we introduce ``Building_Electricity_Meter`` and ``Buliding_Thermal_Meter``?  I am not sure if there are general ``Meters`` that can measure both electrical and thermal energy. This is a low priority compared to Point definitions, but worth considering. Other than this, this PR looks good to me.

mbkj:
A district heating meter measures several quantities:. Temperature and flow. E.g. https://www.kamstrup.com/en-en/heat-solutions/heat-meters/multical-803

Also more advanced electricity meters can measure multiple aspects of electricity supplied. I do not know if this is relevant to consider for this discussion.

jbkoh:
@mbkj It's good to know that. The discussion was to reduce the confusion of what "building" means in the class name ``Building_Meter``. The naming convention we have adopted for sensor names is substance + property + point type while location + property + point type is confusing (e.g., Zone_Temperature_Sensor indicates it measures ``Air_Temperature`` of the zone, which is not explicit in the naming. We addressed that in this PR.) 

In a similar manner, ``Building_Meter`` is ambiguous about what it meters. Your examples comply with the current observation. I.e., advanced electricity meters can measure the multiple aspects of "electricity". A district heating meter measures different aspects of "heat". So I want to propose we introduce ``Building_Eletricity_Meter`` and ``Building_Thermal_Meter`` individually to explain your examples and remove ``Building_Meter``.

jbkoh:
(@mbkj Note that we have moved ``Meter`` to ``Equipment`` from ``Point`` long time ago so that it can be associated with multiple data points as you exemplified.)

gtfierro:
I think `Building_Meter` is important to keep as the superclass of all meters that measure substances at the building level

gtfierro:
@jbkoh when you get a chance can you verify that this PR fixes the issues you raised in https://github.com/BrickSchema/Brick/issues/53 ?

jbkoh:
I am fine with all the changes. I wanted to listen to more people about this but think we have no chance for that soon. 

Please merge this PR. I will close the associated Issue. Thanks!

Issue #74, opened 2019-09-13
Inference with measures for Classes.


jbkoh:
I just noticed that ``brick:measures`` can determine Classes alone. For example,
```turtle
brick:Building_Static_Pressure_Sensor a owl:Class ;
    rdfs:label "Building Static Pressure Sensor" ;
    rdfs:subClassOf brick:Static_Pressure_Sensor ;
    owl:equivalentClass [ owl:intersectionOf ( [ a owl:Restriction ;
                        owl:hasValue brick:Pressure ;
                        owl:onProperty brick:measures ] [ a owl:Restriction ;
                        owl:hasValue brick:Air ;
                        owl:onProperty brick:measures ] ) ],
        [ owl:intersectionOf ( [ a owl:Restriction ;
                        owl:hasValue tag:Building ;
                        owl:onProperty brick:hasTag ] [ a owl:Restriction ;
                        owl:hasValue tag:Static ;
                        owl:onProperty brick:hasTag ] [ a owl:Restriction ;
                        owl:hasValue tag:Pressure ;
                        owl:onProperty brick:hasTag ] [ a owl:Restriction ;
                        owl:hasValue tag:Sensor ;
                        owl:onProperty brick:hasTag ] ) ] .
```
Here, there are two rules: with ``measures`` and ``hasTag``. An inference engine could infer classes only with ``measures``. However, there are many sensors measuring Air and Pressure other than this.

There can be two solutions

1. Define what can be measured throughout the Sensor hierarchy. For example, ``Building_Air`` can be a class that ``Building_Static_Pressure_Sensor`` measures, and then something measuring ``Building_Air`` and ``Pressure`` will be unique. This involves more complete design of "Quantity" and "Substance".

2. A simpler solution that works with our current design would be using ``rdfs:subClassOf`` for ``measures`` Restrictions instead of ``owl:equivalentClass``.

JoelBender:
`Static_Pressure` is an [observable property](https://www.w3.org/TR/vocab-ssn/#SOSAObservableProperty) of air, so a `Building_Static_Pressure_Sensor` is a [sensor](https://www.w3.org/TR/vocab-ssn/#SOSASensor) of `Static_Pressure` associated with a `Building`.  I don't see a `Building_Air` definition, but it would seem that it would be narrower concept of `Air` (purposefully avoiding the subclass/superclass concepts).

[...]

Issue #61, opened 2019-08-26
Subclasses of Quantity are instances of Quantity too.


jbkoh:
After OWL reasoning, Brick.ttl assumes that all the subclasses of ``brick:Quantity`` are instances of ``brick:Quantity`` too. Please check the regenerator code: https://gist.github.com/jbkoh/37fbec75d023eff23baeee0e7c721dd0
Quantity subclasses should not be instances, right?

I am also not sure if we want to instantiate ``brick:Quantity`` as entities. I see potential usage such as provenance tracking, but will it be actually used?

gtfierro:
Read through the section on Substances in the README: https://github.com/BrickSchema/Brick#substances we are punning substances by treating them as singletons. I think this makes the data model more intuitive. The plan is not to instantiate substances beyond the punned instances

jbkoh:
I understand that's for substances but my observation is for Quantities.

gtfierro:
Sorry! I got mixed up. Its the same principle though

jbkoh:
I see, but I found that punning isn't applied to Substances. Can you verify that? The same query does not work for Substances. 

And could you point out which part of the code enables the punning? I just want to understand how it's implemented.

gtfierro:
It's a little weird  because we don't actually instantiate them anywhere, but I think I understand what's going on.

Punning is applied to both Quantities and Substances; the "pun" part is where we say that the value of a certain property is a single instance which is has the same name as the class. Loosely, this is lines https://github.com/BrickSchema/Brick/blob/master/generate_brick.py#L83-L84 instantiated by lines like https://github.com/BrickSchema/Brick/blob/master/sensor.py#L14. When this is serialized into the turtle file, it looks like https://github.com/BrickSchema/Brick/blob/master/Brick.ttl#L252-L254:

When we use a Quantity or Substance in this way as part of a definition, it becomes implicitly "instantiated" . So when you are doing your query above, because you are using `a`/`rdf:type` you are only getting the quantities and substances that have been used in an `owl:Restriction` clause with `owl:hasValue`. Eventually, when all of the substances and quantities are used in class restrictions, then they should all show up in your query.

*However* I would expect some of the substance classes to show up when doing that query. I'm not 100% sure why that's happening, but it smells like there's a bug somewhere. I'll investigate tomorrow

jbulow:
I assume this reaction from owlready2 when loading the Brick ontology is related:

`
[...]

Issue #59, opened 2019-08-20
Align Quantity classes with SSN's Observable Property


gtfierro:

gtfierro:
Addressing this in https://github.com/BrickSchema/Brick/pull/76

PR #76, opened 2019-09-16
Fixing Quantity + Substance

gtfierro:
~~**In progress**~~

Rectifying some of the issues we have uncovered in the modeling of quantities and substances

- quantities + substance classes are instantiated/punned as part of the Brick ontology generation script
- added a test file to verify this is done correctly
- changed the `rdfs:range` property on `brick:measures` because this was improperly inferring some quantities as substances. It now references a new `brick:Measurable` class which groups the Substance and Quantity subclasses
- aligned `brick:Quantity` with `sosa:ObservableProperty` using a subclass relationship
- aligned `brick:Substance` with `sosa:FeatureOfInterest` using subclass relationship
- improve the coverage of `brick:measures` relationships in `sensor` hierarchy
- add tests to verify that all objects of `brick:measures` are defined `Measurable`s

---

**09/24/2019: More major changes**: in the process of performing the above changes, I found an opportunity to help clear up some other issues with Brick while I was at it. These changes are:

- adding a "parent" component to the Python framework to enable "up-pointers" in the class lattice. This helps avoid duplication in the specification
- **important**: removed the `TAG.Setpoint` tag from all `Parameter` classes. Even though `Setpoint` is part of terms like `Max Air Flow Setpoint Limit`, the constraints of open-world set-based formalisms like OWL mean that it is difficult to enforce a class-based separation between setpoints and parameters *without introducing other constraints*. I will get around to writing up those tradeoffs at some later point, but in the mean time this is -- I believe -- an acceptable compromise.

### DIFF
--- a/equipment.py
+++ b/equipment.py
@@ -10,7 +10,6 @@ Set up subclasses of the equipment superclass
 equipment_subclasses = {
     "HVAC": {
         OWL.equivalentClass: "Heating_Ventilation_Air_Conditioning_System",
-        "tags": [TAG.HVAC],
     },
     "Heating_Ventilation_Air_Conditioning_System": {
         OWL.equivalentClass: "HVAC",
@@ -216,21 +215,13 @@ hvac_subclasses = {
     },
     "Coil": {
         SKOS.definition: Literal("Exchanger that transfers heat from an exhaust airstream to a separated supply airstream."),
-        "tagvalues": [
-            [ BRICK.hasTag, TAG.Coil ],
-        ],
+                "tags": [TAG.Equipment, TAG.Coil],
         "subclasses": {
             "Cooling_Coil": {
-                "tagvalues": [
-                    [ BRICK.hasTag, TAG.Coil ],
-                    [ BRICK.hasTag, TAG.Cool ],
-                ]
+                "tags": [TAG.Equipment, TAG.Coil, TAG.Cool],
             },
             "Heating_Coil": {
-                "tagvalues": [
-                    [ BRICK.hasTag, TAG.Coil ],
-                    [ BRICK.hasTag, TAG.Heat ],
-                ]
+                "tags": [TAG.Equipment, TAG.Coil, TAG.Heat],
             },
         },
     },
@@ -271,9 +262,11 @@ valve_subclasses = {
             },
             "Domestic_Hot_Water_Valve": {
                 "tags": [TAG.Domestic, TAG.Water, TAG.Hot, TAG.Valve, TAG.Heat, TAG.Equipment],
+                "parents": [BRICK.Hot_Water_System],
             },
             "Preheat_Hot_Water_Valve": {
                 "tags": [TAG.Preheat, TAG.Water, TAG.Hot, TAG.Valve, TAG.Heat, TAG.Equipment],
+                "parents": [BRICK.Hot_Water_System],
             },
         },
         # OWL.equivalentClass: Restriction(BRICK.hasTag, graph=G, allValuesFrom=BRICK.Valve)
--- a/equipment.py
+++ b/equipment.py
@@ -262,11 +262,11 @@ valve_subclasses = {
             },
             "Domestic_Hot_Water_Valve": {
                 "tags": [TAG.Domestic, TAG.Water, TAG.Hot, TAG.Valve, TAG.Heat, TAG.Equipment],
-                "parents": [BRICK.Hot_Water_System],
+                "parents": [BRICK.Hot_Water_System, BRICK.Water_Valve],
             },
             "Preheat_Hot_Water_Valve": {
                 "tags": [TAG.Preheat, TAG.Water, TAG.Hot, TAG.Valve, TAG.Heat, TAG.Equipment],
-                "parents": [BRICK.Hot_Water_System],
+                "parents": [BRICK.Hot_Water_System, BRICK.Water_Valve],
             },
         },
         # OWL.equivalentClass: Restriction(BRICK.hasTag, graph=G, allValuesFrom=BRICK.Valve)
@@ -276,6 +276,7 @@ valve_subclasses = {
         "subclasses": {
             "Chilled_Water_Valve": {
                 "tags": [TAG.Chilled, TAG.Valve, TAG.Water, TAG.Equipment],
+                "parents": [BRICK.Chilled_Water_System],
             },
         },
     },
--- a/equipment.py
+++ b/equipment.py
@@ -8,9 +8,7 @@ from namespaces import *
 Set up subclasses of the equipment superclass
 """
 equipment_subclasses = {
-    "HVAC": {
-        OWL.equivalentClass: "Heating_Ventilation_Air_Conditioning_System",
-    },
+    "HVAC": {},
     "Heating_Ventilation_Air_Conditioning_System": {
         OWL.equivalentClass: "HVAC",
     },
@@ -95,7 +93,6 @@ hvac_subclasses = {
         # subclasses defined in 'valve_subclasses'
     },
     "VFD": {
-        OWL.equivalentClass: "VFD",
         "subclasses": {
             "Heat_Wheel_VFD": {},
             "Preheat_Valve_VFD": {},
@@ -110,14 +107,17 @@ hvac_subclasses = {
             "Fan_Coil_Unit": {
                 OWL.equivalentClass: "FCU",
             },
+            "FCU": {},
             "Variable_Air_Volume_Box": {
                 OWL.equivalentClass: "VAV",
                 "subclasses": {
                     "Variable_Air_Volume_Box_With_Reheat": {
                         OWL.equivalentClass: "RVAV",
                     },
+                    "RVAV": {},
                 },
             },
+            "VAV": {},
         },
     },
     "Space_Heater": {
@@ -142,6 +142,7 @@ hvac_subclasses = {
             "Condenser_Heat_Exchanger": {},
         },
     },
+    "HX": {},
     "Fume_Hood": {
         SKOS.definition: Literal("A fume-collection device mounted over a work space, table, or shelf and serving to conduct unwanted gases away from the area enclosed."),
     },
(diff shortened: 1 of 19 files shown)


jbkoh:
@gtfierro I know you are still working on this, but I believe ``sosa:ObservableProperty`` has different meaning than ``brick:Measurable`` that you are adding. In our case,
``sosa:ObservableProperty`` == ``brick:Quantity``
``sosa:FeatureOfInterest`` == ``brick:Substance`` (maybe ``brick:Equipment`` too)
Merging ``brick:Quantity`` and ``brick:Substance`` sounds not right to me.

gtfierro:
`brick:Quantity` and `brick:Substance` are not merged. They are still distinct, but are now also annotated as being measurable. This is orthogonal to the alignment with `sosa`. I think a good `sosa` alignment would be

`brick:Quantity rdfs:subClassOf sosa:ObservableProperty`

`brick:Substance rdfs:subClassOf sosa:FeatureOfInterest`

jbkoh:
Yeah, now the alignment looks good. Still, the question is, why do we need ``Measurable`` then? Let me know if you would like to discuss these after you have everything ready.

gtfierro:
We need `Measurable` as a single class that is the range of the `brick:measures` object property.

jbkoh:
Doesn't the current implementation of having two lines for measuers' range work?
```turtle
brick:measures  rdfs:range  brick:Substance;
                rdfs:range brick:Quantity.
```

gtfierro:
It's valid in RDFS, but it means that any object of the `brick:measures` relationship will be inferred as *both* a Substance and a Quantity, which isn't what we want. Its safer to infer the more generic `brick:Measurable` class

gtfierro:
This is ready to merge! Addresses #74, #61, #59 and we are finally passing the test suite :)

@jbkoh let me know what you think and if you have any questions about the implementation or changes made. It's not 100% ideal yet (especially removing `Setpoint` tag from all of the `Parameter` classes), but it lets us move forward.

I also apologize for not separating this into separate PRs. After I fixed everything it became much more difficult to separate the changes out.

status.py
```diff
@@ -177,22 +183,27 @@
...
                         "tags": [ TAG.Standby, TAG.Unit, TAG.On, TAG.Off, TAG.Status ],
+                        "subclasses": {
+                            "Standby_Glycool_Unit_On_Off_Status": {
+                                "tags": [ TAG.Standby, TAG.Glycool, TAG.Unit, TAG.On, TAG.Off, TAG.Status ],
+                            },
+                        },
                     }
                 },
                 "tags": [ TAG.On, TAG.Off, TAG.Status ],
+                "parents": [BRICK.On_Status, BRICK.Off_Status],
```
jbkoh: ``On_Status`` is not defined.

jbkoh: ``On_Status`` is not defined but how did it pass the test?

equipment.py
```diff
@@ -271,9 +262,11 @@
             },
             "Domestic_Hot_Water_Valve": {
                 "tags": [TAG.Domestic, TAG.Water, TAG.Hot, TAG.Valve, TAG.Heat, TAG.Equipment],
+                "parents": [BRICK.Hot_Water_System, BRICK.Water_Valve],
```
jbkoh: I am not sure if ``Hot_Water_System`` is a superclass of everything related to ``Hot_Water``. I know we haven't precisely defined it but I always thought it is a collection of entities for controlling hot water like ``AHU`` for air handling. For example, ``Supply_Fan`` is not a subclass of ``AHU``. I "feel" a ``Valve`` is a part of ``Water_System`` likewise, but  I can't strongly argue it as we don't have the right definition. If you feel that the current hierarchy is correct, we can discuss this later if somebody complains.

sensor.py
```diff
@@ -15,13 +15,16 @@
...
+                        "substances": [ [ BRICK.measures, BRICK.Outside_Air ], [ BRICK.measures, BRICK.Grains ], ],
                     },
                     "Return_Air_Grains_Sensor": {
                         "tags": [ TAG.Return, TAG.Air, TAG.Grains, TAG.Sensor ],
+                        "substances": [ [ BRICK.measures, BRICK.Return_Air ], [ BRICK.measures, BRICK.Grains ], ],
                     }
                 }
             },
             "Angle_Sensor": {
+                #TODO substances
```
jbkoh: Is it still TODO?

sensor.py
```diff
@@ -487,8 +514,24 @@
...
                                 }
                             },
+                            "Supply_Air_Temperature_Sensor": {
+                                "tags": [ TAG.Sensor, TAG.Temperature, TAG.Air, TAG.Supply ],
+                                "substances": [ [ BRICK.measures, BRICK.Temperature ], [ BRICK.measures, BRICK.Supply_Air ], ],
+                                "subclasses": {
+                                    "Cooling_Coil_Supply_Air_Temperature_Sensor": {
+                                        "tags": [ TAG.Cooling, TAG.Coil, TAG.Supply, TAG.Air, TAG.Temperature, TAG.Sensor ],
+                                    },
+                                    "Heat_Wheel_Supply_Air_Temperature_Sensor": {
```
jbkoh: I think we decided to remove equipment tag from point names.

status.py
gtfierro: Oops! Forgot to remove the reference to `On_Status` and `Off_Status`; these passed tests because only the `rdfs:subClassOf` relationship gets added so it doesn't fall under the purview of the existing tests. I'll make sure we catch stuff like this in the future

equipment.py
gtfierro: Fair point! `Supply_Fan` is not a subclass of `AHU`, but both are subclasses of `HVAC` equipment; I feel this is similar. `Hot_Water_Valve` is similarly component of the `Hot_Water_System`. Our class structure is consistent with this pattern for now; if we want to change it in the future, that could be its own pull request

sensor.py
gtfierro: It is still a TODO. I wasn't sure how we wanted to handle concepts like "position" or "angle". Would you say these are quantities? Or substances?

setpoint.py
```diff
@@ -223,6 +240,9 @@
                 "subclasses": {
                     "Max_Load_Setpoint": {
```
jbkoh: Not sure if this is actually a setpoint. It reads like a ``Limit``.

status.py
```diff
@@ -150,9 +153,12 @@
             },
             "Off_Status": {
                 "tags": [ TAG.Off, TAG.Status ],
-            },
-            "On_Status": {
-                "tags": [ TAG.On, TAG.Status ],
+                "subclasses": {
+                    "Emergency_Power_Off_Status": {}, # defined elsewhere
```
jbkoh: I'd rather use "parents" annotation to be consistent with others. Then, we can easily say it is wrong if the class has an empty definition.

status.py
```diff
@@ -227,19 +239,20 @@
                     },
                     "Run_Status": {
                         "tags": [ TAG.Run, TAG.Status ],
+                        "subclasses": {
+                            "Run_Enable_Status": {}, # defined elsewhere
```
jbkoh: Same as one of the previous comments; I'd rather use "parents" annotation for the consistency.

substances.py
```diff
@@ -90,9 +125,32 @@
                                 "tags": [TAG.Liquid, TAG.Water, TAG.Domestic],
                                 SKOS.definition: Literal("Tap water for drinking, washing, cooking, and flushing of toliets"),
                             },
+                            "Entering_Water": {
+                                "tags": [TAG.Liquid, TAG.Water, TAG.Entering],
+                            },
+                            "Leaving_Water": {
+                                "tags": [TAG.Liquid, TAG.Water, TAG.Leaving],
+                            },
+                            "Return_Water": {},
```
jbkoh: Empty definition here.

status.py
gtfierro: This is detected in a new `tests/test_class_structure.py`

sensor.py
gtfierro: Good catch! I removed these plus a few others

setpoint.py
gtfierro: Fixed

status.py
gtfierro: Good point. There's a few of these laying around from before I added the "parents" annotation. I fixed this one but I'll try to find the others as well

sensor.py
jbkoh: I think both angle and position are quantities. Position of a damper, angle of a fan, etc. In these cases, substances are actually equipment, which I think we discussed a bit about whether we need to allow equipment as tags of sensors for such cases. I think it's worth adding them, but we can discuss that after this release too.

jbkoh:
I also did an interesting practice at [OOPS! (OntOlogy Pitfall Scanner!) ](http://oops.linkeddata.es/response-advanced.jsp#)

This checks common pitfalls in ontology design. I would like to include this in the eval process, but there is no API to use it I guess.

Here's the whole result:
[brick-oops-scanner-result.pdf](https://github.com/BrickSchema/Brick/files/3655624/brick-oops-scanner-result.pdf)

I just looked up Critical Results for now but I think the others are interesting too. The critical ones are:
- ``isMeasuredBy`` should have the domain of ``Measurable``.
- ``Start_Stop_Command`` is not defined.
- ``Real_Power`` is not defined.

gtfierro:
That's a very helpful tool! I've fixed the issues you mentioned, but I'll go through the document sometime soon to address the other issues. It would be nice if we could automate this, but it might just need to be a "best practices" for now until we can find an automatable solution.

status.py
gtfierro: Fixed!

substances.py
gtfierro: Fixed

sensor.py
gtfierro: Ok cool; I added Angle and Position and associated them with Angle Sensor and Damper Position Sensor. Agreed that this should be a future conversation as well!

gtfierro:
We are green across the board, at last! @jbkoh you ok with me merging this? Thanks for all your comments and feedback on this

jbkoh:
LGTM!

Issue #29, opened 2017-08-16
Use of Luminance


blip2:
Typically lighting systems within buildings are designed, specified and installed based on illuminance levels (measured in lux). The use of luminance for a number of classes (particularly for LightingSystem) while technically still a valid measurement, not actually the correct terminology what what happens in practice.

jbkoh:
In-depth discussion continues at Brick Google Groups. The discussion will be summarized here once it's finalized there:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/brickschema/crnEA450VJ8/ZTdORh1UCAAJ

Any comments and discussions are welcome here too.

jbkoh:
Summary of the discussions so far. 

1. Introduce following TagSets: (I have aligned the style to other existing Brick tagsets.)
Equipment | Lighting
Equipment>Lighting | Luminaire
Equipment>Lighting>Luminaire | Maintained Emergency Luminaire
Equipment>Lighting>Luminaire | Nonmaintained Emergency Luminaire
Equipment>Lighting | Luminaire Driver
Point>Sensor | Absence Sensor
Point>Sensor | Illuminance Sensor
Point>Sensor > Illuminance Sensor | Photocell Illuminance Sensor

2. Introduce following TagSets (need discussions for proper hierarchy.)
Equipment> Switch
Equipment> Lighting | Dimmer
Equipment> Control Module
Equipment | Touchpanel

3. Possible relationship
Luminaire feeds (or feedsLight) Lighting_Zone

4. Do we need Luminance Sensor? (different from Illuminance Sensor)

5. Introduce a common concept: Interface
Dimmer, Scene Set, Touch Panel in Lighting Systems,
Thermostat in HVAC.

6. Introduce a common concept: Control Module.
Control Module in Lighting System
NAE or others in HVAC.

pisuke:
I'd like to get this discussion higher in the list of issues to be addressed.
Luminance sensors are not needed (the term is wrong). 
All the rest of changes is appropriate. 
I'm happy to do the changes. What is the appropriate way: fork and do a pull request?

gtfierro:
Yep! @pisuke just fork, make the changes in the `src/` directory to `TagSets.csv`, and then run `build.sh` which will generate the turtle files. Please separate the edits to `TagSets.csv` and the execution of `build.sh` into separate commits (they can be more than 1 commit each, but please don't combine the two steps into a single commit).

[...]

PR #77, opened 2019-09-23
Add Illuminance_Sensor

jbkoh:
- Adding ``Illuminance Sensor``. This seems to be the minimum required point for lighting system based on #29. 
- Fixing ``Outside_Luminance_Sensor`` to ``Outside_Illuminance_Sensor``.

For luminance vs illuminance Please refer https://sensing.konicaminolta.us/blog/luminance-vs-illuminance/ 

I would like to remove ``Luminance_Sensor`` as it's confusing but not commonly used. A user already pointed it out before https://github.com/BrickSchema/Brick/issues/29#issuecomment-395724673. I haven't removed in this PR, but let me know if you have any thoughts.

Other than this, we could defer detailed lighting system modeling for next versions.


gtfierro:
@jbkoh Sorry we didn't get to this before the build broke. Did you have a separate branch where you were working on lighting equipment? I remember you mentioning that on a previous call

jbkoh:
Summary:
- Add `Illuminance_Sensor`
- Remove `Luminance_Sensor`

More discussions for "future dev"
- `Luminance_Setpoint` and `Luminance_Command`: Do they make sense?
- `Luminous_Flux` and `Luminous_Intensity`: I see that these are technical terms but where are they used?
- I guess merging this discussion with the Haystac/ANSI C137 discussion would be helpful hereafter.

jbkoh:
I will merge this one unless I see any objection in the next few days.

PR #78, opened 2019-09-24
Add Brick example file: executable tutorial with detailed comments

gtfierro:
Put this together for a community member and thought it might be more generally useful. Any thoughts?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,10 +7,7 @@ install:
   - pip install -r requirements.txt
 
 script:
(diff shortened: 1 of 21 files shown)


jbkoh:
Sorry I missed this one before. The example looks good. We can migrate the other examples from the previous BuildSys tutorial repo to here at some point too.

Can you add a brief description of the example in the root ``README.md``?

Once it's done, I think we can just add this one.

PR #79, opened 2019-09-26
Restructure codebase

jbkoh:
I would like to organize the entire dir a bit because currently there are too many Python files exposed to the root dir, which could confuse firstcomers.

I will update the ``README.md`` and ``CONTRIBUTING.md`` if all think that this structure is fine.


jbkoh:
@gtfierro Done. Please check if there are other places I need to update the file paths. Thanks.

gtfierro:
Go ahead and merge it once the tests pass

PR #80, opened 2019-09-27
Compare versions

jbkoh:
- I add a script to visualize new updates in terms of classes. The current comparison logic is naive but should be useful for curious people. 

- I try to map removed classes to new ones based on tag similarity (https://github.com/BrickSchema/Brick/blob/compare-versions/tools/compare_versions.py#L66). It's not perfect, but we can use it to manually list them later.

- We could reuse the code in `bricksrc` such as versions but then we need to install it as a package, which I think would complicate the entire procedure without enough gain. I think it requires a bit of code restructure more.


gtfierro:
Good idea to put this together! I'll get a chance to look at it this weekend and provide feedback

gtfierro:
@jbkoh this is a nice tool! Some questions/suggestions

- can you add a README blurb documenting how to run this? It would be nice to take the Brick source files as command line arguments
- how did you determine the magic `0.7` parameter for the Jaccard calculation? We might want to generalize this 
- have you looked at other ontology comparison tools? I know that Protege has one. How does this compare?

jbkoh:
@gtfierro 

- *can you add a README blurb documenting how to run this? It would be nice to take the Brick source files as command line arguments*
    - Done
- *how did you determine the magic 0.7 parameter for the Jaccard calculation? We might want to generalize this*
    - I just selected it arbitrarily. It's hard to get a general number for this. I do not think we can easily come up with an algorithm for generalizing this for now. We can gradually update this algorithm if anybody has an idea.
- have you looked at other ontology comparison tools? I know that Protege has one. How does this compare?
    - I have not. However, as far as I know, the algorithms usually take the structural similarity but not string similarity. We can't exploit structural similarity for now from 1.0.3 to 1.1.0. We can try that out in the future versions though.
- how do you feel about hiding the history/ directory from Git so that it's never checked in? We could include the output in the releases of minor/major versions of Brick, but checking it in to master feels like overkill
    - I would like people to be able to easily look up the difference. I think the history files would be smaller between minor versions, but we can do it on just release pages as you suggested too. I just emptied the history folder.

gtfierro:
@jbkoh just pinging this issue. Let me know if you have time to delete the `history/.placeholder` file; else I can take care of that and merge.

shreyasnagare:
@jbkoh When I try to compare Brick v1.0.2 to v1.0.3, I get a blank `./history/1.0.3/removed_classes.txt` file and zero mappings. 

Arguments used:
```
python3 tools/compare_versions.py --oldbrick 1.0.2 https://github.com/BrickSchema/Brick/releases/download/v1.0.2/Brick.ttl --newbrick 1.0.3 https://github.com/BrickSchema/Brick/releases/download/v1.0.3/Brick.ttl
```

I want to make sure this is not just a 'MacOS' thing like the urllib SSL errors that we observed.

gtfierro:
This could mean that there are no removed classes between 1.0.2 and 1.0.3; are you able to find any evidence to the contrary?

shreyasnagare:
> This could mean that there are no removed classes between 1.0.2 and 1.0.3; are you able to find any evidence to the contrary?

Yeah, for instance, v1.0.3 has no `brickl:Power_Meter` or any of its subclasses. Moreover, if there were no removed classes, the file should consist of an empty array, at the least.

gtfierro:
- [ ] Remove `pdb`
- [ ] Remove `placeholder` file

jbkoh:
@shreyasnagare I will look into the problem later today.

shreyasnagare:
I think it would make more sense if every tool has a subfolder inside the `./tools` directory.

Something like:
Screen Shot 2019-10-25 at 3 06 44 PM

We can either add the tools here dircetly or add submodules to the repo.

gtfierro:
@shreyasnagare good idea! @jbkoh could you put these files inside a nested directory within the `tools/` folder? We can start moving other tools into that folder as well as necessary

jbkoh:
@gtfierro Moved the script into a dir. I will merge this one once the verification is done.

jbkoh:
> > @shreyasnagare I will look into the problem later today.
> 
> @jbkoh Did you face the same problem with the `--oldbrick 1.0.2` argument?

It is fixed at https://github.com/BrickSchema/Brick/pull/89 I will merge that once the verify is done.

PR #81, opened 2019-09-30
Fix band parameter subclasses

shreyasnagare:
brick:Supply_Water_Differential_Pressure_Proportional_Band_Parameter shouldn't be a subclass of itself.


shreyasnagare:
The only tag lost because of this operation is "TAG.Bandsetpoint". If it is clear what it means, we can add it to the remaining list of tags.

gtfierro:
Good catch. LGTM!

PR #85, opened 2019-10-22
SHACL Shapes for Brick (proof of concept)

gtfierro:
**Do not merge this branch yet!**

This PR is a forum for discussing the integration of SHACL into Brick. The existing implementation is very spare and incomplete, and I expect it to change over time. Below is the content of `DESIGN.md` from the branch:

---

# SHACL for Brick

The SHACL Shapes Constraint Language validates RDF graphs against conditions (called "shapes"), which are expressed themselves as an RDF graph. This document discusses the application of SHACL to Brick. In this document we are specifically looking at writing SHACL shapes for validating Brick *models* (instances of buildings). Using SHACL to validate the Brick ontology deserves its own investigation.

SHACL shapes for Brick should validate the following:

- validating endpoints of Brick relationships (properties):
  - check type of subject/object: are the `rdfs:range`, `rdfs:domain` properties respected?
  - cardinality checks: with varying severity
    - floors without rooms
    - equipment missing certain points (e.g. thermostat should have a temperature sensor, at least one setpoint)
    - all named individuals should have `rdf:type`
- compatibility checking:
  - all objects of `rdf:type` should be Brick classes (subclassof, etc)
  - check for deprecated classes, relationships
- idioms, semantic checks:
  - these may be optional; users can "opt-in" to check for certain idioms
  - equipment along a chain of `brick:feeds` should all operate on the same substance
  - proper relationships exist between equipment: e.g. thermostat should have `controls` relationship to RTU

## Getting Started

Install `pyshacl` from pip.

`generate_shacl.py` uses `bricksrc/` to generate a Shapes graph that is serialized to `shacl_test.ttl`.

To validate a graph, use [pySHACL](https://github.com/RDFLib/pySHACL):

```bash
pyshacl -s shacl_test.ttl -m -e Brick.ttl -f human sample_graph.ttl
```

You may want to expand a graph before validating it. In that case, run

```bash
python expand_graph.py sample_graph.ttl
```

which will drop the expanded graph into `output.ttl` in the same directory.

## Brick SHACL shapes

Going through what the SHACL shapes would look for for different tests. The examples below declare shapes in the `bsh` ("brickshape") namespace for clarity

### Property Constraints

Property definitions (from the `bricksrc/properties.py` file) define inverses, domains/ranges of properties and subproperties. The shape graph would contain one shape for each Brick property that encodes constraints on the domain and/or range of the property.

**TODO**: so far, I'm only sure how to encode the `rdfs:range` property

For example, the definition of `brick:hasLocation` is

```python
{
  "hasLocation": {
    A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
    OWL.inverseOf: "isLocationOf",
    RDFS.range: BRICK.Location,
    SKOS.definition: Literal("Subject is physically located in the location given by the object"),
  },
}
```

This definition informs the following Shape for the `rdfs:range` property

```ttl
bsh:HasLocationRangeShape
    a   sh:NodeShape ;
    sh:targetObjectsOf brick:hasLocation ;
    sh:property [
        sh:class    brick:Location ;
        sh:path     brick:hasLocation ;
        sh:nodeKind sh:IRI ;
    ] .
```

For properties with both a domain and a range, we will need both a "domain" Shape and a "range" Shape; however, at this time I'm not sure how to encode constraints for `rdfs:domain`.

### Special Property Constraints

We will also want to define special shapes for certain properties that have additional semantics. For example, the `feedsAir` subproperty should require that all equipment on its endpoints should have the `brick:regulates brick:Air` property.

We can model this by requiring that all subjects *and* objects of the `brick:feedsAir` property have the `brick:Air` value for either a `brick:requlates` or `brick:measures` property.

```ttl
bsh:FeedsAirSubjectShape a sh:NodeShape ;
    sh:targetSubjectsOf brick:feedsAir ;
    sh:or (
        sh:property [
            sh:path brick:regulates ;
            sh:class brick:Air ;
        ]
        sh:property [
            sh:path brick:measures ;
            sh:class brick:Air ;
        ]
    ) .

bsh:FeedsAirObjectShape a sh:NodeShape ;
    sh:targetObjectsOf brick:feedsAir ;
    sh:or (
        sh:property [
            sh:path brick:regulates ;
            sh:class brick:Air ;
        ]
        sh:property [
            sh:path brick:measures ;
            sh:class brick:Air ;
        ]
    ) .
```

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -30,6 +30,14 @@ To validate a graph, use [pySHACL](https://github.com/RDFLib/pySHACL):
 pyshacl -s shacl_test.ttl -m -e Brick.ttl -f human sample_graph.ttl
 ```
 
+You may want to expand a graph before validating it. In that case, run
+
+```bash
+python expand_graph.py sample_graph.ttl
+```
+
+which will drop the expanded graph into `output.ttl` in the same directory.
+
 ## Brick SHACL shapes
 
 Going through what the SHACL shapes would look for for different tests. The examples below declare shapes in the `bsh` ("brickshape") namespace for clarity
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -1,0 +1,117 @@
+# SHACL for Brick
+
+The SHACL Shapes Constraint Language validates RDF graphs against conditions (called "shapes"), which are expressed themselves as an RDF graph. This document discusses the application of SHACL to Brick. In this document we are specifically looking at writing SHACL shapes for validating Brick *models* (instances of buildings). Using SHACL to validate the Brick ontology deserves its own investigation.
+
+SHACL shapes for Brick should validate the following:
+
+- validating endpoints of Brick relationships (properties):
+  - check type of subject/object: are the `rdfs:range`, `rdfs:domain` properties respected?
+  - cardinality checks: with varying severity
+    - floors without rooms
+    - equipment missing certain points (e.g. thermostat should have a temperature sensor, at least one setpoint)
+    - all named individuals should have `rdf:type`
+- compatibility checking:
+  - all objects of `rdf:type` should be Brick classes (subclassof, etc)
+  - check for deprecated classes, relationships
+- idioms, semantic checks:
+  - these may be optional; users can "opt-in" to check for certain idioms
+  - equipment along a chain of `brick:feeds` should all operate on the same substance
+  - proper relationships exist between equipment: e.g. thermostat should have `controls` relationship to RTU
+
+
+## Brick SHACL shapes
+
+Going through what the SHACL shapes would look for for different tests. The examples below declare shapes in the `bsh` ("brickshape") namespace for clarity
+
+### Property Constraints
+
+Property definitions (from the `bricksrc/properties.py` file) define inverses, domains/ranges of properties and subproperties. The shape graph would contain one shape for each Brick property that encodes constraints on the domain and/or range of the property.
+
+For example, the definitions of `brick:isLocationOf` and `brick:hasLocation` are
+
+```python
+{
+  "isLocationOf": {
+    A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+    OWL.inverseOf: "hasLocation",
+    RDFS.domain: BRICK.Location,
+    SKOS.definition: Literal("Subject is the physical location encapsulating the object"),
+  }
+  "hasLocation": {
+    A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+    OWL.inverseOf: "isLocationOf",
+    RDFS.range: BRICK.Location,
+    SKOS.definition: Literal("Subject is physically located in the location given by the object"),
+  },
+}
+```
+
+This definition informs the following Shapes
+
+```ttl
+bsh:IsLocationOfShape
+    a   sh:NodeShape ;
+    sh:targetSubjectsOf brick:isLocationOf ;
+    sh:property [
+        sh:class    brick:Location ;
+        sh:nodeKind sh:IRI ;
+    ] ;
+    sh:closed false .
+
+bsh:HasLocationShape
+    a   sh:NodeShape ;
+    sh:targetObjectsOf brick:hasLocation ;
+    sh:property [
+        sh:class    brick:Location ;
+        sh:nodeKind sh:IRI ;
+    ] ;
+    sh:closed false .
+```
+
+For properties with both a domain and a range, we will need both a "domain" Shape and a "range" Shape:
+
+```ttl
+bsh:isMeasuredByDomainShape a sh:NodeShape ;
+    sh:property [ sh:class brick:Measurable ;
+            sh:nodeKind sh:IRI ] ;
+    sh:targetSubjectsOf brick:isMeasuredBy .
+
+bsh:isMeasuredByRangeShape a sh:NodeShape ;
+    sh:property [ sh:class brick:Point ;
+            sh:nodeKind sh:IRI ] ;
+    sh:targetObjectsOf brick:isMeasuredBy .
+```
+
+### Special Property Constraints
+
+We will also want to define special shapes for certain properties that have additional semantics. For example, the `feedsAir` subproperty should require that all equipment on its endpoints should have the `brick:regulates brick:Air` property.
+
+We can model this by requiring that all subjects *and* objects of the `brick:feedsAir` property have the `brick:Air` value for either a `brick:requlates` or `brick:measures` property.
+
+```ttl
+bsh:FeedsAirSubjectShape a sh:NodeShape ;
+    sh:targetSubjectsOf brick:feedsAir ;
+    sh:or (
+        sh:property [
+            sh:path brick:regulates ;
+            sh:class brick:Air ;
+        ]
+        sh:property [
+            sh:path brick:measures ;
+            sh:class brick:Air ;
+        ]
+    ) .
+
+bsh:FeedsAirObjectShape a sh:NodeShape ;
+    sh:targetObjectsOf brick:feedsAir ;
+    sh:or (
+        sh:property [
+            sh:path brick:regulates ;
+            sh:class brick:Air ;
+        ]
+        sh:property [
+            sh:path brick:measures ;
+            sh:class brick:Air ;
+        ]
+    ) .
+```
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -18,6 +18,17 @@ SHACL shapes for Brick should validate the following:
   - equipment along a chain of `brick:feeds` should all operate on the same substance
   - proper relationships exist between equipment: e.g. thermostat should have `controls` relationship to RTU
 
+## Getting Started
+
+Install `pyshacl` from pip.
+
+`generate_shacl.py` uses `bricksrc/` to generate a Shapes graph that is serialized to `shacl_test.ttl`.
+
+To validate a graph, use [pySHACL](https://github.com/RDFLib/pySHACL):
+
+```bash
+pyshacl -s shacl_test.ttl -m -e Brick.ttl -f human sample_graph.ttl
+```
 
 ## Brick SHACL shapes
 
@@ -27,16 +38,12 @@ Going through what the SHACL shapes would look for for different tests. The exam
 
 Property definitions (from the `bricksrc/properties.py` file) define inverses, domains/ranges of properties and subproperties. The shape graph would contain one shape for each Brick property that encodes constraints on the domain and/or range of the property.
 
-For example, the definitions of `brick:isLocationOf` and `brick:hasLocation` are
+**TODO**: so far, I'm only sure how to encode the `rdfs:range` property
+
+For example, the definition of `brick:hasLocation` is
 
 ```python
 {
-  "isLocationOf": {
-    A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
-    OWL.inverseOf: "hasLocation",
-    RDFS.domain: BRICK.Location,
-    SKOS.definition: Literal("Subject is the physical location encapsulating the object"),
-  }
   "hasLocation": {
     A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
     OWL.inverseOf: "isLocationOf",
@@ -46,41 +53,20 @@ For example, the definitions of `brick:isLocationOf` and `brick:hasLocation` are
 }
 ```
 
-This definition informs the following Shapes
+This definition informs the following Shape for the `rdfs:range` property
 
 ```ttl
-bsh:IsLocationOfShape
-    a   sh:NodeShape ;
-    sh:targetSubjectsOf brick:isLocationOf ;
-    sh:property [
-        sh:class    brick:Location ;
-        sh:nodeKind sh:IRI ;
-    ] ;
-    sh:closed false .
-
-bsh:HasLocationShape
+bsh:HasLocationRangeShape
     a   sh:NodeShape ;
     sh:targetObjectsOf brick:hasLocation ;
     sh:property [
         sh:class    brick:Location ;
+        sh:path     brick:hasLocation ;
         sh:nodeKind sh:IRI ;
-    ] ;
-    sh:closed false .
+    ] .
 ```
 
-For properties with both a domain and a range, we will need both a "domain" Shape and a "range" Shape:
-
-```ttl
-bsh:isMeasuredByDomainShape a sh:NodeShape ;
-    sh:property [ sh:class brick:Measurable ;
-            sh:nodeKind sh:IRI ] ;
-    sh:targetSubjectsOf brick:isMeasuredBy .
-
-bsh:isMeasuredByRangeShape a sh:NodeShape ;
-    sh:property [ sh:class brick:Point ;
-            sh:nodeKind sh:IRI ] ;
-    sh:targetObjectsOf brick:isMeasuredBy .
-```
+For properties with both a domain and a range, we will need both a "domain" Shape and a "range" Shape; however, at this time I'm not sure how to encode constraints for `rdfs:domain`.
 
 ### Special Property Constraints
 
--- a/bricksrc/namespaces.py
+++ b/bricksrc/namespaces.py
@@ -4,6 +4,8 @@ BRICK_VERSION = '1.1.0'
 
 BRICK = Namespace(f"https://brickschema.org/schema/{BRICK_VERSION}/Brick#")
 TAG = Namespace(f"https://brickschema.org/schema/{BRICK_VERSION}/BrickTag#")
+BSH = Namespace(f"https://brickschema.org/schema/{BRICK_VERSION}/BrickShape#")
+SH = Namespace(f"http://www.w3.org/ns/shacl#")
 #BLDG = Namespace(f"https://brickschema.org/schema/{BRICK_VERSION}/ExampleBuilding#")
 OWL = Namespace("http://www.w3.org/2002/07/owl#")
 RDF = Namespace("http://www.w3.org/1999/02/22-rdf-syntax-ns#")
@@ -22,5 +24,7 @@ def bind_prefixes(g):
     g.bind('rdfs', RDFS)
     g.bind('skos', SKOS)
     g.bind('sosa', SOSA)
+    g.bind('sh', SH)
     g.bind('brick', BRICK)
     g.bind('tag', TAG)
+    g.bind('bsh', BSH)
--- a/expand_graph.py
+++ b/expand_graph.py
@@ -1,0 +1,16 @@
+from rdflib import Graph, Namespace
+from bricksrc.namespaces import BRICK, RDF, OWL, DCTERMS, SDO, RDFS, SKOS, BRICK, TAG, SOSA, BSH, SH
+from bricksrc.namespaces import bind_prefixes
+from tests.util.reasoner import reason_brick
+import sys
+
+G = Graph()
+bind_prefixes(G)
+A = RDF.type
+G.parse(sys.argv[1], format='ttl')
+#G.parse('Brick.ttl', format='ttl')
+
+reason_brick(G)
+
+with open('output.ttl', 'wb') as f:
+    f.write(G.serialize(format='ttl'))
--- a/generate_shacl.py
+++ b/generate_shacl.py
@@ -1,16 +1,15 @@
-from rdflib import Graph, Literal, BNode, Namespace, RDF, URIRef, RDFS, OWL
-from rdflib.namespace import XSD
+from rdflib import Graph, Literal, BNode
 from rdflib.collection import Collection
-
-from bricksrc.namespaces import BRICK, RDF, OWL, DCTERMS, SDO, RDFS, SKOS, BRICK, TAG, SOSA, BSH, SH
+from bricksrc.namespaces import RDF, RDFS, BRICK, BSH, SH
 from bricksrc.namespaces import bind_prefixes
+# add SHACL shapes from properties
+from bricksrc.properties import properties as property_definitions
 
 G = Graph()
 bind_prefixes(G)
 A = RDF.type
-
-# add SHACL shapes from properties
-from bricksrc.properties import properties as property_definitions
+rootclasses = ["Equipment", "Location", "Point",
+               "Substance", "Quantity"]
 
 for name, properties in property_definitions.items():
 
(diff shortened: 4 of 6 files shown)


gtfierro:
Some initial items to discuss:

- what do we specifically want to validate?
- how to handle validation of `rdfs:domain` (I'm sure I'm missing something basic here)

xyzisinus:
Here is the Q&A between my and Gabe via email.  It's moved here as part of the conversation on SHACL w/ Brick.

Q:
> 2. In generate_shacl.py, your first todo is "# TODO: broken SHACL for
> rdfs.domain".  I experimentally uncommented the code below that.  The
> program finishes without error but I'm not good enough to see in the
> resulting shacl_test.ttl what is broken as indicated in your
> comments.  Can you explain?

A:
For the rdfs:domain issue, read the "Property Constraints" section on the SHACL pull request. The code will run without errors, but the "shapes" that are generated by generate_shacl.py do not actually validate what is communicated by rdfs:domain. I'm not sure how to validate the "subject" of a relationship using SHACL; validating the "object" of the relationship is fairly straightforward though.  Validating the inverse relationship is a partial solution, but not all of our properties have inverses, so this isn't going to work everywhere.

Put another way, we can express "if entity X has property Y to entity Z, then Z must be of type C1" (rdfs:range), but I don't know how to express "if entity X has property Y to entity Z, then X must be of type C2" (rdfs:domain). This is because SHACL shapes (seem to) only concern the objects of properties and not the subjects. I'm probably missing something though

xyzisinus:
Gabe, Please look at file https://github.com/xyzisinus/Brick/blob/czang-shacl/shacl/sample_graph.ttl.  It's the same file as you checked into the shacl proof of concept branch with an error inserted (line 28).  But it doesn't generate any error message by pyshacl.  I thought it should be found by the bsh:hasLocationRangeShape rule.  Am I missing something here?

gtfierro:
Are you expanding the graph before running SHACL? (e.g. `python expand_graph.py sample_graph.ttl`)

If you expand the graph, then OWL will infer that AHU is also a Location by virtue of it being the object of a `hasLocation` property, so I think my instructions in `DESIGN.md` are incorrect and you will want to run SHACL _before_ expanding the graph.

What will assist in this process is encoding into the Brick ontology that each of the Brick root classes are disjoint: Equipment, Location, Point, etc. Then the OWL inference should at least raise an error that it is inferring disjoint classes.

xyzisinus:
I just tried pyshacl on the expended graph.  The triple in question is expaned into
bldg:Room-412 a rdfs:Resource,
	brick:Room ;
    brick:hasLocation brick:Air_Handler_Unit .
and there is no error raised.  As Gabe points out, if the AHU is inferred as a location, then there is no error.  But if we let the reasoner to do so it defies the purpose of shacl check.  So I'll try to introduce disjointWith to see what happens.

gtfierro:
Before you try with `disjointWith`, can you verify whether or not an error is raised when you run SHACL on the non-expanded graph? SHACL stuff should work independently of the OWL definition

xyzisinus:
It doesn't raise error on non-expanded graph, i.e. sample_graph.ttl + that particular tripple.  I also tried with/without -e Brick.ttl.  The result is the same.

gtfierro:
I would dig into that issue first. It may be that my SHACL shape is incorrect

xyzisinus:
brick:Location a owl:Class ;
+    owl:disjointWith brick:Equipment ;
     rdfs:subClassOf brick:Class ;
I tried the above in Brick.ttl.  Is that the right thing?  Anyway it doesn't make a difference.  I'll play the a bit with the shacl shape, too, although I'm still in fog now :)

gtfierro:
It may be that the reasoner isn't identifying logical conflicts, which is annoying, but I can take a closer look in a week or so. I would focus on the SHACL stuff for now. Get some basic shapes working and work your way up from there.

xyzisinus:
Thanks but take your time.  I do plan to build from the basic stuff up little by little.

gtfierro:
@xyzisinus not sure if you've seen these but I would look at [`targetSubjectsOf`](https://www.w3.org/TR/shacl/#targetSubjectsOf) and [`targetObjectsOf`](https://www.w3.org/TR/shacl/#targetObjectsOf) for handling the domain/range stuff.

xyzisinus:
Thanks Gabe.  Those sections in SHACL have been read multiple times :)  I used my understanding to manually construst the shape constraints for, say, brick:hasLocation, and come up with the shape logically identical to yours.  Currently I'm doing two things:  1. Set the huge Brick graph aside but construct a very small RDF and its shape rules that everyone (we) agree is correct.  And construct a data graph that is clearly a good/bad one.  Then we test pyshacl using the small example.  With that we can also try the shacl playground https://shacl.org/playground/ (js based tool) to see how it works.

xyzisinus:
The second thing I'm doing: I looked at the code of pyshacl and don't see how targetObjectsOf can be collected in https://github.com/RDFLib/pySHACL/blob/master/pyshacl/shapes_graph.py starting from line 187.  Actually running the code with some print statements confirms my hunch.  When I have the little example complete (in a day) then I'll try to doctor the code a bit.   Also the example, if we agree is correct, will be sent to pyshacl as an issue.  Please comment on the plans.

xyzisinus:
Please look at https://github.com/xyzisinus/shacl-example and comment.  I'm not treaking pyshacl code as mentioned above.

xyzisinus:
Also note that we could use the "closed" constraint to generate desired error.

xyzisinus:
I have filed an issue with pyshacl github, pointing them to the shacl-example repo.  They replied very quickly.  The conversation is here https://github.com/RDFLib/pySHACL/issues/37.

gtfierro:
I've had a chance to read through the conversation and the examples, and I want to check if I'm up to speed.

It seems like one of the issues is we were using `targetObjectsOf` when we should have been using `targetSubjectsOf`, so for our `rdfs:range` shape for `brick:hasLocation`, our shape should have been something like

```turtle
bsh:HasLocationRangeShape
    a   sh:NodeShape ;
    # this is the line that changed from targetObjectsOf
    sh:targetSubjectsOf brick:hasLocation ;  
    sh:property [
        sh:class    brick:Location ;
        sh:path     brick:hasLocation ;
        sh:nodeKind sh:IRI ;
    ] .
```

which could also be

```turtle
bsh:HasLocationRangeShape
    a   sh:NodeShape ;
    sh:targetObjectsOf brick:hasLocation ;  
    sh:class    brick:Location .
```

These both seem to work ok on the SHACL playground, and I imagine pySHACL would handle them as well. I think my conceptual issue in the first draft of the SHACL shapes for Brick was confusing how the targets of the shape were established. 

The deeper issue, though, is that the use of RDFS inference (which is necessary for validation) with the `rdfs:range` and `rdfs:domain` properties ends up producing logical conflicts. Making the classes disjoint using `owl:disjointWith` doesn't seem to work as well, so I've added some new SHACL shapes in a recent commit that are one solution to this issue (see my [commit message here](https://github.com/BrickSchema/Brick/commit/1fc8768f6271138cf387aee3cc3219f5cb3aec7a))

The new shapes seem to work ok. For the example you had where an air handler unit is the object of a `brick:hasLocation` relationship, I get the following.

```bash
pyshacl -s shacl_test.ttl -m -i rdfs -a -f human -e Brick.ttl sample_graph.ttl

Validation Report
Conforms: False
Results (2):
Constraint Violation in NotConstraintComponent (http://www.w3.org/ns/shacl#NotConstraintComponent):
        Severity: sh:Violation
        Source Shape: bsh:LocationCanNotBeEquipmentShape
        Focus Node: bldg:AHU1A
        Value Node: bldg:AHU1A
Constraint Violation in NotConstraintComponent (http://www.w3.org/ns/shacl#NotConstraintComponent):
        Severity: sh:Violation
        Source Shape: bsh:EquipmentCanNotBeLocationShape
        Focus Node: bldg:AHU1A
        Value Node: bldg:AHU1A
```

We may want to fold in the disjointness using `sh:not` into the actual property shapes (e.g. `hasLocationShape`) instead so that we know where the issue is coming from.

---

To one of your other comments: I'm reluctant to adopt `sh:closed` for the shapes, because we may have proprietary/external properties that may not have shapes. Including `sh:closed` for those scenarios would result in those external properties being flagged.

xyzisinus:
Thanks Gabe.  Your two shapes for hasLocation are identical to the "update 2" in issue https://github.com/RDFLib/pySHACL/issues/37
That is great.  I'll integrate you change into my fork.

xyzisinus:
1. With pre-inferencing many constraint violations are generated with command
```
pyshacl -s shacl_test.ttl -i -a -m -f human -e Brick.ttl sample_graph.ttl
```
to stderr.   Are they just the trace of the inferencing process?  Should I ask the pyshacl team about it?
2. Please see the last two tripples in [sample_graph.ttl](https://github.com/xyzisinus/Brick/blob/czang-shacl/shacl/sample_graph.ttl).  I expect them to generate a violation with this shape:
```
bsh:feedsAirSubstanceShape a sh:NodeShape ;
    sh:or ( [ sh:property [ sh:class brick:Air ;
                        sh:path brick:regulates ] ] [ sh:property [ sh:class brick:Air ;
                        sh:path brick:measures ] ] ) ;
    sh:targetSubjectsOf brick:feedsAir .  <<< I added this
```
but they don't, even after adding the disjointness shapes between Gas and Liquid (see generate_shacl.py in the same directory).  From the [stderr output of pyshacl](https://github.com/xyzisinus/Brick/blob/czang-shacl/shacl/pyshacl.stderr), I can find
```
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
        Severity: sh:Violation
        Source Shape: [ sh:class brick:Air ; sh:path brick:regulates ]
        Focus Node: bldg:VAV2-4.DPR
        Value Node: brick:Water
        Result Path: brick:regulates
```
That means the inference has focused on 
```
bldg:VAV2-4.DPR brick:regulates brick:Water .
```
via sh:targetSubjectsOf with the following triple
```
viabldg:VAV2-4.DPR brick:feedsAir bldg:Room-412 .
```
But why it's not considered a real violation?

gtfierro:
For (1); are you talking about your fork of Brick, or the shacl branch on the original repo? I can try to reproduce and take a look.

For (2); what do you mean its not a real violation? Isn't the constraint violation what we expect?

xyzisinus:
Info for
Q 1:  Yes see [branch czang-shacl directory shacl in my forked repo](https://github.com/xyzisinus/Brick/tree/czang-shacl/shacl).  Run ```./validate.sh sample_graph.ttl``` and check ```pyshacl.stderr```.  

Running ```pyshacl -s shacl_test.ttl -m -i rdfs -a -f human -e Brick.ttl sample_graph.ttl``` in YOUR branch off the official Brick repo generates many lines to stderr, too.  With eyes focused on the final report, I didn't pay the stderr output much attention until now.

Q 2: The constraint violation I quoted above and here again
```
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
        Severity: sh:Violation
        Source Shape: [ sh:class brick:Air ; sh:path brick:regulates ]
        Focus Node: bldg:VAV2-4.DPR
        Value Node: brick:Water
        Result Path: brick:regulates
```
is found in the STDERR output which is 8000+ lines long (hence my question 1).  It is NOT in pyshacl's report which is sent to stdout:
```
Validation Report
Conforms: False
Results (4):
Constraint Violation in NotConstraintComponent (http://www.w3.org/ns/shacl#NotConstraintComponent):
	Severity: sh:Violation
	Source Shape: bsh:LocationCanNotBePointShape
	Focus Node: bldg:Room-410
	Value Node: bldg:Room-410
Constraint Violation in NotConstraintComponent (http://www.w3.org/ns/shacl#NotConstraintComponent):
	Severity: sh:Violation
	Source Shape: bsh:LocationCanNotBeEquipmentShape
	Focus Node: bldg:AHU1A
	Value Node: bldg:AHU1A
Constraint Violation in NotConstraintComponent (http://www.w3.org/ns/shacl#NotConstraintComponent):
	Severity: sh:Violation
	Source Shape: bsh:EquipmentCanNotBeLocationShape
	Focus Node: bldg:AHU1A
	Value Node: bldg:AHU1A
Constraint Violation in NotConstraintComponent (http://www.w3.org/ns/shacl#NotConstraintComponent):
	Severity: sh:Violation
	Source Shape: bsh:PointCanNotBeLocationShape
	Focus Node: bldg:Room-410
	Value Node: bldg:Room-410
```
Those errors originate from your shapes for disjointness in generate_shacl.py, caused by the following two lines in sample_graph.ttl:
```
bldg:Room-412 brick:hasLocation bldg:AHU1A .
bldg:VAV2-4.DPR brick:hasPoint bldg:Room-410 .
```
The newly introduced error by the last two line of sample_graph are not in the report.

gtfierro:
For Q1: I'm not able to produce that output from running `validate.sh` for some reason, but it is odd that certain constraint violations are being logged to stderr. These constraint violations don't look like they are explicitly defined in our own shapes file, so perhaps these are implied SHACL shapes defined by pySHACL (implied by our use of OWL or RDFS properties, I think). These hypothesis seems to be supported by the nature of the constraints output to stderr. Can you check the exit code of running pySHACL as well? 

This hypothesis *isn't* supported by having the `VAV2-4.DPR` constraints show up in stderr, however. Maybe this has something to do with severity of the shape? 

Which `Brick.ttl` are you using for your tests, and what version of `pyshacl` are you running?

xyzisinus:
For Q1: please see

https://github.com/xyzisinus/Brick/blob/czang-shacl/shacl/screenlog

which capture the session of pyshacl -s shacl_test.ttl -i -a -m -f human -e Brick.ttl sample_graph.ttl

The exit code of the command is 1 which in correct because it does report 4 errors.

xyzisinus:
For Q1 again: As I recall I didn't see those thousands lines of stderr either when I started playing with pyshacl.  My pyschacl version now is

pyshacl==0.11.3.post1

Maybe that is the difference?

xyzisinus:
Q1 is sort of resolved: I use my other computer to run the exact pyshacl command and do NOT see anything to stderr other than the final report.  But if I run with -d (debug)
```
pyshacl -d -s shacl_test.ttl -i -a -m -f human -e Brick.ttl sample_graph.ttl
```
then I see the same stderr output.  So somehow on one of my machines pyshacl is running with '-d' by default and on the other one it doesn't.  In both virtual environments I have the same pyshacl version.   I'll dig into why but at least we know the stderr output is produced by -d.

xyzisinus:
The Brick.ttl in my shacl directory is a link to the parent's Brick.ttl which is exactly the same with the on in the master branch of the main repo.

xyzisinus:
I forgot to mention that I'm doubtful that manually adding disjointness shapes is the way to solve our problem.  Now we are doing that to the top classes.  But there are many cases where disjointness is needed.  For example,  Gas is disjoint with Liquid, etc.  We can't capture all such cases.  I plan to reduce our problem to a manageable example and ask the pyshacl team why owl:disjointWith doesn't work for us.  Please let me know what you think.

gtfierro:
There are two distinct issues here: (1) is what OWL properties and constructs we use to model the concepts we want to model, (2) is how to write the SHACL shapes and invoke pySHACL such that we can correctly identify violations. `owl:disjointWith` is not involved in what pySHACL is doing and is not included in the SHACL shapes. Why do you think this is related? `owl:disjointWith` is used by a reasoner; it should not be used in the SHACL stuff.

I think you need to reduce the number of shapes you are generating and checking. Maybe start with shapes for the use of Brick relationships or stick with the "disjointness" of the classes (remember: the SHACL shape expression of disjointness is different than `owl:disjointWith` which is used by a reasoner). Keep in mind (or check explicitly) what gets implied by RDFS inference

Figure out which version of pySHACL you want to use. The issue you filed with them probably led you to install an older version

xyzisinus:
I mean to add own:disjointWith in Brick.ttl, for example:
```
brick:Point a owl:Class ;
    owl:disjointWith brick:Equipment ;
    owl:disjointWith brick:Location ;
...
```
With this information, should the reasoner consider Point, Location and Equipment different classes?

xyzisinus:
I just realized:  To make the owl predicates work in reasoning, maybe -i owlrl should be used for inferencing.  I tried it once and aborted the attemp because it took forever.  I'll try it on a smaller test case.  And also use the option and leave it running.

gtfierro:
Adding a `owl:disjointWith` constraint to Brick.ttl will not affect SHACL; it will only affect the reasoning. OWLRL reasoning is very expensive (as you discovered). I'm working on making this faster, but I do *not* recommend using it in the pre-inference step. The SHACL shapes should stand alone with RDFS inferencing

xyzisinus:
Experiment:

1. Remove all RDFS.range from properties.py and run ```generate_brick.py```.
2. Modify the shape in the existing shacl_test.ttl as we did yesterday:
```
bsh:feedsAirSubstanceShape a sh:NodeShape ;
    sh:property [ sh:class brick:Air ;
                  sh:path brick:regulates ] ;
    sh:targetSubjectsOf brick:feedsAir .
```
3. run
```
pyshacl -s shacl_test.ttl -i rdfs -a -m -f human -e Brick.ttl sample_graph.ttl
```
Now we get the violations exactly as we want:
```
Validation Report
Conforms: False
Results (3):
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
	Severity: sh:Violation
	Source Shape: [ sh:class brick:Air ; sh:path brick:regulates ]
	Focus Node: bldg:VAV2-4.DPR
	Value Node: brick:Water
	Result Path: brick:regulates
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
	Severity: sh:Violation
	Source Shape: [ sh:class brick:Point ; sh:message Literal("Property hasPoint has object with incorrect type") ; sh:nodeKind sh:IRI ; sh:path brick:hasPoint ]
	Focus Node: bldg:VAV2-4.DPR
	Value Node: bldg:Room-410
	Result Path: brick:hasPoint
	Message: Property hasPoint has object with incorrect type
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
	Severity: sh:Violation
	Source Shape: [ sh:class brick:Location ; sh:message Literal("Property hasLocation has object with incorrect type") ; sh:nodeKind sh:IRI ; sh:path brick:hasLocation ]
	Focus Node: bldg:Room-412
	Value Node: bldg:AHU1A
	Result Path: brick:hasLocation
	Message: Property hasLocation has object with incorrect type
```

gtfierro:
This is great! One point of clarification is that line 3 of sample_graph.ttl does not make the damper a focus node. Each of the "focus nodes" is computed on a per-shape basis, so I believe the SHACL verification should fail on the sample_graph even without the `feedsAir` line.

xyzisinus:
Thanks for the suggestions of going forward.  As for the focus node, I tried the remove the feedsAir tripple
```
#bldg:VAV2-4.DPR brick:feedsAir bldg:Room-412 .
bldg:VAV2-4.DPR brick:regulates brick:Water .
```
But the next tripple doesn't genete the desired errro.  The reason:
```
bsh:feedsAirSubstanceShape a sh:NodeShape ;
    sh:property [ sh:class brick:Air ;
                  sh:path brick:regulates ] ;
    sh:targetSubjectsOf brick:feedsAir .
```
applies to the subject of brick:feedsAir.   Without a "feedsAir" tripple, it can't find the damper to apply the shape.

xyzisinus:
Hi Gabe, Please look at [the latest in my branch](https://github.com/xyzisinus/Brick/tree/czang-shacl), in particular ```properties.py``` and ```generate_shacl.py```.  As we have discussed, I should remove the .range and .domain properties to generate a new Brick.ttl.  But I still need some hints to generate the range/domain related shapes.  The current solution/hack is to wrap those properties with SKOS.notation so that they can be used to generate the shapes while having no implication in Brick.ttl.  If you think there are better ways or better notation property to use, please let me know.

xyzisinus:
The following is the new violations generated with the domain shapes:
```
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
	Severity: sh:Violation
	Source Shape: bsh:isLocationOfDomainShape
	Focus Node: bldg:VAV2-3
	Value Node: bldg:VAV2-3
	Message: Property isLocationOf has subject with incorrect type
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
	Severity: sh:Violation
	Source Shape: bsh:measuresDomainShape
	Focus Node: bldg:VAV2-4.DPR
	Value Node: bldg:VAV2-4.DPR
	Message: Property measures has subject with incorrect type
```
This second one is a pleasant surprise:  I didn't add
```
bldg:VAV2-4.DPR brick:measures brick:Water .
```
to generate this particular violtation.
```
to generate this particuarl violation.

**Comment by xyzisinus** [comment]
But I'm having trouble with the following triples in ```sample_graph.ttl```:
```
# whatever feedsAir should regulate Air, not water
bldg:VAV2-4.DPR brick:feedsAir bldg:Room-412 .
bldg:VAV2-4.DPR brick:regulates brick:Water .
bldg:VAV2-4.DPR brick:measures brick:Water .
#bldg:VAV2-4.DPR brick:regulates brick:Air .
```
They produce violations:
```
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
	Severity: sh:Violation
	Source Shape: [ sh:class brick:Air ; sh:message Literal("Subject of property feedsAir has object with incorrect type") ; sh:nodeKind sh:IRI ; sh:path brick:regulates ]
	Focus Node: bldg:VAV2-4.DPR
	Value Node: brick:Water
	Result Path: brick:regulates
	Message: Subject of property feedsAir has object with incorrect type
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
	Severity: sh:Violation
	Source Shape: [ sh:class brick:Air ; sh:message Literal("Subject of property feedsAir has object with incorrect type") ; sh:nodeKind sh:IRI ; sh:path brick:measures ]
	Focus Node: bldg:VAV2-4.DPR
	Value Node: brick:Water
	Result Path: brick:measures
	Message: Subject of property feedsAir has object with incorrect type
```
That seems alright.  However, if I change the tripples to
```
# whatever feedsAir should regulate Air, not water
bldg:VAV2-4.DPR brick:feedsAir bldg:Room-412 .
#bldg:VAV2-4.DPR brick:regulates brick:Water .  <<< remove violation
#bldg:VAV2-4.DPR brick:measures brick:Water .  <<< remove violation
bldg:VAV2-4.DPR brick:regulates brick:Air .
```
There shouldn't be any voilation, right?  But I get
```
Constraint Violation in ClassConstraintComponent (http://www.w3.org/ns/shacl#ClassConstraintComponent):
	Severity: sh:Violation
	Source Shape: [ sh:class brick:Air ; sh:message Literal("Subject of property feedsAir has object with incorrect type") ; sh:nodeKind sh:IRI ; sh:path brick:regulates ]
	Focus Node: bldg:VAV2-4.DPR
	Value Node: brick:Air
	Result Path: brick:regulates
	Message: Subject of property feedsAir has object with incorrect type
```
Any idea how this happens?  Thanks!

**Comment by gtfierro** [comment]
Assuming the error message is correct, then we must assume the object of `regulates` is of the incorrect type. We expect it to be an instance of `brick:Air` (which should be punned, so `brick:Air` *should be* an instance of itself), but this isn't actually the case in the ontology definition. This is a bug in the ontology. When I correct this in the ontology definition then the shapes end up validating correctly. I will fix this in the #94 PR soon.

**Comment by gtfierro** [comment]
For the `SKOS.notation` hack, I think it would be preferable to define our own property e.g. `brick:expectedDomain` and `brick:expectedRange` to refer to the associated classes. This way we aren't abusing the original meaning of `SKOS.notation`

**Comment by xyzisinus** [comment]
> For the `SKOS.notation` hack, I think it would be preferable to define our own property e.g. `brick:expectedDomain` and `brick:expectedRange` to refer to the associated classes. This way we aren't abusing the original meaning of `SKOS.notation`

Good idea.  Will do.

**Comment by xyzisinus** [comment]
> Assuming the error message is correct, then we must assume the object of `regulates` is of the incorrect type. We expect it to be an instance of `brick:Air` (which should be punned, so `brick:Air` _should be_ an instance of itself), but this isn't actually the case in the ontology definition. This is a bug in the ontology. When I correct this in the ontology definition then the shapes end up validating correctly. I will fix this in the #94 PR soon.

Yes, I was told that brick:Air, etc, are class and instance of class, too.  But I just tried
```
bldg:something a brick:Air .
bldg:VAV2-4.DPR brick:regulates bldg:something .
```
and there is no more violation.  I'll add a comment for this case for now.

gtfierro:
Closing this PR; development being picked up on #97

PR #88, opened 2019-10-28
Add brick-convert

shreyasnagare:
This tool can update brick versions. If needed, we can write conversions scripts to rollback versions as well (I hope we don't need that). 

Instructions: `./tools/convert/README.md`

This PR will be ready to merge in once we have mappings for all v1.0.2 and v1.0.3 classes and relationships. The work related to mappings is being done on [this Google Sheet](https://docs.google.com/spreadsheets/d/1ZUEe1bouutdAvRAqU-Jq3IjEPGSH9aoRBXIK-t5MXb4/edit#gid=769540701).


shreyasnagare:
@jbkoh @gtfierro Try running the tool when you find some time and tell me if I missed something important. One problem that I know of is that unused namespace prefixes like `brickframe` are not removed. I was looking for an "unbind" function in RDFlib to remove them before serialization, but it doesn't exist. I might have to remove those prefixes after serialization unless you guys know the right way to do it.

Also, can you take a quick look at the third and fourth sheets of [the doc](https://docs.google.com/spreadsheets/d/1ZUEe1bouutdAvRAqU-Jq3IjEPGSH9aoRBXIK-t5MXb4/)?

jbkoh:
@shreyasnagare  thanks for putting all these together. However, I would like to suggest to use a more general framework for converting across different versions using inference engines instead of hard-coding SPARQL per class.

We could have a general framework in a turtle for provenance. For example
```turtle
@prefix brick101: <https://brickschema.org
brick110

prov:wasDerivedFrom
```

shreyasnagare:
@jbkoh Although there are certain limitations of using a generalized way for updating brick modeIs (which we can address later this week), using provenance data feels like the the right way to go. I have some implementation of this on a different [branch](https://github.com/shreyasnagare/Brick/tree/add-convert-prov) which isn't working right now. Can you look at this [query](https://github.com/shreyasnagare/Brick/blob/768cf9728219231fd47feb903beedc52e34d55a5/tools/convert/utils.py#L69) and tell me if I did something wrong here? The turtle file with provenance data can be found [here](https://github.com/shreyasnagare/Brick/blob/768cf9728219231fd47feb903beedc52e34d55a5/tools/convert/conversions/1.0.2-1.0.3.ttl#L48).

Running `python3 tools/convert/convert.py --source 1.0.2 --target 1.0.3 examples/ebu3b_brick.ttl` will parse the provenance data into the model graph and perform certain SPARQL update queries over it.

Regardless of the approach we use, we will need the Google Sheets to be complete first.

jbkoh:
Hi Shreyas, you are right the google sheet should be done regardless. I reviewed the entire 1.0.3->1.1.0. There are a few rules I would like to point out.


- Map `Resources` to `Substances` (not `Sensors`). And I do not think there are actual users of `Resources` in the past. Thus, if you can't find good matches, just don't map.
- When the meaning is unclear, it'd be great to refer to actual examples. If even it doesn't work, we should ask a domain expert.


I also feel there are many missing concepts and confusing concepts in the current dev version yet. I might need to create a separate Issue but let me enumerate them briefly here.

- Deadband vs Dead_Band. There is Proportional_Band, so it'd better stick to the same convention
- Equipment names still exist in Point names.
- Status and Alarm sometimes make little sense without associated Equipment. As always, we can explicitly associate them with Equipment, but I think it might be an interesting distinction to allow Equipment used in Status/Alarm/Command but not in Sensor/Setpoint. That's because Equipment can be a first-class citizen for those Points.

@shreyasnagare please take another look at my updates in the sheet. Then we can create a PR for missing classes before merging this PR as this one would depend on missing classes.

Thanks for the hard work!

jbkoh:
> @jbkoh Although there are certain limitations of using a generalized way for updating brick modeIs (which we can address later this week), using provenance data feels like the the right way to go. I have some implementation of this on a different [branch](https://github.com/shreyasnagare/Brick/tree/add-convert-prov) which isn't working right now. Can you look at this [query](https://github.com/shreyasnagare/Brick/blob/768cf9728219231fd47feb903beedc52e34d55a5/tools/convert/utils.py#L69) and tell me if I did something wrong here? The turtle file with provenance data can be found [here](https://github.com/shreyasnagare/Brick/blob/768cf9728219231fd47feb903beedc52e34d55a5/tools/convert/conversions/1.0.2-1.0.3.ttl#L48).

I myself found using variables for predicates often does not work as expected in SPARQL. I couldn't find any documentation about that, but you could look for documentation.

gtfierro:
> > @jbkoh Although there are certain limitations of using a generalized way for updating brick modeIs (which we can address later this week), using provenance data feels like the the right way to go. I have some implementation of this on a different [branch](https://github.com/shreyasnagare/Brick/tree/add-convert-prov) which isn't working right now. Can you look at this [query](https://github.com/shreyasnagare/Brick/blob/768cf9728219231fd47feb903beedc52e34d55a5/tools/convert/utils.py#L69) and tell me if I did something wrong here? The turtle file with provenance data can be found [here](https://github.com/shreyasnagare/Brick/blob/768cf9728219231fd47feb903beedc52e34d55a5/tools/convert/conversions/1.0.2-1.0.3.ttl#L48).
> 
> I myself found using variables for predicates often does not work as expected in SPARQL. I couldn't find any documentation about that, but you could look for documentation.

There's nothing in SPARQL that precludes this. @shreyasnagare could you produce a minimal reproducible example demonstrating the issue that you are seeing?

shreyasnagare:
@gtfierro After writing a minimal example, I found that query that adds the new classes/relationships works fine if I remove the queries that remove the old classes/relationships. Now, the following operations work:

- [x] Add new classes
- [x] Add new predicates
- [ ] Remove old classes
- [ ] Remove old relationships
- [x] Remove provenance data

I probably messed up [the DELETE queries](https://github.com/shreyasnagare/Brick/blob/d45214caf65ec5fce699b44c933d6fb17151eedc/tools/convert/utils.py#L65) and will look into the problem soon.

shreyasnagare:
**UPDATE:** I [added some filters](https://github.com/shreyasnagare/Brick/blob/48c57e268ee78c23eeabc5f1425be0646a47d626/tools/convert/utils.py#L65) to the DELETE queries and everything works fine now. I still don't understand why it needs the `( ?old != ?new )` filter. The conversion is only available from v1.0.2 to v1.0.3 as of now. 
It converts everything except 27 v1.0.2 classes (can be found on the 4th sheet of [this doc](https://docs.google.com/spreadsheets/d/1ZUEe1bouutdAvRAqU-Jq3IjEPGSH9aoRBXIK-t5MXb4/edit#gid=1040446114))

gtfierro:
Nice work, @shreyasnagare!

It looks like some of the classes in the `brick102` sheet are filled out now. Do you still need help filling some of these out?

 I think the majority of the effort should be on mapping 1.0.2 to 1.1.0 and 1.0.3 to 1.1.0 rather than 1.0.2 to 1.0.3 because we don't expect people to be using 1.0.3 after the new release. What's the status of those?

shreyasnagare:
> Nice work, @shreyasnagare!

Thanks :)

> It looks like some of the classes in the `brick102` sheet are filled out now. Do you still need help filling some of these out?

I'd really appreciate it if you could look at the `brick103` sheet. Set the `INFORMATION LOST?` filter to `(Blanks)`

> I think the majority of the effort should be on mapping 1.0.2 to 1.1.0 and 1.0.3 to 1.1.0 rather than 1.0.2 to 1.0.3 because we don't expect people to be using 1.0.3 after the new release. What's the status of those?

I'm thinking if we could reliably convert 1.0.2 to 1.0.3, we could just convert the intermediate 1.0.3 to 1.1.0. After a 1.0.2 model is converted to a 1.0.3 model, it's just another 1.0.3 model that can be converted to 1.1.0. So, technically it should produce the same results. Let me know what you think.

gtfierro:
There are a lot of these...have you tried removing the equipment name from the front of them? `AHU_`, `VAV_`, etc. Should bring down the number of unknown mappings a fair amount

I agree that 1.0.2 to 1.0.3 to 1.1.0 makes sense, but given that we are not planning on supporting v1.0.3, I think we can save time by just mapping the unknown v1.0.2 points directly to v1.1.0 rather than finding their intermediate mappings in v1.0.3

shreyasnagare:
> There are a lot of these...have you tried removing the equipment name from the front of them? `AHU_`, `VAV_`, etc. Should bring down the number of unknown mappings a fair amount

Yeah, I'll mark the confusing ones distinctly.

> I agree that 1.0.2 to 1.0.3 to 1.1.0 makes sense, but given that we are not planning on supporting v1.0.3, I think we can save time by just mapping the unknown v1.0.2 points directly to v1.1.0 rather than finding their intermediate mappings in v1.0.3

Okay, we can create a direct conversion from 1.0.2 to 1.1.0 since some common classes are missing in 1.0.3.

gtfierro:
@shreyasnagare how is this effort moving along since the big merge of the PRs yesterday? Let me know if I can help fix anything that has broken

shreyasnagare:
Here’s a list of all the "valid" (present in v1.0.2) classes used in the models that have no mapping in v1.1. I've also added the models in which they were used (in case anyone needs more context), and some mappings from the [Google Sheet](https://docs.google.com/spreadsheets/d/1ZUEe1bouutdAvRAqU-Jq3IjEPGSH9aoRBXIK-t5MXb4) (these mappings are not present in v1.1).

| Brick v1.0.2 Class                               | Used in | Possible mapping                                          |
|--------------------------------------------------|---------|-----------------------------------------------------------|
| Average_Cooling_Demand_Sensor                    | ebu3b   | Average_Cooling_Demand_Sensor                                                          |
| Cooling_Coil_Valve_Pressure                      | soda    |                                                          |
| Deceleration_Time_Setpoint                       | ebu3b   | Deceleration_Time_Parameter                               |
| Heat_Exchanger_Supply_Water_Temperature_Setpoint | ebu3b   | Heat_Exchanger_Supply_Water_Temperature_Deadband_Setpoint |
| Hot_Water_Coil_Entering_Temperature              | rice    |                                                           |

All the other classes and relationships (defined in brick/brickframe v1.0.2 and used in the models) are mapped.

gtfierro:
- `Average_Cooling_Demand_Sensor` (and `Average_Heating_Demand_Sensor`) will be added in a new PR soon
- `Cooling_Coil_Valve_Pressure` is a `Pressure_Sensor` associated with a `Cooling_Valve` (which is associated with a `Cooling_Coil`)
- `Deceleration_Time_Parameter` is the correct class
- `Heat_Exchanger_Supply_Water_Temperature_Setpoint` should be `Supply_Water_Temperature_Setpoint`
- `Hot_Water_Coil_Entering_Temperature` is an `Entering_Water_Temperature_Sensor` associated with a `Hot_Water_Coil`

shreyasnagare:
> * `Cooling_Coil_Valve_Pressure` is a `Pressure_Sensor` associated with a `Cooling_Valve` (which is associated with a `Cooling_Coil`)

So both of them become pressure sensors? 

Old | New
-- | --
Cooling_Coil_Valve_Pressure | ???
Cooling_Coil_Valve_Pressure_Sensor | Pressure_Sensor


> * `Deceleration_Time_Parameter` is the correct class
> * `Heat_Exchanger_Supply_Water_Temperature_Setpoint` should be `Supply_Water_Temperature_Setpoint`
> * `Hot_Water_Coil_Entering_Temperature` is an `Entering_Water_Temperature_Sensor` associated with a `Hot_Water_Coil`

gtfierro:
I think they do, yeah. It's common practice to eliminate the word `Sensor` in some BMS

gtfierro:
One thought: is there any way we can determine the "from" ontology version automatically by inspecting the namespaces in the source TTL file? There might also be value in having to explicitly mention this, but just a thought.

I tried to run a conversion example and got an error:

```
$ python convert.py --source 1.0.2 --target 1.1.0 ../../examples/ebu3b_brick.ttl
No conversions available from 1.0.2 to 1.1.0.
```

What am I missing?

tools/convert/convert.py
```diff
@@ -0,0 +1,82 @@
...
+    find_conversions,
+    execute_conversions,
+    standardize_namespaces,
+    get_output_filename,
+)
+
+"""
+Argument Parsing:
+source: source version (for example '1.0.3')
+target: target version (for example '1.1.0')
```
gtfierro: Change example to `1.1` rather than `1.1.0`

tools/convert/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Usage
+```bash
+python convert.py --source 1.0.2 --target 1.1.0 ../../examples/ebu3b_brick.ttl 
+```
+
+For multiple files, use glob patterns or filenames separated by whitespaces.
+```bash
+python convert.py --source 1.0.2 --target 1.1.0 ../../examples/*.ttl
+```
```
gtfierro: Make sure the versions are up to date (e.g. `1.1` instead of `1.1.0`)

gtfierro:
At this point, if @jbkoh is good with the changes then I am as well. Nice work, @shreyasnagare :)

shreyasnagare:
> One thought: is there any way we can determine the "from" ontology version automatically by inspecting the namespaces in the source TTL file? There might also be value in having to explicitly mention this, but just a thought.

I'm assuming that **not** all the input models will perfectly use a single brick namespace. Some users might want to use concepts across versions. And sometimes they might model custom brick subclasses in the same namespace(which we don't want to mingle with). Explicitly mentioning the source and target versions will give the users control over (and secretly hold them accountable for :P) the conversions.

tools/convert/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Help
+```bash
+python convert.py --help     
+```
+
+## Usage
+```bash
+python convert.py --source 1.0.2 --target 1.1 examples/ebu3b_brick.ttl
```
jbkoh: @shreyasnagare  Just want to make sure: is it usable only inside `tools/convert/`?

tools/convert/convert.py
```diff
@@ -0,0 +1,82 @@
...
+)
+args = parser.parse_args()
+
+
+# set log level to INFO if required.
+if args.info:
+    getLogger().setLevel("INFO")
+
+
+def convert(source=args.source, target=args.target, models=args.models):
```
jbkoh: Why do you define the function in this way instead of just putting the arguments at the end of this script?

tools/convert/conversions/1.0.2-1.1.json
```diff
@@ -0,0 +1,11995 @@
...
+    {
+      "description": "Remove brick:AHU_Discharge_Fan_Start_Stop_Command",
+      "query": "DELETE{ ?subject ?predicate brick_v_1_0_2:AHU_Discharge_Fan_Start_Stop_Command . } WHERE { ?subject  ?predicate  brick_v_1_0_2:AHU_Discharge_Fan_Start_Stop_Command . }"
+    },
+    {
+      "description": "Add brick:Start_Stop_Status",
+      "query": "INSERT{ ?subject ?predicate brick:Start_Stop_Status . } WHERE { ?subject  ?predicate  brick_v_1_0_2:AHU_Discharge_Fan_Start_Stop_Status . }"
+    },
+    {
+      "description": "Remove brick:AHU_Discharge_Fan_Start_Stop_Status",
```
jbkoh: @shreyasnagare Sorry, please correct me if I am wrong. It is removing `AHU_Discharge_Fan_Start_Stop_Status` but not replacing it with `Start_Stop_Status`. Could you point out where you add that?

jbkoh: nvm my bad. It's right above the line.

jbkoh:
@shreyasnagare How did you create `conversions/*.json`? If you have another script, I think it's more straightforward to merge insert and delete operations into a single operation. Similar to https://stackoverflow.com/a/38687956. It would be more error-prone and semantically makes sense.

tools/convert/README.md
shreyasnagare: To use it from the root directory, you must use `./tools/convert/convert.py` and `./tools/convert/examples/ebu3b_brick.ttl` instead of `convert.py` and `examples/ebu3b_brick.ttl`. I usually have _the README directory_ as the working directory. We can change the README to have the `./tools/convert/` path prefix in front of all the filenames if you think this is wrong.

tools/convert/convert.py
shreyasnagare: Sorry, I don't understand. Could you please post a sample snippet/link?

tools/convert/conversions/1.0.2-1.1.json
shreyasnagare: > @shreyasnagare Sorry, please correct me if I am wrong. It is removing `AHU_Discharge_Fan_Start_Stop_Status` but not replacing it with `Start_Stop_Status`. Could you point out where you add that?

The query right before this one adds that.

tools/convert/convert.py
jbkoh: Sorry I wasn't clear. Why do you unnecessarily define default values of the arguments here?

tools/convert/README.md
jbkoh: I just think we can make it executable later. Nothing urgent.

PR #92, opened 2019-12-24
Changes to Brick version numbers

gtfierro:
Encoding the version of the ontology into the URI is important for indicating to users which versions of the ontology are "backwards compatible", but we do not want users to have to change their URIs
often. I think that encoding the MAJOR and MINOR revision numbers into the URI (ignoring that PATCH revision) is a nice intermediate point.

This commit encodes the MAJOR, MINOR and PATCH version numbers in `bricksrc/namespaces.py`, and exposes the MAJOR_MINOR and FULL_VERSION constants that provide the full semantic version.

This means that for the upcoming release, v1.1.0, the Brick namespace would be

```
https://brickschema.org/schema/1.1/Brick
```

instead of 

```
https://brickschema.org/schema/1.1.0/Brick#
```

and it would say the former until we decided to release Brick 1.2.

Its worth noting that this is also the approach adopted by RealEstateCore.

Thoughts?

---

TODOs:
- [x] use `http://purl.org/dc/terms/hasVersion` to encode FULL_VERSION


jbkoh:
I like the idea. I have two requests to make this work.

1. Our governance structure should be changed accordingly as well. If we want to introduce or remove any classes for any reason (either a bug fix or new concepts), such changes need to go under MINOR revision in this workflow. We have put such changes under PATCH versions so far. We might have too many MINOR versions with the same practice. Can you clarify what would correspond to PATCH changes in our document?
2. We have to define FULL_VERSION somewhere in our schema so that any users who would like to check more rigorous versioning. We can simply add `http://purl.org/dc/terms/hasVersion` to our Ontology statement while the namespace does not have to be updated.

gtfierro:
I think our bug fixes and new concepts are new PATCH versions. Under the semver v2 semantics (which we don't have to keep, but are good guidelines), backwards compatible bug fixes are new PATCH versions. I think its reasonable to include backwards compatible additions to the concepts/tags to be also worthy of a PATCH version. The criteria for advancing a MINOR version would be a backwards compatible change to the modelling or inference strategy. Anything backwards incompatible would be a v2 release.

The key here is that the version number does not change often; if that means that we only include the MAJOR version in the URI, then that's ok with me.

I like the idea of using `http://purl.org/dc/terms/hasVersion` to include the full version. I'll include that in this PR

gtfierro:
@jbkoh this should be just about ready to merge.

Summary of the changes:

- Added `bricksrc/version.py` to contain the constants defining the current Brick version:

```python
# Major version: increment for backwards compatible changes
BRICK_MAJOR_VERSION = 1

# Minor version: increment for substantial additions to the ontology that
# are backwards compatible
BRICK_MINOR_VERSION = 1

# Patch version: increment for minor additions/changes to the ontology that
# are largely backwards compatible (bug-fixes exempted)
BRICK_PATCH_VERSION = 0

# the simplified (no patch version) version number for Brick. Intended for
# inclusion in the Brick namespace URI
BRICK_VERSION = f"{BRICK_MAJOR_VERSION}.{BRICK_MINOR_VERSION}"

# the full "semantic verersion" including the patch number
BRICK_FULL_VERSION = f"{BRICK_VERSION}.{BRICK_PATCH_VERSION}"
```

- all tests and source files now import from this file so we have a single point of "truth"
- the version number has been embedded in the ontology definition using `dcterms:version` as suggested above
- version number changes have been carried over to the `brickschema` python package, starting in version 0.0.12

bricksrc/version.py
```diff
@@ -0,0 +1,17 @@
+# Major version: increment for backwards compatible changes
+BRICK_MAJOR_VERSION = 1
+
+# Minor version: increment for substantial additions to the ontology that
+# are backwards compatible
+BRICK_MINOR_VERSION = 1
+
+# Patch version: increment for minor additions/changes to the ontology that
+# are largely backwards compatible (bug-fixes exempted)
```
jbkoh: What does "largely backward compatible" mean? It reads like PATCH_VERSION has a looser rule than MINOR_VERSION does.

gtfierro:
Bug fixes can be backwards incompatible (strictly speaking: typos and so
on)  Obviously this isn't carte blanche to make arbitrary changes. We will
have to demonstrate good judgement

generate_brick.py
```diff
@@ -38,24 +39,28 @@ def make_tag_classes(G, tag):
...
     for idnum, item in enumerate(definition):
         restriction = BNode()
-        l.append(restriction)
-        G.add( (restriction, A, OWL.Restriction) )
-        G.add( (restriction, OWL.onProperty, item[0]) )
-        G.add( (restriction, OWL.hasValue, item[1]) )
-    G.add( (BRICK[klass], OWL.equivalentClass, equivalent_class) )
-    G.add( (equivalent_class, OWL.intersectionOf, list_name) )
-    c = Collection(G, list_name, l)
+        elements.append(restriction)
```
jbkoh: Where do these changes come from? It seems unrelated to the version changes. I am fine with out-of-scope changes if they are intended, but I just want to make sure if these are not breaking.

jbkoh: Ah, I think it's related to the code style. I am fine with this.

gtfierro: This might be an artifact of the merge from `master` in the aftermath of merging #94. Sorry the merge was messy. The tests pass though!

gtfierro:
@shreyasnagare if I merge this (which changes the namespace URI), is the Brick website set up to handle it? I won't merge until I get your go-ahead

gtfierro:
It seems easier to fix everything if we don't need to worry about switching branches in order to test. Merging!

PR #93, opened 2019-12-24
Brick 1.1 release candidate 1

gtfierro:
- [x] needs list of added classes
- [x] needs list of removed classes

(these will both be easily fulfilled when #88 is merged)


gtfierro:
@shreyasnagare How hard is it to generate a list of added/removed classes from 1.0.3 to 1.1? Can you email that to me so I can include it in the changelog? Thanks!

shreyasnagare:
> @shreyasnagare How hard is it to generate a list of added/removed classes from 1.0.3 to 1.1? If its not too much trouble, can you email that to me so I can include it in the changelog? Thanks! Otherwise, I can just summarize the changes we made

Done. 

Can you verify that `./tools/compare_versions` is working fine? I'm getting a blank file for `added_classes.txt`.

gtfierro:
Thanks for sending the lists over!

We probably should use the full SemVer string when we're referring to the full versions e.g. for `compare_versions.py`

Is there any documentation for `compare_versions`? I don't know how it works

gtfierro:
@jbkoh let me know what you think of the CHANGELOG.md file and if I need to add anything. I'd like us to release v1.1 as soon as this PR is merged!

shreyasnagare:
> We probably should use the full SemVer string when we're referring to the full versions e.g. for `compare_versions.py`

That is a good idea but if you do that in `compare_versions.py`, you wouldn't get anything in the files -  `added_classes.txt` and `possible_mapping.json`. The version number is used to get the root class URI (which will have `1.1`).

> Is there any documentation for `compare_versions`? I don't know how it works

Try using the `--help` argument.

gtfierro:
> That is a good idea but if you do that in compare_versions.py, you wouldn't get anything in the files - added_classes.txt and possible_mapping.json. The version number is used to get the root class URI (which will have 1.1).

Given that we know how we want to handle the version numbers, its probably a good idea for `compare_versions.py` to handle this logic so that the full SemVer number can be used, but if the abbreviated MAJOR.MINOR number is used then it defaults to whatever the latest version in the `newbrick` TTL file is. 

I was able to get the `compare_versions` tool to give me a mapping from 1.1.0 to 1.0.3 after I edited the Brick.ttl version to 1.1.0, but no `added_classes.txt` appeared

jbkoh:
@gtfierro I am working on fixing compare_versions.py. This is because the version names are not updated from 1.1.0 to 1.1 in the previous PR.

jbkoh:
@gtfierro Please check #113

jbkoh:
- I'd rather be conservative about talking about `regulates`, `hasInputSubstance`, `hasOutputSubstance` because we have not discussed their usage enough. E.g., use cases, the correctness of the modeling, etc. 
- `Brick ontology now fits under OWL RL ontology profile`: What do you mean by this?
- I don't think ontology alignment is under the versioning umbrella unless the target ontology is tightly used inside/with Brick. (E.g., a unit ontology as a standard unit representation for Brick.)
- We could talk more about the tags <-> classes mapping with our model. It is one of the major features in this version.

gtfierro:
> I'd rather be conservative about talking about regulates, hasInputSubstance, hasOutputSubstance because we have not discussed their usage enough. E.g., use cases, the correctness of the modeling, etc. 

I think I'm already being pretty conservative. Would you like me to elide the mention of them completely? I think it's important to at least mention them

> Brick ontology now fits under OWL RL ontology profile: What do you mean by this?

The modeling practices of Brick are consistent with the OWL RL ontology profile; we don't require anything more powerful than what is defined by OWL RL.

> I don't think ontology alignment is under the versioning umbrella unless the target ontology is tightly used inside/with Brick. (E.g., a unit ontology as a standard unit representation for Brick.)

It's in the repo and thus part of the release, just like the tools. Keep in mind this is also what people will see about the Brick v1.1 release, so its important to mention what we have undertaken as part of this process

> We could talk more about the tags <-> classes mapping with our model. It is one of the major features in this version.

Sure, I can add more on this

gtfierro:
@jbkoh I put the Tag/Class discussion in the README and linked to it from the CHANGELOG. I also adjusted the OWL RL text in the CHANGELOG

jbkoh:
In general, I would like to distinguish 1) the official Brick standard and 1) what are in the repo. They are different even though they are in the same repo. People will mix them up if we don't clarify that. For example, Brick->RealEstateOntology is not a standard component of Brick though it will be useful and used by other people. Users need to know what they can heavily rely on.%^

I agree that non-standard updates are worth noting. We can have two sections inside this doc as 1 Brick Official and 2) Other Updates. We should move the description of tools to the later section as well.

> I think I'm already being pretty conservative. Would you like me to elide the mention of them completely? I think it's important to at least mention them

Adding unproved features to the versioning doc will confuse the users a lot. People will definitely ask how to use regulates, hasOutputSubstance, etc. We have no concrete answer to those other than some proofs-of-concept. They are definitely not in Brick standard. Though they may fall into Other Updates, I am worried it would cause more problems than benefits.

> The modeling practices of Brick are consistent with the OWL RL ontology profile; we don't require anything more powerful than what is defined by OWL RL.

I wanted easier words for this change, but I also don't know how to do it better. I will leave it to you.


> It's in the repo and thus part of the release, just like the tools. Keep in mind this is also what people will see about the Brick v1.1 release, so its important to mention what we have undertaken as part of this process

This can fall under Other Updates too.

gtfierro:
I can't say I share your worries above (e.g. "unproved features"; these have been implemented, tested, documented and peer-reviewed!), but I've distinguished between the Brick changes and tooling updates as you suggested, which is a good idea. Are there any other major changes we have to make before releasing v1.1?

jbkoh:
I am fine with this. Thanks!

gtfierro:
Thanks! We absolutely do still need to have conversations about these other modeling components, as you said; releasing v1.1.0 is not the end of this discussion.

gtfierro:
@shreyasnagare where's the list of root classes you are referring to? `brick:Measurable` is already in the generation script

shreyasnagare:
This is in the [**Classes section** of README](https://github.com/BrickSchema/Brick/blob/brick-1-1-release/README.md#classes).

gtfierro:
Fixed! Thanks for the pointer

PR #94, opened 2020-01-04
More changes for v1.1 release

gtfierro:
@jbkoh @shreyasnagare this PR encapsulates many of the changes we discussed in our last meeting. There's quite a few scattered changes:

- adding a first cut at an alarms hierarchy
- adding power/electrical equipment such as batteries
- adding the sensor, parameters, commands and setpoints that were missing
- adding quantities such as acceleration time and air flow loss

One issue I'm having is that tags attached to the substances can confusion when point/equipment classes also have those tags. Two suggestions on how to approach:
  1. remove tag inference from the substance hierarchy
  2. add additional tags to the substance hierarchy so they can be better differentiated

@jbkoh this is a specific instance of some of the tag issues you've mentioned. I'm putting together some of my thoughts on this, but in short, I think its ok if the formal tag mapping uses tags beyond what you might expect. For example, the tags for a `High_Temperature_Alarm_Setpoint` class could be `High`, `Temperature`, `Alarm`, `Setpoint` *and* `Parameter` in order to differentiate from the `setpoint` and `alarm` hierarchies. We can use a less formal inference to translate user-provided tags into Brick classes.

What's left to do:
- [x] incorporate `owl:equivalentClass` into the reasoning implementation
- [x] adding "long tail" classes like `Building_Exhaust_Fans_Stage_Command`

---

Actual changes (just pasted my notes, just in case you are curious):

Quantities:
~~- acceleration time~~
~~- air flow loss~~

Setpoints:
~~- (de)Acceleration time~~
~~- Cooling_Temperature_Setpoint:~~
    - needs to be added as a parent class
~~- Heating_Temperature_Setpoint:~~

Command:
- ~~lead_lag_command~~
- ~~Curtailment_Override_Command~~
- ~~Position_Command~~

Parameters:
- ~~Alarm_Delay_Parameter~~
- ~~Current_Limit_Parameter~~

Points:
~~- Add Alarm~~
~~- remove fan flavors of Air_Flow_Sensor~~
- ~~Cooling_Supply_Air_Temperature_Deadband_Setpoint~~
- ~~Cooling_Supply_Air_Temperature_Integral_Time_Parameter~~

Sensor:
- ~~Steam_Usage_Sensor~~
- ~~DC_Bus_Voltage_Sensor~~

Alarms:
- ~~High_CO2_Alarm~~
- ~~Change_Filter_Alarm~~
- ~~Discharge_Water_Temperature_Alarm~~
- ~~Failure Alarm~~
- ~~Supply_Water_Temperature_Alarm~~
- ~~Communication_Loss_Alarm~~
- ~~Overload_Alarm~~
- ~~Condensate_Leak_Alarm~~
- ~~Deionized_Water_Alarm~~

Equipment:
~~- add Battery under electrical equipment~~
- add ~~Chilled_Water_System~~ (equivalent to CWS)
- add meter:
    - ~~water meter (chilled variety as well)~~
    - ~~Cooling_Thermal_Power_Meter~~
- ~~cooling_valve~~

- ~~amend demand sensor tags/definition to remove power~~

### DIFF
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -1,8 +1,5 @@
-from rdflib import Graph, Literal, BNode, Namespace, RDF, URIRef
-from rdflib.collection import Collection
-from rdflib.extras.infixowl import Restriction
-
-from .namespaces import *
+from rdflib import Literal
+from .namespaces import TAG, OWL, SKOS, BRICK
 
 """
 Set up subclasses of the equipment superclass
@@ -15,9 +12,61 @@ equipment_subclasses = {
     "Weather": {
         "tags": [TAG.Weather],
     },
+    "Electrical_System": {
+        "subclasses": {
+            "Energy_Storage": {
+                "tags": [TAG.Energy, TAG.Storage, TAG.Equipment],
+                "subclasses": {
+                    "Battery": {
+                        "tags": [TAG.Battery, TAG.Energy, TAG.Storage,
+                                 TAG.Equipment],
+                    },
+                },
+            },
+            "Power_System": {
+                "tags": [TAG.Power, TAG.Equipment],
+            },
+            "PlugStrip": {
+                "tags": [TAG.Plugstrip, TAG.Equipment],
+            },
+            "Meter": {
+                "tags": [TAG.Meter, TAG.Equipment],
+                "subclasses": {
+                    "Water_Meter": {
+                        "tags": [TAG.Meter, TAG.Equipment, TAG.Water],
+                        "subclasses": {
+                            "Chilled_Water_Meter": {
+                                "tags": [TAG.Meter, TAG.Equipment, TAG.Water, TAG.Chilled],
+                            },
+                        },
+                    },
+                    "Power_Meter": {
+                        "tags": [TAG.Meter, TAG.Equipment, TAG.Power],
+                        "subclasses": {
+                            "Thermal_Power_Meter": {
+                                "tags": [TAG.Meter, TAG.Equipment, TAG.Power, TAG.Thermal],
+                                "subclasses": {
+                                    "Cooling_Thermal_Power_Meter": {
+                                        "tags": [TAG.Meter, TAG.Equipment, TAG.Power, TAG.Thermal, TAG.Cooling],
+                                    },
+                                    "Heating_Thermal_Power_Meter": {
+                                        "tags": [TAG.Meter, TAG.Equipment, TAG.Power, TAG.Thermal, TAG.Heating],
+                                    },
+                                },
+                            },
+                        },
+                    },
+                    "Building_Meter": {
+                        "tags": [TAG.Meter, TAG.Equipment, TAG.Building],
+                    },
+                },
+            },
+        },
+    },
     "Water_System": {
         "subclasses": {
             "Chilled_Water_System": {
+                OWL.equivalentClass: "CWS",
                 "tags": [TAG.Water, TAG.Chilled, TAG.Equipment],
             },
             "Hot_Water_System": {
@@ -33,20 +82,6 @@ equipment_subclasses = {
     },
     "Shading_System": {
         "tags": [TAG.Shade, TAG.Equipment],
-    },
-    "Power_System": {
-        "tags": [TAG.Power, TAG.Equipment],
-    },
-    "PlugStrip": {
-        "tags": [TAG.Plugstrip, TAG.Equipment],
-    },
-    "Meter": {
-        "tags": [TAG.Meter, TAG.Equipment],
-        "subclasses": {
-            "Building_Meter": {
-                "tags": [TAG.Meter, TAG.Equipment, TAG.Building],
-            },
-        },
     },
     "Lighting_System": {
         "tags": [TAG.Lighting, TAG.Equipment],
@@ -72,8 +107,6 @@ equipment_subclasses = {
     "Furniture": {
     },
     "Fire_Safety_System": {
-    },
-    "Energy_Storage": {
     },
     "Elevator": {
     },
@@ -270,6 +303,9 @@ valve_subclasses = {
                 "parents": [BRICK.Hot_Water_System, BRICK.Water_Valve],
             },
         },
+    },
+    "Cooling_Valve": {
+        "tags": [TAG.Valve, TAG.Cool, TAG.Equipment],
     },
     "Water_Valve": {
         "tags": [TAG.Valve, TAG.Water, TAG.Equipment],
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -5,7 +5,9 @@ from .namespaces import TAG, OWL, SKOS, BRICK
 Set up subclasses of the equipment superclass
 """
 equipment_subclasses = {
-    "HVAC": {},
+    "HVAC": {
+        OWL.equivalentClass: "Heating_Ventilation_Air_Conditioning_System",
+    },
     "Heating_Ventilation_Air_Conditioning_System": {
         OWL.equivalentClass: "HVAC",
     },
@@ -73,8 +75,15 @@ equipment_subclasses = {
                 "tags": [TAG.Water, TAG.Chilled, TAG.Equipment],
             },
             "Hot_Water_System": {
+                OWL.equivalentClass: "HWS",
                 "tags": [TAG.Water, TAG.Hot, TAG.Equipment],
             },
+            "CWS": {
+                OWL.equivalentClass: "Chilled_Water_System",
+            },
+            "HWS": {
+                OWL.equivalentClass: "Hot_Water_System",
+            }
         }
     },
     "Steam_System": {
@@ -282,7 +291,9 @@ hvac_subclasses = {
                 OWL.equivalentClass: "RTU",
                 "tags": [TAG.Equipment, TAG.Rooftop, TAG.AHU],
             },
-            "RTU": {},
+            "RTU": {
+                OWL.equivalentClass: "Rooftop_Unit",
+            },
         },
     },
 }
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -300,7 +300,7 @@ hvac_subclasses = {
         OWL.equivalentClass: "AHU",
     },
     "AHU": {
-        "tags": [ TAG.Equipment, TAG.AHU],
+        "tags": [TAG.Equipment, TAG.AHU],
         "subclasses": {
             "Rooftop_Unit": {
                 OWL.equivalentClass: "RTU",
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -16,6 +16,9 @@ equipment_subclasses = {
     },
     "Electrical_System": {
         "subclasses": {
+            "Emergency_Power_Off_System": {
+                "tags": [TAG.Emergency, TAG.Power, TAG.Off, TAG.Equipment],
+            },
             "Energy_Storage": {
                 "tags": [TAG.Energy, TAG.Storage, TAG.Equipment],
                 "subclasses": {
@@ -72,6 +75,7 @@ equipment_subclasses = {
         },
     },
     "Water_System": {
+        "tags": [TAG.Water, TAG.Equipment],
         "subclasses": {
             "Chilled_Water_System": {
                 OWL.equivalentClass: "CWS",
@@ -275,7 +279,7 @@ hvac_subclasses = {
     },
     "Coil": {
         SKOS.definition: Literal("Exchanger that transfers heat from an exhaust airstream to a separated supply airstream."),
-                "tags": [TAG.Equipment, TAG.Coil],
+        "tags": [TAG.Equipment, TAG.Coil],
         "subclasses": {
             "Cooling_Coil": {
                 "tags": [TAG.Equipment, TAG.Coil, TAG.Cool],
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -34,14 +34,17 @@ equipment_subclasses = {
                 "subclasses": {
                     "Water_Meter": {
                         "tags": [TAG.Meter, TAG.Equipment, TAG.Water],
+                        "parents": [BRICK.Water_System],
                         "subclasses": {
                             "Chilled_Water_Meter": {
                                 "tags": [TAG.Meter, TAG.Equipment, TAG.Water, TAG.Chilled],
+                                "parents": [BRICK.Chilled_Water_System],
                             },
                         },
                     },
                     "Power_Meter": {
                         "tags": [TAG.Meter, TAG.Equipment, TAG.Power],
+                        "parents": [BRICK.Power_System],
                         "subclasses": {
                             "Thermal_Power_Meter": {
                                 "tags": [TAG.Meter, TAG.Equipment, TAG.Power, TAG.Thermal],
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -32,7 +32,7 @@ equipment_subclasses = {
                 "tags": [TAG.Power, TAG.Equipment],
             },
             "PlugStrip": {
-                "tags": [TAG.Plugstrip, TAG.Equipment],
+                "tags": [TAG.PlugStrip, TAG.Equipment],
             },
             "Meter": {
                 "tags": [TAG.Meter, TAG.Equipment],
--- a/bricksrc/parameter.py
+++ b/bricksrc/parameter.py
@@ -383,10 +383,10 @@ parameter_definitions = {
                         },
                     },
                     "Fresh_Air_Setpoint_Limit": {
-                        "tags": [TAG.Fresh, TAG.Air, TAG.Setpoint, TAG.Limit],
+                        "tags": [TAG.Fresh, TAG.Air, TAG.Limit],
                         "subclasses": {
                             "Min_Fresh_Air_Setpoint_Limit": {
-                                "tags": [TAG.Min, TAG.Fresh, TAG.Air, TAG.Setpoint, TAG.Limit],
+                                "tags": [TAG.Min, TAG.Fresh, TAG.Air, TAG.Limit],
                                 "parents": [BRICK.Min_Limit],
                             },
                         },
@@ -402,7 +402,7 @@ parameter_definitions = {
                                 "tags": [TAG.Max, TAG.Static, TAG.Pressure, TAG.Limit, TAG.Parameter],
                             },
                             "High_Static_Pressure_Cutout_Setpoint_Limit": {
-                                "tags": [TAG.High, TAG.Static, TAG.Pressure, TAG.Cutout, TAG.Limit, TAG.Setpoint],
+                                "tags": [TAG.High, TAG.Static, TAG.Pressure, TAG.Cutout, TAG.Limit],
                             },
                         },
                     },
--- a/bricksrc/parameter.py
+++ b/bricksrc/parameter.py
@@ -227,155 +227,154 @@ parameter_definitions = {
                 }
             },
             "Limit": {
-                "tags": [TAG.Parameter, TAG.Limit],
+                "tags": [TAG.Parameter, TAG.Limit, TAG.Parameter],
                 "subclasses": {
                     "Speed_Setpoint_Limit": {
-                        "tags": [TAG.Speed, TAG.Limit],
+                        "tags": [TAG.Speed, TAG.Limit, TAG.Parameter],
                         "subclasses": {
                             "Max_Speed_Setpoint_Limit": {
-                                "tags": [TAG.Max, TAG.Speed, TAG.Limit],
-                                "parents": [BRICK.Max_Limit],
+                                "tags": [TAG.Max, TAG.Speed, TAG.Limit, TAG.Parameter],
+                                "parents": [BRICK.Max_Limit, TAG.Parameter],
                             },
                             "Min_Speed_Setpoint_Limit": {
-                                "tags": [TAG.Min, TAG.Speed, TAG.Limit],
-                                "parents": [BRICK.Min_Limit],
+                                "tags": [TAG.Min, TAG.Speed, TAG.Limit, TAG.Parameter],
+                                "parents": [BRICK.Min_Limit, TAG.Parameter],
                             },
                         },
                     },
                     "Air_Flow_Setpoint_Limit": {
-                        "tags": [TAG.Air, TAG.Flow, TAG.Limit],
+                        "tags": [TAG.Air, TAG.Flow, TAG.Limit, TAG.Parameter],
                         "subclasses": {
                             "Max_Air_Flow_Setpoint_Limit": {
-                                "tags": [TAG.Max, TAG.Air, TAG.Flow, TAG.Limit],
+                                "tags": [TAG.Max, TAG.Air, TAG.Flow, TAG.Limit, TAG.Parameter],
                             },
                             "Min_Air_Flow_Setpoint_Limit": {
-                                "tags": [TAG.Min, TAG.Air, TAG.Flow, TAG.Limit],
+                                "tags": [TAG.Min, TAG.Air, TAG.Flow, TAG.Limit, TAG.Parameter],
                             },
                         },
                     },
                     "Current_Limit": {
-                        "tags": [TAG.Current, TAG.Limit],
+                        "tags": [TAG.Current, TAG.Limit, TAG.Parameter],
                     },
                     "Damper_Position_Limit": {
-                        "tags": [TAG.Damper, TAG.Position, TAG.Limit],
+                        "tags": [TAG.Damper, TAG.Position, TAG.Limit, TAG.Parameter],
                         "subclasses": {
                             "Max_Damper_Position_Setpoint_Limit": {
-                                "tags": [TAG.Max, TAG.Damper, TAG.Position, TAG.Limit],
+                                "tags": [TAG.Max, TAG.Damper, TAG.Position, TAG.Limit, TAG.Parameter],
                             },
                             "Min_Damper_Position_Setpoint_Limit": {
-                                "tags": [TAG.Min, TAG.Damper, TAG.Position, TAG.Limit],
+                                "tags": [TAG.Min, TAG.Damper, TAG.Position, TAG.Limit, TAG.Parameter],
                             },
                         },
                     },
                     "Differential_Pressure_Setpoint_Limit": {
-                        "tags": [TAG.Differential, TAG.Pressure, TAG.Limit],
+                        "tags": [TAG.Differential, TAG.Pressure, TAG.Limit, TAG.Parameter],
                         "subclasses": {
                             "Max_Chilled_Water_Differential_Pressure_Setpoint_Limit": {
-                                "tags": [TAG.Max, TAG.Chilled, TAG.Water, TAG.Differential, TAG.Pressure, TAG.Limit],
+                                "tags": [TAG.Max, TAG.Chilled, TAG.Water, TAG.Differential, TAG.Pressure, TAG.Limit, TAG.Parameter],
                             },
                             "Min_Chilled_Water_Differential_Pressure_Setpoint_Limit": {
-                                "tags": [TAG.Min, TAG.Chilled, TAG.Water, TAG.Differential, TAG.Pressure, TAG.Limit],
+                                "tags": [TAG.Min, TAG.Chilled, TAG.Water, TAG.Differential, TAG.Pressure, TAG.Limit, TAG.Parameter],
                             },
                             "Max_Hot_Water_Differential_Pressure_Setpoint_Limit": {
-                                "tags": [TAG.Max, TAG.Hot, TAG.Water, TAG.Differential, TAG.Pressure, TAG.Limit],
+                                "tags": [TAG.Max, TAG.Hot, TAG.Water, TAG.Differential, TAG.Pressure, TAG.Limit, TAG.Parameter],
                             },
                             "Min_Hot_Water_Differential_Pressure_Setpoint_Limit": {
-                                "tags": [TAG.Min, TAG.Hot, TAG.Water, TAG.Differential, TAG.Pressure, TAG.Limit],
+                                "tags": [TAG.Min, TAG.Hot, TAG.Water, TAG.Differential, TAG.Pressure, TAG.Limit, TAG.Parameter],
                             },
                         },
                     },
                     "Static_Pressure_Setpoint_Limit": {
-                        "tags": [TAG.Static, TAG.Pressure, TAG.Limit],
+                        "tags": [TAG.Static, TAG.Pressure, TAG.Limit, TAG.Parameter],
                         "subclasses": {
                             "Min_Static_Pressure_Setpoint_Limit": {
-                                "tags": [TAG.Min, TAG.Static, TAG.Pressure, TAG.Limit],
+                                "tags": [TAG.Min, TAG.Static, TAG.Pressure, TAG.Limit, TAG.Parameter],
                             },
                             "Max_Static_Pressure_Setpoint_Limit": {
-                                "tags": [TAG.Max, TAG.Static, TAG.Pressure, TAG.Limit],
+                                "tags": [TAG.Max, TAG.Static, TAG.Pressure, TAG.Limit, TAG.Parameter],
                             },
                         },
-                        "parents": [BRICK.Static_Pressure_Setpoint],
                     },
                     "Max_Limit": {
-                        "tags": [TAG.Max, TAG.Limit],
+                        "tags": [TAG.Max, TAG.Limit, TAG.Parameter],
                         "subclasses": {
                             "Max_Speed_Setpoint_Limit": {
-                                "tags": [TAG.Max, TAG.Speed, TAG.Limit],
+                                "tags": [TAG.Max, TAG.Speed, TAG.Limit, TAG.Parameter],
                             },
                             "Max_Discharge_Air_Static_Pressure_Setpoint_Limit": {
-                                "tags": [TAG.Max, TAG.Discharge, TAG.Air, TAG.Static, TAG.Pressure, TAG.Limit],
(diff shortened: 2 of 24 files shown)


gtfierro:
Quick update: the substance hierarchy is not actually an issue (although it could be). Current difficulty is dealing with `alarm setpoint`s in the Paramater hierarchy; the `alarm` tag on these points means they also get inferred as Alarms in addition to Parameters. I'm trying out some ideas for how to handle this

bricksrc/parameter.py
```diff
@@ -301,21 +301,21 @@
                 },
             },
             "Limit": {
-                "tags": [TAG.Parameter, TAG.Limit],
+                "tags": [TAG.Parameter, TAG.Limit, TAG.Parameter],
```
shreyasnagare:
```suggestion
                "tags": [TAG.Parameter, TAG.Limit],
```

gtfierro:
# PR Notes

## Classes

Equipment:
- file: `bricksrc/equipment.py`
- removed unnecessary imports, tightened up Python styling (extra whitespace,
  etc)
- added missing classes from conversion spreadsheet
- added Electrical system classes with meters, inverters, batteries, etc:
    - some classes e.g. meter moved under the `Electrical_System` clas
- several classes were previously referred to only in `owl:equivalentClass`
  predicates, which caused some issues. These classes are now explicitly
  mentioned in the equipment class structure

### Points

Alarms:
- file: `bricksrc/alarm.py`
- added missing Alarm classes from the Brick 1.0.2 and 1.0.3 hierarchies

Command:
- file: `bricksrc/command.py`
- removed unnecessary imports, tightened up Python styling (extra whitespace, etc)
- added some missing classes from the conversion spreadsheet

Parameter:
- file: `bricksrc/parameter.py`
- removed unnecessary imports, tightened up Python styling (extra whitespace,
  etc)
- **Big change**: removed all mention of the `setpoint` tag from this hierarchy
  in order to avoid conflicts with the `Setpoint` point hierarchy. See the
  section on tag inference below for more
- added `Parameter` tag to *all* classes in the hierarchy; previously the
  `Limit` classes were exempt from this, but no longer
- fixed many small bugs: inconsistent tags, mistaken superclasses, etc
- added missing classes from conversion spreadsheet

Properties:
- file: `bricksrc/properties.py`
- removed unnecessary imports

Sensor:
- file: `bricksrc/sensor.py`
- removed unnecessary imports, extra whitespace
- added missing classes from conversion spreadsheet

Setpoint:
- file: `bricksrc/setpoint.py`
- removed unnecessary imports, extra whitespace
- added missing classes from conversion spreadsheet

Status:
- file: `bricksrc/status.py`
- removed unnecessary imports, extra whitespace

### Substances + Quantities

**Class Structure**: The class structure has been changed slightly.
`Measurable` is the new root class for `Substance` and `Quantity`. This makes
`Measurable` on the same level as `Point` and `Equipment` and the other
"classic" root classes.  Before, `Measurable` was a dangling class not
connected to the rest of the ontology.  `Measurable` is used as the
domain/range of the `brick:measures` and `brick:isMeasuredBy` properties.  This
makes `Quantity` and `Substances` "metaclasses" of the punned quantities and
substances: things like `brick:Air` are now *instances* of `brick:Substance` as
well as classes.

**Unit tests**: I added the `test_measurable_inference` test in order to verify
that using the `brick:measures` property correctly leads to the inference of
the proper classes.

Quantities:
- file: `bricksrc/quantities.py`
- removed unnecessary imports
- added missing classes from conversion spreadsheet

Substances:
- file: `bricksrc/substances.py`
- added `Fluid` tag to all members of the `Liquid` subclass, which is a
  subclass of `Fluid`. This was being done automatically before; now it is more
  explicit

## Brick Modeling

Through the course of making the additions and changes to the class structuve
above, I also delved into the use of reasoning in Brick v1.1. Specifically, I
was trying to address the following issues in Brick:

### Unsatisfiable Tag-based Inferences

Due to overlaps in the sets of tags used to describe a concept, it was
possible for inference to mistakenly connect an instance to two disjoint
classes. For example, considerthe following naive equivalencies:

| class | tags |
+-------+------+
| `Air_Flow_Setpoint` | `Air`, `Flow`, `Setpoint` |
| `Max_Air_Flow_Setpoint_Limit` | `Max`, `Air`, `Flow`, `Setpoint`, `Limit` |

In Brick, `Air_Flow_Setpoint` is a subclass of the `Setpoint` class.
`Max_Air_Flow_Setpoint_Limit` is a subclass of the `Parameter` class. The
`Setpoint` and `Parameter` classes are disjoint. If an entity `e` has the
requisite tags for `Max_Air_Flow_Setpoint_Limit`, because this set of tags is a
superset of those for `Air_Flow_Setpoint`, entity `e` will be inferred as an
instance of *both* `Max_Air_Flow_Setpoint_Limit` (correctly) and
`Air_Flow_Setpoint` (incorrectly). We want to avoid these types of confusions.
Technically, this makes classes with such an overlap "unsatisfiable" because
the two inferred classes violate the disjointness axiom.

I considered the following solutions to this issue:

1. Change the assignment of tags to classes such that no overlaps exist:
    - **pros**: doesn't require computing complements (see below); modeling strategy is straightforward with no complex logic; requires no changes to the data model
    - **cons**: in order to maintain the invariant that the "lattice" of tags does not cross the disjoint class boundary, we may need to adopt slightly less intuitive sets of tags, e.g. `Max_Air_Flow_Setpoint_Limit` may have tags `Max`, `Air`, `Flow`, `Limit` and `Parameter`.
2. Add additional restrictions to the tag-to-class logic so that unsatisfiable inference is impossible; for example "entities with the `Parameter` cannot be instances of `Setpoint`"
    - **pros**: lets us use sets of tags that feel more "intuitive", e.g. using the tags in the table above; do not need these rules for every class, just potentially conflicting ones
    - **cons**: requires computing complements, which can be expensive; I have an implementation but there's some trickiness in the logic that I still have to work through
3. Remove the OWL-based inference of tags to classes and implement this externally:
    - **pros**: removes all modeling issues; very simple; we can still associate tags with classes, we just can't use them for OWL-based inference
    - **cons**: having a formal equivalence between tags and classes means that the rules for translating between them have a formal definition. If we were to implement this translation in code, then we would have to write the logic down somewhere and in some form. Doing this through OWL gives us a path for now

Option 1 is the most ideal to me because it doesn't involve changing the data
model. I've implemented this to some extent for the `Parameter` hierarchy: none
of the classes in `Parameter` have the `Setpoint` tag, even if `Setpoint`
appears in the text of the class. This may feel unintuitive for cases such as
`Max_Air_Flow_Setpoint_Limit`, but I think its justifiable because the mapping
between classes and tags in Brick is *by definition*.

I've also implemented some of Option 2 as spot fixes for other parts of the
hierarchy.  I create a new anonymous class that is the complement of having a
tag -- this lets us model *not* having a tag.  Then, we can model certain
classes in terms of the tags they don't have. We apply this in two cases:

- `Alarm`: We add an additional constraint to all of the tag mappings in the
  `Alarm` hierarchy that they don't have either the `Parameter` nor the
  `Setpoint` tag. This means that classes of `Alarm` cannot have those two
  tags.
- `Temperature_Sensor`: to avoid overlap between `Temperature_Adjust_Sensor`
  and `Temperature_Sensor` classes (which are distinct), we add the restriction
  that `Temperature_Sensor` subclasses cannot have the `Adjust` tag.

### Reasoning Time

You may recall (or have had the misfortune of discovering for yourself) how
slow OWL reasoning can be, especially on an ontology as large and as complex as
Brick.  Before working on this PR, the reasoning time for our test cases was
easily in the hours to 10s of hours of time. This happend for several reasons:

1. The reasoner we are using --- the Python `owlrl` library --- adopts a brute
   force implementation.
2. The anonymous classes used to convey having a tag or not having a tag are
   redundant and numerous! This bogs down the reasoner

Good news is I've addressed these issues!

For Problem 2, I've stopped creating redundant anonymous classes for each of
the "having a tag" or "not having a tag" classes that are created in
`generate_brick.py`. This is accomplished by using RDFlib's super-helpful
`BNode()` class that takes an optional name argument. For each of the tags, we
create a single anonymous class that is equivalent to entities that have the
tag:


```turtle
# before, with random identifiers
_:abc123def a owl:Restriction ;
            owl:hasValue tag:Air ;
            owl:onProperty brick:hasTag .

# now, with consistent identifiers
_:has_Air a owl:Restriction ;
            owl:hasValue tag:Air ;
            owl:onProperty brick:hasTag .
```

This reduces the number of created anonymous classes by 10 or 20 thousand, and
reduces the inference time for `owlrl` from many hours to 1-2 minutes.
Although this addresses Problem 1, I've also added a `reason_agraph` method to
our `tests/` directory (in `util/reasoner.py`) that uses Allegrograph to
perform OWL-RL reasoning. You will need to run Docker and get the container
first:

```bash
$ docker pull franzinc/agraph
```

Allegrograph's reasoning is faster than `owlrl` and both seem to give the same
results (still checking this). It is currently the default reasoning method
used in the Brick unit tests.

gtfierro:
@jbkoh @shreyasnagare let me know what you think!

jbkoh:
Hi @gtfierro Thanks for putting all this together. I can feel your effort here.

Let me talk about the "Unsatisfiable Tag-based Inferences" first before double-check all the class changes later.

- Option 1. requires heuristics in Brick specification and I would like to avoid such things as much as possible. When we see overlapping tags between classes, what are the right tags to remove from which class? This is totally arbitrary (i.e., decided by a human designer) in the class definition for now and later, which could cause modeling inconsistency eventually.

- Option 1. loses necessary semantics. You have removed `Setpoint` tag for related `Parameters`. I'd argue we need to ge them back. Removing tags mainly causes two problems. Let's think about the `Max_Air_Flow_Setpoint_Parameter` example.
    - Imprecise definitions: This parameter defines what value the associated `Setpoint` can have at maximum. It may not be actual maximum value of the air flow depending on the control sequence. (A similar though different case is `Cooling_Temperature_Setpoint`. Actual temperature can be more than the setpoint.) So removing the tag is lossy.
    - Because of the lost semantics, we lose the possibility of verifying the semantics in instance models. For example, `Max_Air_Flow_Setpoint_Parameter` is associated with a `Air_Flow_Setpoint` (whose relation is currently `controls` though we can discuss it separately later), and such relation should be automatically verifiable in Brick (as we discussed in the NREL meeting.)

- Option 2. still requires heuristics similar to Option 1. A Brick designer should decide which one should be excluded as a rule in the inference process. It's better than Option 1. because we are not losing the semantics, but the design process is similar.

- It is unclear what logic exactly you are implementing for Option 2, but I'd probably reject the idea if it is too off from existing standards (i.e., if ours is the only implementation).

- It is starightforward to map tags to classes. We do not need OWL for the procedure at all. OWL is useful for another purpose which I will describe below. The mapping algorithm is simply like this:
    - Get all the tags of entities in a instance model with SPARQL.
    - Look up a tags->classes dictionary.
    This is basically what we do in OWL for mapping tags to classes anyway.

- Inferring classes from Tags with OWL is useful when we verify our class hierarchy. However, because of the problems we've discussed, I do not think it can be used as strict rules though it is still a useful indicator. 


I am open to (and expect) more discussions, but I am still with Option 3. as I proposed in the last December. And we can use OWL inference separately outside the release schema to inform ourselves whether the class hierarchy looks good or not.

jbkoh:
Btw, just FYI, OntoText's [GraphDB](https://www.ontotext.com/products/graphdb/) (a horrible name in my opinion) supports OWL2 RL/QL and RDFS reasoning. I haven't tested it out but heard a positive review.

JoelBender:
> Btw, just FYI, OntoText's [GraphDB](https://www.ontotext.com/products/graphdb/) (a horrible name in my opinion) supports OWL2 RL/QL and RDFS reasoning. I haven't tested it out but heard a positive review.

It's pretty good, just be aware that it won't [show blank nodes]9https://stackoverflow.com/questions/58454440/visual-graph-showing-blank-nodes).  When I create a graph I've been using my own `BlankNode` rather than `BNode` which is fine when building triples but if you use Turtle strings like Jinja templates it's a drag.

gtfierro:
I'll respond point by point below, but I will definitely put this on the agenda
for the Brick meeting next Friday.

> Option 1. requires heuristics in Brick specification and I would like to
> avoid such things as much as possible. When we see overlapping tags between
> classes, what are the right tags to remove from which class? This is totally
> arbitrary (i.e., decided by a human designer) in the class definition for now
> and later, which could cause modeling inconsistency eventually.

Inspecting the class hierarchy, I think its fairly straightforward which tags
get removed in most cases. Recall that most of the conflicts are due to the
base point tags overlapping (Setpoint, Sensor, Alarm, etc). With our unit tests
and with sufficient documentation, I don't think that the choice of tags is
arbitrary *or* inconsistent. Remember: we get to define the mapping between
tags and classes, so its not arbitrary at all.

> Option 1. loses necessary semantics....Imprecise definitions:

Can you clarify what you mean by the definitions being imprecise? I did a lot
of work to make sure that they *are* precise (because they are defined by the
mapping). I think you mean that its just not the set of tags you would have
chosen? In any case, I don't buy the argument that semantics are "lost" because
the semantics are given by the mapping from the tags to the classes and the
definitions of the classes. This is also independent of the relationship
between `Air_Flow_Setpoint_Parameter` and `Air_Flow_Setpoint`, so I'm not sure
what you mean there either.

Removing the tags is not lossy because you always have the mapping to the Brick
class which carries the actual definition. 

> Option 2. still requires heuristics similar to Option 1. A Brick designer
> should decide which one should be excluded as a rule in the inference
> process. It's better than Option 1. because we are not losing the semantics,
> but the design process is similar. 

We're not "losing semantics" in either case. The rules of which tags can appear
in which hierarchy vs another are rules that we would have to implement
regardless of the mechanism of mapping tags to classes. At least this way there
is a formal definition rather than ad-hoc logic.

> It is unclear what logic exactly you are implementing for Option 2, but I'd
> probably reject the idea if it is too off from existing standards (i.e., if
> ours is the only implementation).

The logic is in the code and the `Brick.ttl` file and is documented in the text
I wrote above: I'm using `owl:complementOf` to define the property "does not
have tag X", which is enough to separate the groups of tags among hierarchies.
It all uses standard OWL 2 properties; what do you mean "if ours is the only
implementation"?

> It is straightforward to map tags to classes. We do not need OWL for the
> procedure at all. OWL is useful for another purpose which I will describe
> below. The mapping algorithm is simply like this:

Your algorithm doesn't actually solve the issue when subsets of tags are shared
across hierarchies. You can use this approach as a "first pass" (which is what
I describe in the Brick+ paper) and then apply OWL reasoning afterwards, but I
don't like splitting the inference logic across two different systems like
this.

> Btw, just FYI, OntoText's GraphDB (a horrible name in my opinion) supports
> OWL2 RL/QL and RDFS reasoning. I haven't tested it out but heard a positive
> review.

Why is GraphDB a bad name? Its *is* a graph database after all. We make heavy
use of blank nodes so it may not be the best choice. In any case, Allegrograph
and OWL-RL (python library) have been working great

jbkoh:
> I'll respond point by point below, but I will definitely put this on the agenda
> for the Brick meeting next Friday.
> 
> > Option 1. requires heuristics in Brick specification and I would like to
> > avoid such things as much as possible. When we see overlapping tags between
> > classes, what are the right tags to remove from which class? This is totally
> > arbitrary (i.e., decided by a human designer) in the class definition for now
> > and later, which could cause modeling inconsistency eventually.
> 
> Inspecting the class hierarchy, I think its fairly straightforward which tags
> get removed in most cases. Recall that most of the conflicts are due to the
> base point tags overlapping (Setpoint, Sensor, Alarm, etc). With our unit tests
> and with sufficient documentation, I don't think that the choice of tags is
> arbitrary _or_ inconsistent. Remember: we get to define the mapping between
> tags and classes, so its not arbitrary at all.
> 
Once we define the mapping between tags and classes, it would be not arbitrary. However, the choice of tags to be removed at the design time is arbitrary. E.g., Setpoint is removed from some Parameter classes because it's overlapped with other classes but not because it is not necessary. So the choice of tags is determined by other classes and whether the choice is correct or not is decided by the schema designer (== us). It means the schema design becomes more subjective to our choice instead of comprehensively representing classes by tags.
> > Option 1. loses necessary semantics....Imprecise definitions:
> 
> Can you clarify what you mean by the definitions being imprecise? I did a lot
> of work to make sure that they _are_ precise (because they are defined by the
> mapping). I think you mean that its just not the set of tags you would have
> chosen? In any case, I don't buy the argument that semantics are "lost" because
> the semantics are given by the mapping from the tags to the classes and the
> definitions of the classes. This is also independent of the relationship
> between `Air_Flow_Setpoint_Parameter` and `Air_Flow_Setpoint`, so I'm not sure
> what you mean there either.
> 
> Removing the tags is not lossy because you always have the mapping to the Brick
> class which carries the actual definition.
I mean it's imprecise by not representing necessary information when we remove tags. The class hierarchy and the mapping between tags and classes will be consistent as we made that way. However, some useful information associated with classes would be missing. E.g., Setpoint tag in the associated parameters means that the parameters are associated with specific Setpoints. We are arbitrarily removing such Tags just to keep the superset/subset relations across the hierarchy. So the meanings of classes represented by tags becomes imprecise. Of course, the meaning of Classes themselves could remain same if we document them. And there is no meaning of using tags to represent Classes other than just mapping a set of tags to a class.
> 
> > Option 2. still requires heuristics similar to Option 1. A Brick designer
> > should decide which one should be excluded as a rule in the inference
> > process. It's better than Option 1. because we are not losing the semantics,
> > but the design process is similar.
> 
> We're not "losing semantics" in either case. The rules of which tags can appear
> in which hierarchy vs another are rules that we would have to implement
> regardless of the mechanism of mapping tags to classes. At least this way there
> is a formal definition rather than ad-hoc logic.
Please refer to the other items for this argument. I am saying it is not right to remove some tags from some classes because the tags are overlapped with other classes. The tags should be removed when they do not mean anything for the classes but not in other cases.
> 
> > It is unclear what logic exactly you are implementing for Option 2, but I'd
> > probably reject the idea if it is too off from existing standards (i.e., if
> > ours is the only implementation).
> 
> The logic is in the code and the `Brick.ttl` file and is documented in the text
> I wrote above: I'm using `owl:complementOf` to define the property "does not
> have tag X", which is enough to separate the groups of tags among hierarchies.
> It all uses standard OWL 2 properties; what do you mean "if ours is the only
> implementation"?
I thought you kept the overlapping tags between the classes of our interest and implemented another logic to correctly infer their classes. Instead, you implemented logic to verify their negative relationships, right? Please correct me if I am wrong. If my understanding is right, the negation relationships are useful for verifying the class hierarchy, but does not resolve overlapping tags problem, does it?
> 
> > It is straightforward to map tags to classes. We do not need OWL for the
> > procedure at all. OWL is useful for another purpose which I will describe
> > below. The mapping algorithm is simply like this:
> 
> Your algorithm doesn't actually solve the issue when subsets of tags are shared
> across hierarchies. You can use this approach as a "first pass" (which is what
> I describe in the Brick+ paper) and then apply OWL reasoning afterwards, but I
> don't like splitting the inference logic across two different systems like
> this.
There are two purposes of formalizing relations of tags<->classes. The first one is just to ease the usage of classes by tags, which one-to-one mapping from sets of tags to classes can achieve. The other one is to verify whether the class hierarchy is correct or not based on the semantics represented by tags.

The simple algorithm I mentioned achieves the first one because we just need one-to-one mapping for that.

For the second purpose, we need a reasoning engine to calculate the subsumption relations, for which we definitely need an OWL engine. However, it is only useful when the sets of tags comprehensively/correctly describe the semantics of classes, which I argue is not due to the reasons I described in the above issues. Thus, we should make an agreement on whether tags can properly represent classes' semantics. A subset of the issue is whether removing tags from some classes is fine even though they mean something for the classes (as detailed in the above bullet points.) You know, we found these issues after we actually verified the hierarchy thanks to the paper.
> 
> > Btw, just FYI, OntoText's GraphDB (a horrible name in my opinion) supports
> > OWL2 RL/QL and RDFS reasoning. I haven't tested it out but heard a positive
> > review.
> 
> Why is GraphDB a bad name? Its _is_ a graph database after all. We make heavy
> use of blank nodes so it may not be the best choice. In any case, Allegrograph
> and OWL-RL (python library) have been working great
nvm. I just think GraphDB is too generic. It's like naming something as KeyValueStore or RelationalDatabase, etc.

gtfierro:
> Once we define the mapping between tags and classes, it would be not
> arbitrary. However, the choice of tags to be removed at the design time is
> arbitrary. E.g., Setpoint is removed from some Parameter classes because it's
> overlapped with other classes but not because it is not necessary. So the
> choice of tags is determined by other classes and whether the choice is
> correct or not is decided by the schema designer (== us). It means the schema
> design becomes more subjective to our choice instead of comprehensively
> representing classes by tags.

This I agree with; it would be nice to include the Setpoint tags where one would expect them.

> I thought you kept the overlapping tags between the classes of our interest
> and implemented another logic to correctly infer their classes. Instead, you
> implemented logic to verify their negative relationships, right? Please
> correct me if I am wrong. If my understanding is right, the negation
> relationships are useful for verifying the class hierarchy, but does not
> resolve overlapping tags problem, does it?

You should really check out the implementation in `Brick.ttl` to get an idea of
how I was trying to handle it. The gist of the approach relies on the sets of
tags for each disjoint point class (Setpoint, Sensor, etc) being different,
though they can be subsets.  Setpoints are those entities that have the
Setpoint tag and *not* the Parmaeter tag; Parameters are those entities that
have the Parameter tag; etc. This requires the use of negative property
assumptions. For example, for `Max_Air_Flow_Setpoint_Limit`, we could have the
requirement that this class is the intersection of all individuals who have the
`Max`, `Air`, `Flow`, `Setpoint`, `Limit` and `Parameter` tags.
`Air_Flow_Setpoint` would be the intersection of all individuals who have the
`Air` tag, `Setpoint` tag, `Flow` tag and *not* the `Parameter` tag.

Basically, the only way we can express negative property assertions in OWL is
through the use of complements. The issue with this, under OWL's open world
assumption, is that it is difficult to make the statement that an entity has
*only* the tags that appear in the graph and no other tags could exist. This
sounds a little silly at first, but it is the root cause of these modeling
difficulties.

There are some variations of description logics out there that enforce local
closed world assumptions that would make our job easier, but I haven't had time
to really dig into those. I've been playing around with different techniques of
making the `brick:hasTag` property a "closed" property, but they are either
very awkward or don't quite work. I think there might still be a way of doing
it, but in the interest of getting Brick 1.1 released, we can adopt a
non-formal mapping of tags to classes at least until I can figure out a better
way of doing the formal mapping (which I think is ultimately the correct
solution to this problem).

I'm working on a Python package for Brick that handles the paper's inference
algorithm (among other things), and I'll make a PR for that in the next day or
so. After that, I can use the package to adjust the test cases as well; look
for an update on this PR in the next day or so.

jbkoh:
> You should really check out the implementation in `Brick.ttl` to get an idea of
> how I was trying to handle it. The gist of the approach relies on the sets of
> tags for each disjoint point class (Setpoint, Sensor, etc) being different,
> though they can be subsets. Setpoints are those entities that have the
> Setpoint tag and _not_ the Parmaeter tag; Parameters are those entities that
> have the Parameter tag; etc. This requires the use of negative property
> assumptions. For example, for `Max_Air_Flow_Setpoint_Limit`, we could have the
> requirement that this class is the intersection of all individuals who have the
> `Max`, `Air`, `Flow`, `Setpoint`, `Limit` and `Parameter` tags.
> `Air_Flow_Setpoint` would be the intersection of all individuals who have the
> `Air` tag, `Setpoint` tag, `Flow` tag and _not_ the `Parameter` tag.
> 
> Basically, the only way we can express negative property assertions in OWL is
> through the use of complements. The issue with this, under OWL's open world
> assumption, is that it is difficult to make the statement that an entity has
> _only_ the tags that appear in the graph and no other tags could exist. This
> sounds a little silly at first, but it is the root cause of these modeling
> difficulties.
> 
> There are some variations of description logics out there that enforce local
> closed world assumptions that would make our job easier, but I haven't had time
> to really dig into those. I've been playing around with different techniques of
> making the `brick:hasTag` property a "closed" property, but they are either
> very awkward or don't quite work. I think there might still be a way of doing
> it, but in the interest of getting Brick 1.1 released, we can adopt a
> non-formal mapping of tags to classes at least until I can figure out a better
> way of doing the formal mapping (which I think is ultimately the correct
> solution to this problem).
> 
> I'm working on a Python package for Brick that handles the paper's inference
> algorithm (among other things), and I'll make a PR for that in the next day or
> so. After that, I can use the package to adjust the test cases as well; look
> for an update on this PR in the next day or so.

- I've looked at Brick.ttl but couldn't locate where exactly the Option 2. was implemented.
    - `Temperature_Adjust_Sensor` is not in the file though you mentioned it as an example.
    - `owl:complementOf` is only observable in `Brick.ttl`
    - `Max_Air_Flow_Setpoint_Limit` is I think an unimplemented example?
    - So, it's hard to grasp how you implemented Option 2. in Brick.ttl. I'd appreciate if you could point in the corresponding lines in the schema file.
- I am fine with the current version as 1.1 release (after another pass of verifying the class defs, e.g., `Temperature_Adjust_Sensor` is not defined). However, IF, in the future, we'd conclude OWL would not be the framework for inferring classes from tags because of the reasons we discussed, I would like to make clear to users that OWL tagging rules are not a part that they primarily need to care about. That's why I wanted to separate out the tag mapping from the standard file. We could first release the schema which we have verified many times and then keep updating the inference procedure separately. We can discuss the pros and cons at the next meeting.

gtfierro:
I've made a couple changes to the Brick modeling over the past couple days
to get around the issues we've been discussing. I do want to reiterate
that I would prefer a formal mapping of tags to classes, but I think this 
involves investment in a LCWA-based model (or some other method of making
the `brick:hasTag` property closed), so I'm leaving this for future work 
(I'll probably look into it after this PR is merged).

**Change #1**: Brick classes have a new property `brick:hasAssociatedTag` that
points towards a set of tags. This annotation is there for any external processes
that want to use the class -> tag mapping. `brick:hasAssociatedTag` has an inverse
property `brick:isAssociatedWith` that points from tags to classes. No OWL-based
inference of tags or classes is performed using this property.

```turtle
# partial class definition
brick:Max_Air_Flow_Setpoint_Limit a owl:Class ;
    brick:hasAssociatedTag tag:Air,
        tag:Flow,
        tag:Limit,
        tag:Max,
        tag:Parameter,
        tag:Setpoint .
```

**Change #2**: In order to provide tags -> classes inference, I've wrapped up the logic
from the House of Sticks paper (BuildSys 2019) into a [Python package](https://github.com/BrickSchema/py-brickschema)
named `brickschema`.  The package has plenty of other functionality, but here's an
example of how to do tag-to-class inference:


1. For a sample entity modeled with tags:

    ```turtle
    # myfile.ttl
    @prefix brick: <https://brickschema.org/schema/1.1.0/Brick#> .
    @prefix tag: <https://brickschema.org/schema/1.1.0/BrickTag#> .
    @prefix bldg: <https://example.org/example#> .

    bldg:my_entity  brick:hasTag    tag:Air, tag:Flow, tag:Setpoint .
    ```

2. Load the graph into a `brickschema.Graph` and run the TagInferenceSession:
    ```python
    import brickschema
    g = brickschema.graph.Graph()
    g.load_file("myfile.ttl")
    g = brickschema.inference.TagInferenceSession(approximate=False).expand(g)
    
    print(g.query("SELECT ?type WHERE { bldg:my_entity a ?type }"))
    # => ['Brick:Air_Flow_Setpoint']
    ```

**Change #3**: In order to provide class -> tag inference, each of the Brick classes
is a *subclass* of the intersection of its tags:

```turtle
brick:Max_Air_Flow_Setpoint_Limit a owl:Class ;
    rdfs:label "Max Air Flow Setpoint Limit" ;
    rdfs:subClassOf [ owl:intersectionOf ( _:has_Max _:has_Air _:has_Flow _:has_Limit _:has_Parameter _:has_Setpoint ) ],
        brick:Air_Flow_Setpoint_Limit,
        brick:Max_Limit ;
    brick:hasAssociatedTag tag:Air,
        tag:Flow,
        tag:Limit,
        tag:Max,
        tag:Parameter,
        tag:Setpoint .
```

Each of the `_:has_{tag}` classes is an anonymous class equivalent to entities that have the tag:

```turtle
_:has_Max a owl:Restriction ;
    owl:hasValue tag:Max ;
    owl:onProperty brick:hasTag .
```

The subclass-based modeling means that if an entity is an instance of the Brick class, then it will "inherit"
the intersection class, and thus the tags associated with each of the intersection classes.

gtfierro:
Following up from the call today:

@jbkoh let me know what you think of the modeling + implementation
@shreyasnagare can you pull in the Brick.ttl into your process and figure out what is missing?

We should be really close to merging this (finally) :)

shreyasnagare:
> @shreyasnagare can you pull in the Brick.ttl into your process and figure out what is missing?

Here's a list of a few more missing classes:
- Acceleration_Time_Setpoint
- Discharge_Air_Static_Pressure_Deadband_Parameter
- Discharge_Air_Static_Pressure_Integral_Time_Setpoint
- Discharge_Air_Temperature_Reset_Differential_Setpoint
- Discharge_Air_Temperature_Reset_High_Setpoint
- Discharge_Air_Temperature_Reset_Low_Setpoint
- Min_Damper_Position_Setpoint_Limit
- Discharge_Air_Temperature_Deadband_Parameter
- Emergency_Power_Off_System
- Enable_Hot_Water_System_Outside_Air_Temperature_Parameter
- Exhaust_Air_Stack_Flow_Deadband_Parameter
- Medium_Temperature_Hot_Water_Supply_Temperature_High_Reset_Setpoint
- Medium_Temperature_Hot_Water_Supply_Temperature_Low_Reset_Setpoint
- Close_Limit_Parameter

These classes come from the  ["mappings" sheet](https://docs.google.com/spreadsheets/d/1ZUEe1bouutdAvRAqU-Jq3IjEPGSH9aoRBXIK-t5MXb4/edit#gid=1583326302). I have updated the list of valid classes to have all the new changes and we should be able to merge this PR once we take care of all the data validation errors in column B.

gtfierro:
Thanks for the prompt reply, @shreyasnagare !

I've pushed changes to Brick as well as the spreadsheet:


- Acceleration_Time_Setpoint: added
- these were omitted due to a typo:
    - Discharge_Air_Temperature_Reset_Differential_Setpoint
    - Discharge_Air_Temperature_Reset_High_Setpoint
    - Discharge_Air_Temperature_Reset_Low_Setpoint
    - Medium_Temperature_Hot_Water_Supply_Temperature_High_Reset_Setpoint
    - Medium_Temperature_Hot_Water_Supply_Temperature_Low_Reset_Setpoint
- `Deadband` is a setpoint, not a param. Fixed this in the spreadsheet:
    - Discharge_Air_Static_Pressure_Deadband_Parameter
    - Exhaust_Air_Stack_Flow_Deadband_Parameter
    - Discharge_Air_Temperature_Deadband_Parameter
- `Integral Time` are parameters:
    - Discharge_Air_Static_Pressure_Integral_Time_Setpoint
- Min_Damper_Position_Setpoint_Limit: Min_Position_Setpoint_Limit
- this is a setpoint: Enable_Hot_Water_System_Outside_Air_Temperature_Parameter
- Close_Limit_Parameter: named "Close_Limit"

Only one I'm not sure how to add is `Emergency_Power_Off_System`; @jbkoh how do you want to model this?

jbkoh:
A quick note before myself diving into the modeling framework.

@gtfierro 
1. I believe `Integral_Time` is a parameter used inside a PID loop. What's is it changed to a `Setpoint`?
2. It'd be useful to have textual definitions of Setpoint and Parameter in this release so that people understand the difference.
3. In the current hierarchy, `Discharge_Air_Static_Pressure_Deadband_Setpoint` is a subclass of `Discharge_Air_Static_Pressure_Setpoint` where their meanings diverge. `Discharge_Air_Static_Pressure_Deadband_Setpoint` should be strictly a subclass `Deadband_Setpoint` subtree. If there are similar relationships like this, we have to fix it too.
4. `Emergency_Power_Off_System` seems not falling into any existing equipment classes. We might want to subclass `Electrical_Systems` but I'd rather start with a top equipment class before we are sure about it. Emergency_Power_Off_System with Emergency, Power, Off, System, and Equipment should be fine to me.
5. Actually, 4. reminds me of another question I had before. @gtfierro why did you remove `System` tag to all the system classes?

bricksrc/equipment.py
```diff
@@ -1,28 +1,97 @@
...
+                },
+            },
+            "Inverter": {
+                "tags": [TAG.Inverter, TAG.Equipment],
+            },
+            "Power_System": {
+                "tags": [TAG.Power, TAG.Equipment],
+            },
+            "PlugStrip": {
+                "tags": [TAG.Plugstrip, TAG.Equipment],
```
jbkoh: Typo: `TAG.Plugstrip` -> `TAG.PlugStrip`

bricksrc/sensor.py
```diff
@@ -255,379 +266,469 @@
...
+                SKOS.definition: Literal("senses or detects gas concentration (other than CO2)"),
+                "tags": [TAG.Sensor, TAG.Gas],
             },
             "Hail_Sensor": {
-                "tags": [ TAG.Sensor, TAG.Hail ],
-                "substances": [ [ BRICK.measures, BRICK.Hail ], ],
+                "tags": [TAG.Sensor, TAG.Hail],
+                "substances": [[BRICK.measures, BRICK.Hail]],
+            },
+            "Heat_Sensor": {
```
jbkoh: What's heat sensor? heating power? thermal energy? temperature?

bricksrc/equipment.py
gtfierro: Good catch; I'll standardize on "PlugStrip"

bricksrc/sensor.py
jbkoh: Let's leave it as it is for now as it is used in practice though we don't have a clear def.

jbkoh:
Summarizing the call, we are all good to go except issues need bigger discussions in separate threads.

Remaining issues are:
- Is `Discharge_Air_Static_Pressure_Deadband_Setpoint` a subclass of `Discharge_Air_Static_Pressure_Setpoint`?
- Separate out system classes like `Water_System` as separate classes from the current subtrees.

gtfierro:
22 days later, we're ready to go! Thanks for the feedback everyone

PR #95, opened 2020-01-15
Fix incomplete class defs

jbkoh:
Add Tags for Classes without Tags. Augment the generation script to display untagged Classes.

There are remaining issues that are out of scope in this PR. I suggest to accept this PR first and fix the remaining issues in other PRs.

The remaining issues are
- `Water_Pump` is inferred as a subclass of `Water_System`. We should discuss whether this is correct or not though I think it's correct.
- The inference engine should infer equivalent classes.
- Overlapping tags between classes causes incorrect inferences.

### DIFF
--- a/generate_brick.py
+++ b/generate_brick.py
@@ -1,4 +1,6 @@
 from collections import defaultdict
+import warnings
+
 from rdflib import Graph, Literal, BNode, URIRef
 from rdflib.namespace import XSD
 from rdflib.collection import Collection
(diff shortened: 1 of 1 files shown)


gtfierro:
I'm in the middle of a major set of fixes in https://github.com/BrickSchema/Brick/pull/94. My guess is it will probably address a number of these fixes that you're doing, so do you mind if we investigate this after I finish? There's  tag/class/inference changes wrapped up in the work so I'm trying to make sure its complete and that the unit tests are appropriate. One good piece of news is I think I've solved our inference issues

jbkoh:
Close this one so that I can separate out this PR into two.

PR #96, opened 2020-01-22
Brickschema Python package

gtfierro:
In working on Brick v1.1, I found I often wanted a set of utilities to make working with the graphs easier. Over time, that set of tools grew into this small Python package, which wraps several features:
- simplified interface to RDFlib + associated facilities under the assumption you are working with Brick
- Haystack-to-Brick inference, OWL inference, tag inference all in the same package
- basic SQL-based ORM for those who want a different view of Brick (in progress)
-  I hope to add some GraphQL stuff soon as well

You can install it with pip:

```
pip install brickschema
```

Some basic usage at the [pypi page](https://pypi.org/project/brickschema/)

Currently at version 0.0.1. Docs are not hosted online (yet!!) but you can build them with `make docs` in the `brickschema/` directory. I'll host them online after this is merged into master

I've put in some real effort to make sure this is all documented and tested. Even though this package is not perfect or complete, I hope we can merge this soon because I want to use the tag inference capabilities in the test-cases for finishing up #94


gtfierro:
On second thought, this might do better in its own repository. I'll just do that!

PR #97, opened 2020-01-28
SHACL work for Brick

xyzisinus:
This PR is a continuation of [PR 85](https://github.com/BrickSchema/Brick/pull/85).  It encompasses the following items:

1. A script generating shapes using Brick properties and its accompany test cases.
2. A thin wrapper of pyshacl finding offending triples reported as constraint violations.

Special notes:

1. rdfs:domain and rdfs:range are temporarily replaced with brick:expectedDomain and brick:expectedRange as a workaround for lacking of disjointness enforcement.  However I choose not to modify the Brick.ttl file at the root of the repo.  Instead, a Brick.ttl is generated with above said properties and used only as the ontology file for pyshacl.
2. The substance_subproperties in Gabe's generate_shacl.py has been moved into properties.py as a sub-property of brick:feeds, expressed as brick:hasSubproperty.  This is a modification to Brick Schema -- needing special attention by the reviewer.


xyzisinus:
I see that quite a few tests have failed.  Should I help fix them?

gtfierro:
@xyzisinus I'll take a look at this tomorrow. Make sure you merge from the `master` branch to take advantage of the recent PRs

xyzisinus:
> The Brick URI changed (`1.1.0` -> `1.1`) which may affect the tests.
I did a search in the master branch of the master repo.  All I can see is brick version 1.1.0, including BRICK_VERSION in namespace.py.

gtfierro:
Ah, right. We haven't merged that one yet. I would take a closer look to see what changed when you merged

gtfierro:
I've since merged in #92 ; merge that in and see if that has fixed the test issues

xyzisinus:
tests/test_quantities.py::test_measurables_defined FAILED

Much better.  This the only failure now.

xyzisinus:
I find the particular query has something to do with subClassOf, it happens to be related to the new property I added
```
   "hasSubproperty": {
	A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
        OWL.inverseOf: RDFS.subPropertyOf,
        BRICK.expectedRange: RDFS.Property,
	BRICK.expectedDomain: RDFS.Property,
        SKOS.definition: Literal("Subject is a property that has the object as its sub-properity")
    },
```

bricksrc/properties.py
```diff
@@ -5,27 +5,50 @@
 Defining properties
 """
 properties = {
+    # About expectedDomain and expectedRange: They replace temporarily
+    # RDFS.domin and RDFS.range respectively.  The .domain and .range
+    # properties confuse pre-inferencing of a SHACL validator without
+    # respecting disjointness.  Thus we want to remove them for now
+    # but still need the replacements to guide the shape generator.
```
gtfierro: Its good to have this explanation here, but I would amend it slightly.

`rdfs:domain` and `rdfs:range` infer the type of the subject and object of a property, respectively. However, rather than imply the type of the subject or object by the use of the property, we want to verify that the property is used correctly given the types of the subject and object. Thus, we adopt the `brick:expectedDomain` and `brick:expectedRange` properties which imply nothing about the subject/object of the property they are annotating, but can be used to validate the use of Brick relationships.

shacl/DESIGN.md
```diff
@@ -0,0 +1,111 @@
+# SHACL for Brick
+
+The SHACL Shapes Constraint Language validates RDF graphs against conditions (called "shapes"), which are expressed themselves as an RDF graph. This document discusses the application of SHACL to Brick. In this document we are specifically looking at writing SHACL shapes for validating Brick *models* (instances of buildings). Using SHACL to validate the Brick ontology deserves its own investigation.
+
+SHACL shapes for Brick should validate the following:
+
+- validating endpoints of Brick relationships (properties):
+  - check type of subject/object: are the `rdfs:range`, `rdfs:domain` properties respected?
```
gtfierro: Change `rdfs:range` to `brick:expectedRange` and do the same for `rdfs:domain`

shacl/DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+To validate a graph, use [pySHACL](https://github.com/RDFLib/pySHACL):
+
+```bash
+pyshacl -s shacl_test.ttl -m -e Brick.ttl -f human sample_graph.ttl
+```
+
+You may want to expand a graph before validating it. In that case, run
+
+```bash
+python expand_graph.py sample_graph.ttl
```
gtfierro: We should probably remove this because the `expand_graph.py` script is gone; the inference should be handled by `pyshacl` directly if it is needed

shacl/DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+```
+
+which will drop the expanded graph into `output.ttl` in the same directory.
```
gtfierro: Drop this as well (part of the above section)

shacl/DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+**TODO**: so far, I'm only sure how to encode the `rdfs:range` property
+
+For example, the definition of `brick:hasLocation` is
+
+```python
+{
+  "hasLocation": {
+    A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+    OWL.inverseOf: "isLocationOf",
+    RDFS.range: BRICK.Location,
```
gtfierro: Remember to change the `domain/range` stuff to `expected{Domain, Range}`

shacl/DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+    SKOS.definition: Literal("Subject is physically located in the location given by the object"),
+  },
+}
+```
+
+This definition informs the following Shape for the `rdfs:range` property
+
+```ttl
```
gtfierro: Is this shape up to date with what we decided on?

shacl/expand_graph.py
```diff
@@ -0,0 +1,20 @@
...
+G = Graph()
+bind_prefixes(G)
+A = RDF.type
+G.parse(sys.argv[1], format='ttl')
+#G.parse('Brick.ttl', format='ttl')
+
+reason_brick(G)
+
+with open('output.ttl', 'wb') as f:
+    f.write(G.serialize(format='ttl'))
```
gtfierro: Do we still need this file? Reasoning can be handled by the internal pySHACL reasoner if we need it

shacl/generate_shacl.py
```diff
@@ -0,0 +1,60 @@
...
+            G.add((BSH[shapename], SH.targetSubjectsOf, BRICK[name]))
+            G.add((sh_prop, SH['message'],
+                   Literal(f"Property {name} has object with incorrect type")))
+
+        if pred == BRICK.expectedDomain:
+            shapename = f"{name}DomainShape"
+            G.add( (BSH[shapename], A, SH.NodeShape))
+            G.add( (BSH[shapename], SH.targetSubjectsOf, BRICK[name]))
+            G.add( (BSH[shapename], SH['class'], obj))
+            G.add( (BSH[shapename], SH['message'],
```
gtfierro: To keep style consistent, remove the leading space in front of the argument to `G.add()`

shacl/validate.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
```
gtfierro: This should probably be a Makefile rather than a shell script. Follow the example of the Makefile in the root of this repository to avoid regenerating Brick or SHACL shapes when the source files haven't changed

shacl/offendingTriples.py
```diff
@@ -0,0 +1,177 @@
...
+
+import argparse
+from rdflib import Graph, Namespace, URIRef, Literal
+from rdflib.plugins.sparql import prepareQuery
+import sys
+
+sys.path.append('..')
+from bricksrc.namespaces import RDF, RDFS, BRICK, BSH, SH, SKOS
+
+global result_graph
```
gtfierro: Can you refactor this to not use global variables? I think this could do nicely as a class that wraps up the logic

xyzisinus:
I ran the ci tests against master and czang-shacl in my repo.  The tests in the masster branch all passed.  So the failure in czang-shacl is likely caused by property changes.

tools/addPrefix/make_prefixable.py
```diff
@@ -0,0 +1,63 @@
+from argparse import ArgumentParser
```
gtfierro: Why is the `make_prefixable` script included in this PR? Are you making use of it in the SHACL work? If not, we might want to separate this out into another PR so we can discuss it directly. The building models are going to be changed by another PR that updates them to the latest version of the ontology.

xyzisinus:
Good catch.  I didn't think about DESIGN.md at all.  Thanks for the very thorough review.

tools/addPrefix/make_prefixable.py
xyzisinus: Oh no I meant to move this set of files into a different branch.

shacl/offendingTriples.py
xyzisinus: Done

shacl/validate.sh
xyzisinus: I thought about a makefile but I generally don't like make's way of passing arguments arg=xxx.  It's clumsy for the user (not auto-completion).  Since generating Brick.ttl and shacl_test.ttl is pretty fast (pyshacl is the slow one), I'd say we leave it like this until we find the speed intolerable.

shacl/generate_shacl.py
xyzisinus: done

shacl/expand_graph.py
xyzisinus: Deleted.

bricksrc/properties.py
xyzisinus: Please see how it looks now.

xyzisinus:
The tests have passed.  That makes me feel uneasy:   It consistently failed yesterday.  There are no obvious changes in between that should affect the outcome of the tests.

xyzisinus:
Should have pushed button to request review again days ago.

gtfierro:
No worries; I'm super busy today but I'll try to get to the review tonight. The tests pass but something in the `generate_brick.py` script is breaking the build. Can you look into that and see what's going on? Look for any changes in `bricksrc/` or `generate_brick.py` that deviate from what's in `master`

gtfierro:
Actually, what is probably breaking the build is when you use strings as values in `properties.py`. You should have

```python
 OWL.inverseOf: BRICK.hasLocation,
```

instead of

```python
 OWL.inverseOf: "hasLocation"
```

One of the recent PRs changed the logic so that the value-side needs to be a valid Literal or URIRef, for consistency

xyzisinus:
> ```python
>  OWL.inverseOf: "hasLocation"
> ```

I'm puzzled -- I don't have that line in the [file](https://github.com/xyzisinus/Brick/blob/czang-shacl/bricksrc/properties.py).

But I notice that in the master of Brick repo, the notation for the brick terms has changed from `BRICK.hasLocation` to `BRICK["hasLocation"]`.  I thought that's some global decicion and I'm going through the file to convert things like `BRICK.expectedDomain` into `BRICK["expectedDomain"]`.  Is that what I should be doing?

gtfierro:
`BRICK.hasLocation` and `BRICK["hasLocation"]` are equivalent. I prefer the latter, but it shouldn't make a difference from a correctness standpoint. I would still check that the value-side is a valid Literal or URIRef, as that is what is suggested by the build errors

xyzisinus:
> `BRICK.hasLocation` and `BRICK["hasLocation"]` are equivalent. I prefer the latter, but it shouldn't make a difference from a correctness standpoint. I would still check that the value-side is a valid Literal or URIRef, as that is what is suggested by the build errors
```
       "subproperties": {
            "feedsAir": {
                SKOS.definition: Literal("Passes air"),
                'substance': BRICK["Air"],    <<<< error 
                'properties': [BRICK.regulates, BRICK.measures]
            },
        },
```
This the offending part.  I have a print before the failed line in generate_brick.py

gtfierro:
Oh, bizarre. Why is the `'substance'` key in there? That's not an accepted key in the Brick generation process

xyzisinus:
> Oh, bizarre. Why is the `'substance'` key in there? That's not an accepted key in the Brick generation process

I added this part.  It's was hard coded in generate_shacl.py.  I moved it here.  Is that the right thing to do?

gtfierro:
The files used by `generate_brick.py`  have a certain structure that, while we can expand, does not accept keys/values that don't adhere to the expected types. Leave the "substance" annotation in `generate_shacl.py` for now (or eliminate it all together if you don't need it). After you remove it from `generate_brick.py`, the tests should start passing

shacl/DESIGN.md
```diff
@@ -0,0 +1,92 @@
...
+* generate ontology file Brick.ttl
+* generate shapes grapph shacl_test.ttl
+* execute a wrapper of the `pyshacl` tool to validate the building graph and report violations
+
+As an example, try `./validate.sh sample_graph.ttl`.
+
+## Brick SHACL shapes
+
+Those are the rules focusing on certain type of triples in the building graph and enforcing restrictions.  The
+examples below declare shapes in the `bsh` ("brickshape") namespace for clarity.
```
gtfierro: Can you include the full URI of the `bsh` namespace as well, for clarity?

shacl/generate_shacl.py
```diff
@@ -0,0 +1,60 @@
...
+
+for name, properties in property_definitions.items():
+
+    for pred, obj in properties.items():
+        sh_prop = BNode()
+
+        # We are only concerned with properties that have RDFS.domain or .range
+        # predicate.  The temporary replacements for those predicates are
+        # brick:expectedDomain and :expectedRange.  See properties.py for
+        # explanation.
```
gtfierro: Remember to change `RDFS.domain` to `BRICK.expectedDomain` and so on for `range`

xyzisinus: Yeah,  two more properties have crept in this week.  Playing the catch up work...

shacl/DESIGN.md
xyzisinus: Yes

shacl/DESIGN.md
xyzisinus: done

shacl/DESIGN.md
xyzisinus: done

bricksrc/properties.py
```diff
@@ -41,24 +72,28 @@
...
-                # Restriction (onProperty=brick:hasOutputSubstance, hasValue=brick:Air)
-
-                # looks something like this
-                #"domain_value_prop": [
-                #    [BRICK.hasTag, TAG.Air],
-                #],
-                #"range_value_prop": [
-                #    [BRICK.hasTag, TAG.Air],
-                #],
+                'substance': BRICK.Air,
```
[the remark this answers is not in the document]
xyzisinus: Changed to
```
"subproperties": {
            "feedsAir": {
		SKOS.definition: Literal("Passes air")
                BRICK.expectedRange: BRICK.Air,
            },
        },
```
Will change `generate_shacl.py` accordingly to say: Anything that feedsAir should
1.  regulates Air
2. hasOutputSubstance Air
That change will be done on separately.

bricksrc/properties.py
```diff
@@ -5,6 +5,36 @@
 Defining properties
 """
 properties = {
+    # About expectedDomain and expectedRange: They replace temporarily
+    # RDFS.domin and RDFS.range. The RDFS.domain and RDFS.range infer
+    # the type of the subject and object of a property, respectively.
+    # However the type inference can lead to the root class of the types,
+    # making all types compatible with each other, thus rendering type
+    # contraint using SHACL ineffective.  This problem should be solved
```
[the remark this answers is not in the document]
xyzisinus: done

xyzisinus:
> * [ ]  SHACL shape and unit test for each Brick property
> * [ ]  examples of SHACL shapes (with lower severity) for an "equipment template". What I mean by this is a shape that encapsulates the expected points (and other properties) for a given type of equipment. For example, a VAV shape for a particular building might require that the VAV has a Damper (through `hasPart`), a temperature sensor and setpoint (through `hasPoint`), an upstream AHU (`isFedBy`) and a downstream HVAC Zone (`feeds`). These won't be "universal" in that we'll use them for all Brick models, but we should show how one would create such a shape for a particular deployment

Also it will be good to draw insights via sparql from Brick.ttl.  For example when I make shapes for feedsAir as it is, I have to specify substanceSuperClass.  But this info can be obtained by querying for classes that have Brick:Air as a subclass.

I'll put those thoughts into the future plan Google doc.

xyzisinus:
I have send thought on expectedDomain/Range stuff in properties.py.  They should not be there because the normal Brick.ttl is generated using the file, too.  I'd put .domain and .range back.  Then use a simple tool to remove .domain and .range from standard Brick.ttl.  Use this modified Brick.ttl for shacl reasoning.

xyzisinus:
Put Gabe's email here for the records:

1. SHACL definitions for Brick: We need a set of restrictions (called "shapes" in SHACL-speak) that verify that the Brick relationships are being used appropriately. E.g. if the "brick:feeds" relationship appears in a Brick model, then the subject and object of that relationships should have compatible types. Charlene has at least a few of these working; I need to verify the completeness but I think we're almost there

2. Integrate SHACL into the tooling: how do people use the SHACL shapes we are defining? I know Charlene has a `validate.sh` script but it would be nice if some of this functionality was folded into the https://github.com/BrickSchema/py-brickschema package so that it was only a couple lines of Python away from being used. Maybe in a `validation.py` submodule? We also need to put documentation for the SHACL validation procedure into the main README of the Brick repository

3. Examples of SHACL shapes for equipment templates: a use case that has been brought up several times by potential users of Brick is how to ensure that certain modeling idioms are followed. These exist at a site- or model- level, rather than at the level of the ontology like the shapes I mentioned above. For a particular site, we may know that all of our VAVs should have a `brick:isFedBy` relationship to an AHU; we can "enforce" this idiom on the model through the use of a SHACL shape (all VAVs should have an isFedBy relationship to an instance of the AHU class). We can't place this restriction on *all* uses of Brick, but we should make it possible to define a shape that enforces this restriction for a particular model. Eventually, it would be nice to have a library of these. Developing the library is a subject of future work. The concrete task I'd like Charlene to handle is implementing and documenting the process of including such optional shapes into the validation process. This part should probably be implemented in the https://github.com/BrickSchema/py-brickschema package

xyzisinus:
Before going ahead with major changes let us see if the following moves sound reasonable.
1. First I propose to remove expectedDomain, expectedRange from properties.py.  Since Brick.ttl is generated using the file, changing the file means changing the Brick schema -- not a light design decision.  Instead, I propose that we feed pyShacl a modified Brick.ttl (with rdfs:domain and :range reamoved) for pyshacl reasoning.  This approach has been coded and tested.  The remaining work is to move the Brick.ttl modification code from generate_shacl.py into the shacl submodule in py-brickschema.  DESIGN.md should be changed accordingly -- maybe moved to py-brickschema, too?
2. Build a submodule shacl (name?) in py-brickschema.  The module should have the following api: a validate() function very much like pyShacl's but it takes Brick.ttl and the output of generate_shacl.py by default; a way to add a user-defined shape as a graph (.addShape()) and a way to add a user-defined shape file (.importShapes()).  Outpuf of validate(), the results_graph is similar to pyShacl's but with extra triples predicated with bsh:offendingTriple to tell user what triples are incorrect.
3. Code samples in the shacl module's document to show how to use the package and explain the expected outcome.
4. Unit tests in py-brickshema.
5. Remaining issues: I assume generate_shacl.py will remain in the Brick repo.  Then how does the shacl module in py-brickschema get the generated shape file?  Should the file be available from our Brick website like Brick.ttl?

PR #98, opened 2020-01-28
Check tags are defined or not

jbkoh:
This change will produce a warning for a class if there is no associated tag with the class. We have to review the results after merging this one.

### DIFF
--- a/generate_brick.py
+++ b/generate_brick.py
@@ -1,4 +1,6 @@
 from collections import defaultdict
+import warnings
+
 from rdflib import Graph, Literal, BNode, URIRef
 from rdflib.namespace import XSD
 from rdflib.collection import Collection
@@ -18,6 +20,13 @@ from bricksrc.substances import substances
 from bricksrc.quantities import quantity_definitions
 from bricksrc.properties import properties
 from bricksrc.tags import tags
+
+
+def warning_on_one_line(message, category, filename, lineno, file=None, line=None):
+    return ' %s:%s: %s:%s\n' % (filename, lineno, category.__name__, message)
+
+warnings.formatwarning = warning_on_one_line
+
 
 G = Graph()
 bind_prefixes(G)
@@ -57,6 +66,9 @@ def has_tags(tagset, definition):
     return all([t in definition for t in tagset])
 
 def add_tags(klass, definition):
+    if not definition:
+        warnings.warn('{0} has no tags.'.format(klass))
+        return
     all_restrictions = []
     equivalent_class = BNode()
     list_name = BNode()
--- a/generate_brick.py
+++ b/generate_brick.py
@@ -25,6 +25,7 @@ from bricksrc.tags import tags
 def warning_on_one_line(message, category, filename, lineno, file=None, line=None):
     return ' %s:%s: %s:%s\n' % (filename, lineno, category.__name__, message)
 
+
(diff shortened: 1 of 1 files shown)


generate_brick.py
```diff
@@ -108,42 +124,28 @@ def add_class_restriction(klass, definition):
...
+def define_subclasses(definitions, superclass, punning_level=0):
+    G.add((superclass, A, OWL.Class))
     for subclass, properties in definitions.items():
         G.add((BRICK[subclass], A, OWL.Class))
-        G.add((BRICK[subclass], A, BRICK[subclass]))
-        G.add((BRICK[subclass], RDFS.label, Literal(subclass.replace("_"," "))))
-        # first level: we are instances of the measurable_class
-        G.add((BRICK[subclass], A, measurable_class))
+        G.add((BRICK[subclass], RDFS.label, Literal(subclass.replace("_", " "))))
+        if punning_level >= 1:
```
gtfierro: I'm *very* confused what you mean by "punning" here. This is not the common definition of "punning" as it is used in OWL 2 modeling. Maybe this PR can stick to checking whether or not tags are applied and not touch the logic for how Brick gets generated?

generate_brick.py
```diff
@@ -260,8 +251,10 @@ def define_properties(definitions, superprop=None):
 #               brick:measures  brick:Air ,
 #                               brick:Temperature .
 # This makes Substance and Quantity metaclasses.
-define_measurable_subclasses(substances, BRICK.Substance)
-define_measurable_subclasses(quantity_definitions, BRICK.Quantity)
+define_subclasses(substances, BRICK.Substance, punning_level=2)
```
gtfierro: This should also probably stay as is

generate_brick.py
jbkoh: There are a couple of problems in the previous code:
1. `add_tags` is not called if there is no tag defined. Thus, it's hard to put the check logic.
2. The parsing logic of `bricksrc/*.py` is implemented in two different places, so it's hard to maintain the readability and the sanity of the logic.
3. The logic `apply_prop` is not straightforward as it's not just applying remaining properties but also used to check if it needs to add subclasses. They should be separated out.

The current implementation make the logic more readable and unifies the parsing logic for substances/properties and the others though there are some rooms for improvement. I can do as you suggested, but, then, I have to duplicate the logic in different places, which I commonly avoid.

I defined `punning_level` to indicate the level of hierarchy you want to associate with the class.
- `Temperature a Temperature` -> Level 1
- `Temperature a Quantity` -> Level 2 (based on the hierarchy)
I can use a different variable name if you have other suggestions.

A different note (may need a different thread): what is the benefit of using `a` for hierarchical relations other than subclasses? I thought a class being an instance of itself was enough for our use cases. What's the benefit of the above design compared with this:
- `Temperature a Temperature`
- `Temperature rdfs:subClassOf Quantity`

gtfierro: I think the proposed implementation complicates the generation of the ontology, and confuses the issues of what is explicitly communicated by the ontology vs what is implied by the model. Please allow me to file my own PR on refactoring the logic to make it more readable

gtfierro:
@jbkoh now that we've merged #102 do you want to close this PR and then open a new PR with all of the tags added to the Brick classes?

jbkoh:
Sure

PR #99, opened 2020-01-28
Fix Meter hierarchy

jbkoh:
`Meter` is not a subclass of `Electrical_System` but devices for a collection of points related to metering. Moving it as a top class of `Equipment`.

### DIFF
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -42,20 +42,56 @@ equipment_subclasses = {
     "Meter": {
         "tags": [TAG.Meter, TAG.Equipment],
         "subclasses": {
+            "Electrical_Meter": {
+                "tags": [TAG.Electrical, TAG.Meter, TAG.Equipment],
+                "subclasses": {
+                    "Building_Electrical_Meter": {
+                        "tags": [TAG.Building, TAG.Electrical, TAG.Meter, TAG.Equipment],
+                        "parents": [BRICK.Building_Meter],
+                    }
+                }
+            },
+            "Gas_Meter": {
+                "tags": [TAG.Meter, TAG.Equipment, TAG.Gas],
+                "subclasses": {
+                    "Building_Gas_Meter": {
+                        "tags": [TAG.Building, TAG.Gas, TAG.Meter, TAG.Equipment],
+                        "parents": [BRICK.Building_Meter],
+                    }
+                }
+            },
             "Water_Meter": {
                 "tags": [TAG.Meter, TAG.Equipment, TAG.Water],
                 "parents": [BRICK.Water_System],
                 "subclasses": {
+                    "Building_Water_Meter": {
+                        "tags": [TAG.Building, TAG.Water, TAG.Meter, TAG.Equipment],
+                        "parents": [BRICK.Building_Meter],
+                    },
                     "Chilled_Water_Meter": {
                         "tags": [TAG.Meter, TAG.Equipment, TAG.Water, TAG.Chilled],
                         "parents": [BRICK.Chilled_Water_System],
+                        "subclasses": {
+                            "Building_Chilled_Water_Meter": {
+                                "tags": [TAG.Building, TAG.Chilled, TAG.Water, TAG.Meter, TAG.Equipment],
+                                "parents": [BRICK.Building_Meter],
+                            },
+                        },
+                    },
+                    "Hot_Water_Meter": {
+                        "tags": [TAG.Meter, TAG.Equipment, TAG.Water, TAG.Hot],
+                        "parents": [BRICK.Chilled_Water_System],
+                        "subclasses": {
+                            "Building_Hot_Water_Meter": {
+                                "tags": [TAG.Building, TAG.Hot, TAG.Water, TAG.Meter, TAG.Equipment],
+                                "parents": [BRICK.Building_Meter],
+                            },
+                        },
                     },
                 },
             },
             "Building_Meter": {
                 "tags": [TAG.Meter, TAG.Equipment, TAG.Building],
-                "subclasses": {
-                }
             },
         },
     },
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -52,9 +52,9 @@ equipment_subclasses = {
                     "Chilled_Water_Meter": {
                         "tags": [TAG.Meter, TAG.Equipment, TAG.Water, TAG.Chilled],
                         "parents": [BRICK.Chilled_Water_System],
-                        },
                     },
                 },
+            },
             "Power_Meter": {
                 "tags": [TAG.Meter, TAG.Equipment, TAG.Power],
                 "parents": [BRICK.Power_System],
@@ -64,19 +64,19 @@ equipment_subclasses = {
                         "subclasses": {
                             "Cooling_Thermal_Power_Meter": {
                                 "tags": [TAG.Meter, TAG.Equipment, TAG.Power, TAG.Thermal, TAG.Cooling],
-                                },
+                            },
                             "Heating_Thermal_Power_Meter": {
                                 "tags": [TAG.Meter, TAG.Equipment, TAG.Power, TAG.Thermal, TAG.Heating],
-                                },
                             },
                         },
                     },
                 },
+            },
             "Building_Meter": {
                 "tags": [TAG.Meter, TAG.Equipment, TAG.Building],
-                },
             },
         },
+    },
     "Water_System": {
         "tags": [TAG.Water, TAG.Equipment],
         "subclasses": {
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -52,24 +52,10 @@ equipment_subclasses = {
                     },
                 },
             },
-            "Power_Meter": {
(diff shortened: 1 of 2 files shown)


gtfierro:
I believe this was originally changed at your request, but we can definitely move it back. Make sure you merge in the latest changes from `master` and I can approve when the tests are fixed

gtfierro:
@jbkoh do you have cycles to address this or should I fix it up?

jbkoh:
Let me update this afternoon. Thanks for the heads up.


With regards,
Jason Koh

bricksrc/equipment.py
```diff
@@ -40,43 +40,43 @@
...
                                 },
                             },
                         },
                     },
-                    "Building_Meter": {
-                        "tags": [TAG.Meter, TAG.Equipment, TAG.Building],
-                    },
+                },
+            "Building_Meter": {
+                "tags": [TAG.Meter, TAG.Equipment, TAG.Building],
```
gtfierro: Can we add `Building_Power_Meter` and `Building_Energy_Meter` to this part of the hierarchy? Would be nice for completeness

bricksrc/equipment.py
```diff
@@ -40,43 +40,43 @@
             "PlugStrip": {
                 "tags": [TAG.PlugStrip, TAG.Equipment],
             },
-            "Meter": {
-                "tags": [TAG.Meter, TAG.Equipment],
+        },
+    },
+    "Meter": {
```
gtfierro: For completeness, can we add `Energy_Meter` to the list of `Meter` subclasses?

bricksrc/equipment.py
```diff
@@ -40,43 +40,43 @@
             "PlugStrip": {
                 "tags": [TAG.PlugStrip, TAG.Equipment],
             },
-            "Meter": {
-                "tags": [TAG.Meter, TAG.Equipment],
```
gtfierro: I agree with moving `Meter` out of the Electrical subsystem hierarchy, but we might want to also eliminate the `Power_System` class (its referred to as the parent of some classes in `Meter` as well). It's not clear what the Power System of a building is, so I think we should remove until we have a better/clearer structure

gtfierro:
The `Meter` is an important concept because it is used for tracking usage such as for billing services. I think if we fully eliminate Meters, then people will be confused. This motivates keeping `Power_Meter` and adding `Energy_Meter`

`Building_Power_Meter` and `Building_Energy_Meter` are useful concepts to refer to as the meter for a full building. We could definitely model these concepts with a `hasPoint` relationship to an instance of Building. While that's the direction to go, its maybe worth keeping these around until we've touched up the Location class structure

jbkoh:
Then, why not `Building_Thermal_Power_Meter`, `Building_Electrical_Power_Meter`, etc.? We will have a lot of redundant classes (co-existing `Power_Meter` and `Power_Sensor`, which will confuse the users more in my opinion.

gtfierro:
`Power_Meter` and `Power_Sensor` are different things they aren't redundant. One is equipment one is a point.

We probably should have `Building`-flavored meter classes as well! Good point

jbkoh:
They are redundant as they are modeling the same concepts in two different places, i.e., measuring some quantities. The difference is whether it's a collection of such measurements or a single point of measurement. People would be confused about what to choose for their modeling regardless of Equipment vs Point. Basically, we should follow the principle of Occum's razor.

Still, I don't mind putting quantity-specific meters if they suffice either of the two things:
- Do they exist?
- Are they useful?

Note that we are "introducing" these classes, which requires reasoning.

gtfierro:
Why are we bringing in "reasoning" into this context? This was always the case and is true independent of the structure of the classes. Its also dubious how relevant Occam's razor is here: the goal is to model what exists in the world, not the simplest possible reduction of it, which overlooks the fine semantic distinctions made by real practitioners.

Recall that equipment and points are different: meters are equipment and sensors are points. Meters are physical pieces of hardware and it will become important to model them as such as we expand into modeling the electrical hierarchy. The choice to make is whether or not it is worth pushing the descriptions of the substances being measured into the name of the equipment that is doing that measurement. How do you feel about an initial step which is the `Building_Meter`?

jbkoh:
> Why are we bringing in "reasoning" into this context? This was always the case and is true independent of the structure of the classes. Its also dubious how relevant Occam's razor is here: the goal is to model what exists in the world, not the simplest possible reduction of it, which overlooks the fine semantic distinctions made by real practitioners.

Yeah, it's exactly what I am saying. We need to model what exists in the world and care about real practitioners. My points are
- Do Power_Meter exist if we have Power_Sensor? In other words, is there equipment dedicated to measuring multiple powers "only"?
    - If such Power_Meter exists in the real world, I have no objection.
- If it does not exist, why is it useful to have both Power_Sensor and Power_Meter, whereas they will confuse real practitioners?

> 
> Recall that equipment and points are different: meters are equipment and sensors are points. Meters are physical pieces of hardware and it will become important to model them as such as we expand into modeling the electrical hierarchy. The choice to make is whether or not it is worth pushing the descriptions of the substances being measured into the name of the equipment that is doing that measurement. 

I am not sure what you are trying to bring this here. If you are saying Power_Meter is the equipment corresponding to Power_Sensor, we need to create all types of equipment for all the Point classes. Moreover, we never limited the definition of Point to virtual points only. 

> How do you feel about an initial step which is the `Building_Meter`?

There are instances of meter equipment measuring different types of points for the entire building. So I'd say Building_Meter is a collection of different metering points dedicated to building, as I saw in our building.

gtfierro:
Power meters are definitely real (e.g. [Smart Meters](https://en.wikipedia.org/wiki/Smart_meter)). They are the physical equipment. Points are _by _definition the digital representation of the sensors, setpoints, etc in the building.  For this reason, a meter may have multiple sensors/points (power sensor, gas usage sensor, etc). I think we could use some better documentation to point out that the points should be attached to a meter where appropriate, to be sure. This doesn't "slippery slope" into making Equipment for all the points, either. This is a very concrete case where we are drawing a distinction the physical piece of metering equipment and the sensor streams it produces.

jbkoh:
[Smart Meters](https://en.wikipedia.org/wiki/Smart_meter) are *an electronic device that records consumption of electric energy and communicates the information to the electricity supplier for monitoring and billing.*

The above def is not for "power"-specific devices so smart meters are not a "Power_Meter" but may have "Power_Sensor". I still can't see a real-world example of `Power_Meter`. I can see a device like [this](https://en.wikipedia.org/wiki/Electricity_meter#/media/File:Transparent_Electricity_Meter_found_in_Israel.JPG) dedicated to electrical energy, which we should interpret as a sensor in Brick.

On the other hand, `Electricity_Meter` should be there as there are devices measuring different points of electrical properties like smart meters in your link.

I looked up power meter here and there
- [this](https://www.amazon.com/Electricity-Monitor-Voltage-Overload-Protection/dp/B07DPJ3RGB/ref=sr_1_2?keywords=electrical+power+meter&qid=1581044682&sr=8-2): measuring power AND energy.
- [that](https://www.powermeterstore.com/P16777/dent-elitepro-esp-xc-power-meters): measuring voltage, power, energy, etc.

Again, I haven't seen any examples where a device is dedicated to only power but multiple instances of power. Instead, people commonly refer to "electrical meters" for measuring different electrical properties.

JoelBender:
Using concepts in [square brackets], I would rephrase this as "drawing a distinction the [physical piece of metering equipment] the [features of interest] that the equipment measures, the [sensors] that measure the feature, and the streams of [observations] that some [measurement procedure] executed in the equipment produces."  This more detailed distinction uses SSN/SOSA terms as I understand them.  Not all of those [things] are necessarily [brick things].

JoelBender:
The FSGIM only talks about electric meters which provide both power and energy measurements, and emissions meters that provide emission measurements (CH4, CO, CO2, CO2e, Hg).  At Cornell we also have chilled water meters (cooling energy) and potable water meters (total quantity and rate).

Curiously, the ASHRAE glossary defines **metering** as a process, "collection of energy data using a measuring device or group of measuring devices", and **metered data** as "energy end use data collected over time using a measuring device or group of measuring devices."  But doesn't define a meter as a measuring device.

The [SEAS Smart Meter Ontology](https://ci.mines-stetienne.fr/seas/SmartMeterOntology-1.1) defines lots of meters, but leaves out chilled water which I guess is a subclass of Thermal Meter.

gtfierro:
The most useful distinction for us is between "the [physical piece of metering equipment] the [features of interest] that the equipment measures". Having the equipment brought forward explicitly means that it can be positioned in terms of a meter hierarchy or downstream/upstream of other electric-related entities. 

As a concrete proposal: how about `Meter` (generic) and `Building_Meter` as equipment. Then the different sensors may be associated with these equipment. 

(going to avoid a digression where a "Power Meter" is a meter that only measures power, which is feasible and real, albeit not as helpful as one which measures other electrical characteristics)

gtfierro:
Decided on the call:

- `Meter` with substance-related subclasses: `Water_Meter`, `Electric_Meter`, etc
- Add `Building_Meter` with the same substance-related subclasses above (until Location hierarchy is fleshed out)

PR #101, opened 2020-01-28
Add missing RDFS.subclassOf property back to measurable hierarchy

gtfierro:
This got removed by accident at some point; this small PR makes sure that the substance + quantity hierarchy is preserved when Brick.ttl is generated

Added a unit test to make sure these hierarchies are preserved when Brick.ttl is compiled as well

### DIFF
--- a/tests/test_measures_inference.py
+++ b/tests/test_measures_inference.py
@@ -1,6 +1,8 @@
 import sys
 sys.path.append('..')
 from bricksrc.namespaces import BRICK
+from bricksrc.substances import substances
+from bricksrc.quantities import quantity_definitions
 import rdflib
 import json
 from collections import defaultdict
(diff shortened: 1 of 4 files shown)


gtfierro:
@jbkoh there is very little here to review but let me know what you think

tests/test_measures_inference.py
```diff
@@ -9,6 +11,44 @@
...
+    for klass, defn in hierarchy.items():
+        if 'subclasses' not in defn:
+            continue
+        for subklass, subdefn in defn['subclasses'].items():
+            yield (klass, subklass)
+        for tup in _get_subclasses(defn['subclasses']):
+            yield tup
+
+
+def test_measurable_hierarchy():
```
jbkoh: I am fine with this, but shouldn't this test cover any classes than just substances/quantities? (and I believe they should be handled in the same way as I modified in the https://github.com/BrickSchema/Brick/pull/98.)

gtfierro: I like factoring the substances and quantity tests into their own case, but I can expand these test cases for the rest of the hierarchy as well.

PR #102, opened 2020-01-29
Refactoring Brick generation script

gtfierro:
Props to @jbkoh and #98 for the impetus and initial implementation

Wanted to take another stab at reworking `generate_brick.py` to make it cleaner, easier to understand, more consistent, and to remove old code. The new structure has fewer methods and has a more structured approach from which it should be more evident how the code is working.

This includes the Python `logging` package to log the stages of ontology generation and to issue warnings when Tags are not defined. As @jbkoh pointed out, we will still need a PR to fill in the rest of the tags.

The new code structure uncovered a few issues in the existing definition: consistent datatypes in the ontology definition files and some mistakes in the structure. This explains why the PR touches files that aren't `generate_brick.py`

### DIFF
--- a/generate_brick.py
+++ b/generate_brick.py
@@ -222,6 +222,7 @@ def define_properties(definitions, superprop=None):
                 G.add((BRICK[prop], propname, propval))
 
 
+logging.info("Beginning BRICK Ontology compilation")
 # handle ontology definition
 define_ontology(G)
 
@@ -244,9 +245,11 @@ roots = {
 }
 define_classes(roots, BRICK.Class)
 
+logging.info("Defining properties")
 # define BRICK properties
 define_properties(properties)
 
+logging.info("Defining Point subclasses")
 # define Point subclasses
 define_classes(setpoint_definitions, BRICK.Point)
 define_classes(sensor_definitions, BRICK.Point)
@@ -261,12 +264,14 @@ for pc in pointclasses:
     for o in filter(lambda x: x != pc, pointclasses):
         G.add((BRICK[pc], OWL.disjointWith, BRICK[o]))
 
+logging.info("Defining Equipment, Location subclasses")
 # define other root class structures
 define_classes(location_subclasses, BRICK.Location)
 define_classes(equipment_subclasses, BRICK.Equipment)
 define_classes(hvac_subclasses, BRICK.HVAC)
 define_classes(valve_subclasses, BRICK.Valve)
 
+logging.info("Defining Measurable hierarchy")
 # define measurable hierarchy
 G.add((BRICK.Measurable, RDFS.subClassOf, BRICK.Class))
 # set up Quantity definition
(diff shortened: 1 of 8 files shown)


gtfierro:
@jbkoh let me fix the unit tests before you take a look; sorry!

gtfierro:
Ok, you are good to look! Adding the "Point" tag to the root classes stuff caused a test to break

generate_brick.py
```diff
@@ -86,90 +103,70 @@ def add_tags(klass, definition):
...
+        # property-object pairs
+        expected_properties = ['parents', 'tags', 'substances', 'subclasses']
+        other_properties = [prop for prop in defn.keys()
+                            if prop not in expected_properties]
+        for propname in other_properties:
+            propval = defn[propname]
+            G.add((classname, propname, propval))
 
 
 def apply_prop(prop, pred, obj):
```
jbkoh: Shouldn't this be removed based on the changes?

gtfierro: This adds the "miscellaneous" properties in: OWL.equivalentClass, SKOS.definition, etc. The first part of the method handles the "special cases" where the keys are not URIs

generate_brick.py
```diff
@@ -215,35 +239,40 @@ def define_properties(definitions, superprop=None):
         "tags": [TAG.Location],
     },
     "Point": {
-        # "tags": [TAG.Point],
+        "tags": [TAG.Point],
```
jbkoh: I wanted to have this from the beginning, but it means that we have to add `TAG.Point` to all the point classes (and better be inside `bricksrc/*.py` instead of this script for consistency imho.)

jbkoh: So my suggestion is to add `TAG.Point` to all the point classes.

gtfierro: We do need to add `TAG.Point` to all of the point classes, but we also need it here because it is the tag of the root `BRICK.Point` class. We can handle adding the tags in a subsequent PR where we add all of the tags to the classes

jbkoh: Yeah I didn't mean to remove this one. This needs to be here anyway. Okay, let's create another PR for missing tags.

generate_brick.py
jbkoh: But it is not used anywhere in the code. Is it used somewhere else?

gtfierro: Which part specifically isn't used in the code? All definitions should be going through this `define_classes` method

jbkoh: btw, I am talking about `apply_prop`.

gtfierro: Oh, your comment is on the `define_classes` method. You are correct I forgot to remove `apply_prop`. Addressed in the latest commit

jbkoh: I added a comment to the exact line and Github focused on the changes I think. Thanks for the fix.

PR #100, opened 2020-01-28
Reworking `Water_System` Hierarchy

jbkoh:
The current model has `Water_System` as a general type of any equipment related to water. However, `Water_System` commonly refers to a collection of devices for water control as Brick 1.0 modeled. I am reverting the current definition to Brick 1.0's in this PR.

For that, I made changes as:
- Add `System` tag to `Water_Systems`.
- Remove `Water_System` as a parent of `Pumps`.

PR #103, opened 2020-01-29
Add missing tags for classes

jbkoh:
I add missing tags for certain classes. Some bugs are fixed during the work such as incorrect hierarchy.

### Progress
- [x] Setpoint
- [x] Sensor
- [x] Status
- [x] Parameter
- [x] Command
- [x] Equipment
- ~~Substance~~
- ~~Quantity~~
- [x] Add `Point` to all point classes

I'd skip for substance/quantity in this PR if it's fine.

### Issues
- It is hard to define `Lighting` because it shares the same tags with `Lighting_System`. We need to address this in a separate PR.

### DIFF
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -7,14 +7,17 @@ Set up subclasses of the equipment superclass
 equipment_subclasses = {
     "HVAC": {
         OWL.equivalentClass: BRICK["Heating_Ventilation_Air_Conditioning_System"],
+        "tags": [TAG.Heating, TAG.Ventilation, TAG.Air, TAG.Conditioning, TAG.System],
     },
     "Heating_Ventilation_Air_Conditioning_System": {
         OWL.equivalentClass: BRICK["HVAC"],
+        "tags": [TAG.HVAC],
     },
     "Weather": {
         "tags": [TAG.Weather],
     },
     "Electrical_System": {
+        "tags": [TAG.Electrical, TAG.System],
         "subclasses": {
             "Emergency_Power_Off_System": {
                 "tags": [TAG.Emergency, TAG.Power, TAG.Off, TAG.Equipment],
@@ -92,9 +95,11 @@ equipment_subclasses = {
             },
             "CWS": {
                 OWL.equivalentClass: BRICK["Chilled_Water_System"],
+                "tags": [TAG.CWS],
             },
             "HWS": {
                 OWL.equivalentClass: BRICK["Hot_Water_System"],
+                "tags": [TAG.HWS],
             }
         }
     },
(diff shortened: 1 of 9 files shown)


gtfierro:
@jbkoh let me know when this is ready for review

jbkoh:
@gtfierro Thanks for the follow-up. It's ready for your review. I was considering to put other issues I found during the change, but they can be separate issues. Please take a look and let me know!

bricksrc/equipment.py
```diff
@@ -7,14 +7,17 @@
 equipment_subclasses = {
     "HVAC": {
         OWL.equivalentClass: BRICK["Heating_Ventilation_Air_Conditioning_System"],
+        "tags": [TAG.Heating, TAG.Ventilation, TAG.Air, TAG.Conditioning, TAG.System],
```
gtfierro: Do we want to have both the `HVAC` and `Heating`, `Ventilation`, etc tags at the same time? They are equivalent classes. I like having just the `HVAC` tag (probably `HVAC` + `System` belong together?) but I could be convinced otherwise

bricksrc/equipment.py
```diff
@@ -92,9 +95,11 @@
             },
             "CWS": {
                 OWL.equivalentClass: BRICK["Chilled_Water_System"],
+                "tags": [TAG.CWS],
```
gtfierro: Some thing here as with `HVAC`; do we want to have the tags for the expanded name, or the abbreviation?

bricksrc/equipment.py
```diff
@@ -92,9 +95,11 @@
...
             },
             "HWS": {
                 OWL.equivalentClass: BRICK["Hot_Water_System"],
+                "tags": [TAG.HWS],
```
gtfierro: Ditto for `HWS` as for `CWS`; same discussion

bricksrc/equipment.py
```diff
@@ -148,6 +169,7 @@
 """
 hvac_subclasses = {
     "Variable_Frequency_Drive": {
+        "tags": [TAG.Equipment, TAG.Variable, TAG.Frequency, TAG.Drive],
```
gtfierro: Abbreviation or expanded name?

bricksrc/equipment.py
```diff
@@ -156,70 +178,107 @@
...
     "Thermostat": {
+        "tags": [TAG.Equipment, TAG.Thermostat],
         SKOS.definition: Literal("An automatic control device used to maintain temperature at a fixed or adjustable setpoint."),
     },
     "Terminal_Unit": {
+        "tags": [TAG.Equipment, TAG.Terminal, TAG.Unit],
         SKOS.definition: Literal("A device that regulates the volumetric flow rate and/or the temperature of the controlled medium."),
         "subclasses": {
             "Fan_Coil_Unit": {
+                "tags": [TAG.Equipment, TAG.Fan, TAG.Coil, TAG.Unit],
```
gtfierro: Abbreviation or written out?

bricksrc/status.py
```diff
@@ -2,272 +2,289 @@
...
-                    "Emergency_Power_Off_Activated_By_Leak_Detection_System_Status": {},
-                    "Emergency_Power_Off_Enable_Status": {},
-                    "Emergency_Power_Off_System_Enable_Status": {}
+                    "Emergency_Power_Off_Activated_By_High_Temperature_Status": {
+                        "tags": [TAG.Point, TAG.Emergency, TAG.Power, TAG.Off, TAG.High, TAG.Temperature, TAG.Status],
+                    },
+                    "Emergency_Power_Off_Activated_By_Leak_Detection_System_Status": {
+                        "tags": [TAG.Point, TAG.Emergency, TAG.Power, TAG.Off, TAG.Leak, TAG.Detection, TAG.Status],
+                    },
+                    "Emergency_Power_Off_Enable_Status": {
```
gtfierro: Incorrect tags for `Emergency_Power_Off_Enable_Status`

jbkoh:
> 
> 
> Nice work! Found a couple small mistakes that should have comments next to them. One question is whether or not we want to have the abbreviated name in a tag, or the expanded name in tags (`HVAC` tag vs `Heating` + `Ventilation` + `Air` + `Conditioning` tags together).

Yeah, it was one of the issues I wanted to discuss as well. I want to stick to one of them only but followed the previous discussion of allowing different sets of tags to classes for now. For consistency, expanding tags makes more sense. However, people use abbreviated tags very commonly. For example, if we want to provide a mapping from haystack tags to brick tags, we need the mapping of a set of haystack tags to a brick tag as well. 

So I am not 100% confident for now, I'd want to allow both abbreviation tags and expanded tags. Any other reasonings and arguments are more than welcome. 

> 
> Can you clarify what you mean about the Lighting system?

`Lighting` is a specific type of `Lighting_System` but their associated Tags are the same for now.
`Lighting` == {`Lighting', 'Equipment`}
`Lighting_System` == {`Lighting`, `Equipment`}
So their tag -> class inference will be nondeterministic.

gtfierro:
> Yeah, it was one of the issues I wanted to discuss as well. I want to stick to one of them only but followed the previous discussion of allowing different sets of tags to classes for now. For consistency, expanding tags makes more sense

Given that the classes are modeled as `owl.equivalentClass` and the current style of tag <-> class inference, it might be fine to leave them as-is. 

> Lighting is a specific type of Lighting_System but their associated Tags are the same for now.
Lighting == {Lighting', 'Equipment}
Lighting_System == {Lighting, Equipment}

You were adding the `System` tag. Maybe add it to `Lighting_System` so that it has different tags? In that case, `Lighting_System` should be the top-level class.

>  So their tag -> class inference will be nondeterministic.

I think you mean "ambiguous" or "overinferred" here; "nondeterministic" has a specific meaning that I don't think applies here.

jbkoh:
> You were adding the `System` tag. Maybe add it to `Lighting_System` so that it has different tags? In that case, `Lighting_System` should be the top-level class.

Yeah, I didn't finish that PR, but the intention of having `System` tag was for a collection of devices for the purpose (not a super class of the devices). I see there are several places where it's controversial to use `System` for a collections of devices. I'd leave Lighting modeling for now and then discuss this after the `System` discussion is done.

gtfierro:
I'm ok with merging this now and then picking up the issue of `System` in #100 after you fix the build. I can also maybe put together a PR today that demonstrates the `System` hierarchy I'm imagining. How does that sound?

jbkoh:
@gtfierro Hi Gabe, I have a draft note to post at #100 but I'd be happy to see another proposal.

I am merging this one and let's continue the other discussions separately. Thanks for your review.

PR #105, opened 2020-01-30
Add test output to .gitignore

gtfierro:
Noticed that a `.ttl` file got committed and decided to add the .ttl and .json files to `.gitignore`


gtfierro:
Bumping @jbkoh. Are you ok with me merging this in? Should be pretty minimal

jbkoh:
Yeah, it annoyed me too. Thanks for the fix.

Issue #104, opened 2020-01-30
Need a coding convention


jbkoh:
We need a coding convention to follow in the projects in Brick. I personally follow PEP8 with 100 chars per line for Python, but happy to settle with anything else.

I am just creating this issue for now so that we can refer to it later as a future task.

JoelBender:
[Black](https://black.readthedocs.io/en/stable/) and [Mypy](https://mypy.readthedocs.io/en/stable/) work for me, some folks include [isort](https://timothycrosley.github.io/isort/).

gtfierro:
[Black](https://github.com/psf/black) is nice because it doesn't have "any" configuration options, which means that we can't spend anymore time bike-shedding the Python style :). [yapf](https://github.com/psf/black) also looks good. I'd just like it to adhere to some standard, and to not deviate unless we have a good reason. Any thoughts?

jbkoh:
Thanks for the suggestions!

For formatting, I vote for Black. 

There are also non-formatting conventions such as [True/False Evaluation](http://google.github.io/styleguide/pyguide.html#214-truefalse-evaluations) I don't think Black can format such things as well, can it?

So, I'd suggest to follow everything defined in Black. Otherwise, follow Google Python conventions.

shreyasnagare:
I tried both `yapf` and `Black` and found `yapf` to produce non-deterministic results. `Black` really is uncompromising. :)

Let's go with Black.

gtfierro:
Sounds like we're going with Black. Does anyone have any issues with the implementation in https://github.com/BrickSchema/Brick/pull/106?

gtfierro:
Closed with #106 being merged!

PR #106, opened 2020-01-30
Trying out the Black Python formatter

gtfierro:
Investigations on #104:

Added a rule `make format` to the Makefile and ran it on the Python files in the repo. I think there is a way of integrating the format into the PR process or automatically applying it when code is merged in, but the Makefile is a good intermediate step for now

Some observations:
- I like the code style it produces; nice to have this handled automatically
- the `bricksrc/` files can have fairly deep nesting, which makes the formatting a little awkward; we might want to adopt some project-specific defaults to make that a little easier, or just disable formatting on those dictionary files

@jbkoh what do you think?


gtfierro:
Some recent changes:
- merged in `master` so there are no conflicts
- using [pre-commit](https://pre-commit.com/) to automatically run the Black code formatter whenever you commit :)
- added development environment notes to CONTRIBUTING

Makefile
```diff
@@ -1,6 +1,13 @@
+.PHONY: format
+
 Brick.ttl: bricksrc/*.py generate_brick.py
 	python generate_brick.py
 
+format:
+	black generate_brick.py
+	black bricksrc/
+	black tests/
+
```
shreyasnagare:
```suggestion
	black tools/

```

bricksrc/location.py
```diff
@@ -4,42 +4,20 @@
...
-            "HVAC_Zone": {
-                "tags": [TAG.HVAC, TAG.Zone, TAG.Location],
-            },
-            "Lighting_Zone": {
-                "tags": [TAG.Lighting, TAG.Zone, TAG.Location],
-            },
-            "Fire_Zone": {
-                "tags": [TAG.Fire, TAG.Zone, TAG.Location],
-            },
+            "HVAC_Zone": {"tags": [TAG.HVAC, TAG.Zone, TAG.Location],},
```
jbkoh: Why does this have a different format than the others? I expect it to be like 
```python
                    "Cold_Box": {
                        "tags": [
                            TAG.Cold,
                            TAG.Box,
                            TAG.Laboratory,
                            TAG.Room,
                            TAG.Location,
                        ],
                    },
```

Maybe it is because of the different length of the line, but is there a way to force such things with Black?

jbkoh: Second thought: I think it's fine as the formatting is automated anyway. But let me know if there is any other opinion.

gtfierro: The whole point is we are relying on automated formatting :). Looks good to me!

gtfierro:
- Do we need/want flake8? Why is using the Black formatter not enough, especially if we then need to go through the trouble of installing exceptions?
- why are we installing exceptions to the rules? E231 seems like a good rule to keep. Are you saying that Black *allows* no spaces between commas (and other chars), or it does *not* allow it?

jbkoh:
>     * Do we need/want flake8? Why is using the Black formatter not enough, especially if we then need to go through the trouble of installing exceptions?

There are useful rules not defined in Black. Black is, in my understanding, strictly designed for formatting whereas PEP8 has a broader scope of coding conventions. For example, https://www.flake8rules.com/rules/E712.html

> 
>     * why are we installing exceptions to the rules? E231 seems like a good rule to keep. Are you saying that Black _allows_ no spaces between commas (and other chars), or it does _not_ allow it?

Black allows this: https://github.com/BrickSchema/Brick/blob/ac7cc62c46403f0ec8de5c6549735bfefadd500b/generate_brick.py#L225
The key-value pair inside the dictionary is ended with a comma. In PEP8, a comma should be followed by a space, but either Black or the trailing-whitespace rule removes it. So, the best solution is just to remove the comma if it is not followed by any more information.

gtfierro:
Is your goal to programmatically remove these small formatting issues (e.g. trailing commas), or to identify them so that they can be fixed manually? I can get flake8 working easily enough, but I don't know if your preference is to remove the flake8 warning by adding exceptions, or to fix the source of the warning and keep the flake8 config closer to default

jbkoh:
> Is your goal to programmatically remove these small formatting issues (e.g. trailing commas), or to identify them so that they can be fixed manually? I can get flake8 working easily enough, but I don't know if your preference is to remove the flake8 warning by adding exceptions, or to fix the source of the warning and keep the flake8 config closer to default

I added the exception to remove human interruption, but I think it makes sense to fix the source to be compatible with default flake8 as much as possible.

gtfierro:
Do you have a set of flake8 rules that you'd like added as exceptions? Currently I just have the list that the author of the blogpost you linked had mentioned.

gtfierro:
I've been working on fixing some of the flake8 errors and most of them seem pretty good and work well with the Black formatter. I'm using the standard set of exceptions that the blog post author mentioned.

One issue that cropped up: In some of the tests, we use the definitions of `bricksrc` to generate the test cases. The way we are currently handling this is by using `sys.path.append` to add the `bricksrc` directory to the set of paths we consider. This always results in angering flake8.

The fix I'm doing right now is to ignore `E402` for lines I know will violate. For example:

```python
 import sys                                                                                                                                                                                                                                                                              
 import rdflib                                                                                                                                                                                                                                                                           
 import brickschema                                                                                                                                                                                                                                                                      
 from rdflib import RDF, OWL, RDFS, Namespace, BNode                                                                                                                                                                                                                                     
 sys.path.append("..")                                                                                                                                                                                                                                                                   
 from bricksrc.namespaces import BRICK, TAG, SOSA  # noqa: E402      
```

Thoughts?

jbkoh:
I like your approach.

gtfierro:
Thanks! Let me know if you have any other suggestions on the formatting pipeline (try the pre-commit stuff out for yourself to see how it works; I haven't used it before now but its pretty nifty) and maybe we can merge this later on this week? I'll handle the file conflicts soon

jbkoh:
A simple request: if a commit is failed because of the format, the current tool will reformat the corresponding files and reject the commit. However, it is unclear that the commit has been failed in the error message.
image

It'd be great if we can simply display that the commit has been rejected. If it's too difficult to implement, we can just explain that in `CONTRIBUTING.md`

jbkoh:
Other than the above request, it looks great!

gtfierro:
Are you asking if we can alter the pre-commit library so that it is more clear that the commit has failed? This should be obvious from looking at the content of the commit (`git status`, `git diff --staged`), which is how I usually use git.

jbkoh:
Let me rephrase the above issue. With the above message, I thought the format has been automatically corrected and then committed. I had to check either `git log` or `git status` to confirm the commit. Is my concern clearer? I just want to be a little more kind to the contributors. Of course, it's not a big deal if the user is familiar with the workflow.

gtfierro:
Ah, ok. I share your concern, but I'm not sure how feasible it is to edit the tools directly for now. I can definitely add some documentation to the `CONTRIBUTING.md` document. I'll point you to that when that's done and we can merge it in with this PR

gtfierro:
Borked the merge so I had to try and accomplish it manually. Everything looks fixed now, but I wanted to explain why there are ~10 commits :)

Take a look at the CONTRIBUTING docs I augmented and let me know what you think: https://github.com/BrickSchema/Brick/blob/63623a036696c87fd8ed325e607dda716443881a/CONTRIBUTING.md#setting-up-development-environment

gtfierro:
@jbkoh let me know what you think of the CONTRIBUTING docs and then I can merge

jbkoh:
Thanks, the description reads great! I am good with this PR.

PR #108, opened 2020-02-07
Fixing up more classes for #88

gtfierro:
- adds (average) heating/cooling demand sensor
- standardize tags to `heat`/`cool` rather than `heating`/`cooling`

### DIFF
--- a/bricksrc/sensor.py
+++ b/bricksrc/sensor.py
@@ -595,6 +595,21 @@ sensor_definitions = {
                             },
                         },
                     },
+                    "Energy_Usage_Sensor": {
+                        "parents": [BRICK.Energy_Sensor],
+                        "tags": [TAG.Point, TAG.Sensor, TAG.Energy, TAG.Usage],
+                        "subclasses": {
+                            "Yearly_Usage_Energy_Sensor": {
+                                "tags": [TAG.Point, TAG.Sensor, TAG.Energy, TAG.Usage, TAG.Yearly],
+                            },
+                            "Monthly_Usage_Energy_Sensor": {
+                                "tags": [TAG.Point, TAG.Sensor, TAG.Energy, TAG.Usage, TAG.Monthly],
+                            },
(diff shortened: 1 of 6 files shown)


gtfierro:
@shreyasnagare let me know if there are any more outstanding classes left for #88 and I can add them in this PR

bricksrc/sensor.py
```diff
@@ -577,6 +595,21 @@
                             },
                         },
                     },
+                    "Energy_Usage_Sensor": {
+                        "parents": [BRICK.Energy_Sensor],
+                        "tags": [TAG.Point, TAG.Sensor, TAG.Energy, TAG.Usage],
+                        "subclasses": {
+                            "Yearly_Usage_Energy_Sensor": {
```
jbkoh: The order of tags in the class names change in these subclasses from `Energy_Usage_Sensor` to `Usage_Energy_Sensor`. I prefer the former one though I have no strong opinion.

gtfierro: Sorry; I'm very confused by what you mean there. Can you rephrase? Are you saying the order of words in `"tags"` is inconsistent? Or the class names are inconsistent?

jbkoh: I meant the class names. I want to change `Yearly_Usage_Energy_Sensor` to `Yearly_Energy_Usage_Sensor` (same for its siblings). Their superclass's name is `Energy_Usage_Sensor`.

gtfierro: Ah, makes sense. Good change

gtfierro:
@shreyasnagare @jbkoh am I good to merge this now?

shreyasnagare:
@gtfierro Looks good to me.

jbkoh:
LGTM.

gtfierro:
Thanks for all your diligent work on this @shreyasnagare and @jbkoh . Really appreciate the thorough feedback :)

PR #110, opened 2020-02-11
Add RealEstateCore alignment

gtfierro:
This is our copy of the alignment file stored at https://github.com/RealEstateCore/rec/blob/master/alignments/BrickSchema.rdf . The alignment was co-developed by members of the REC team and Brick team and may change over time. This is a vetted first cut


gtfierro:
@jbkoh the alignment here is the produce of several rounds of communication with the REC folks. Its a fairly minimal alignment, but let me know what you think and if you think we missed anything

jbkoh:
@gtfierro Thanks for this integration. It looks very useful. We could come up with actual apps with it later as a showcase of how Brick can work with other ontologies for data integration. Of course, the apps alone will be super useful too!

I will take a deeper look tomorrow, but some questions/suggestions at a glimpse.

- Some questions to understand the alignment. I will pick a snippet as an example:
```turtle
brick:Boiler owl:equivalentClass [ owl:intersectionOf ( rec-core:Device [ a owl:Class,
                            owl:Restriction ;
                        owl:hasValue rec-device:Boiler ;
                        owl:onProperty rec-device:hasDeviceFunctionType ] ) ] .
```

    - I looked at REC, but they don't have `rec-device:Boiler`. Are we creating classes for representing `DeviceFunctionType`? Shouldn't REC define those terms on their side as well? Are they going to do it? Otherwise, it doesn't make sense to use their namespace.
    - I see that we use Boiler for both Device and DeviceFunctionType at the same time even though they are under different namespaces. Their names are equivalent, which means that there are overlapping concepts that we did not clarify between Equipment and DeviceFunctionType. Could you explain why such differentiation is necessary?

Take the above as my questions. I strongly believe you and REC team got across those issues.

- I would like to put the REC generation part into a separate script outside `generate_brick.py`.  We will have other integrations in the future and cannot put all of them into the single file.

- It'd be useful to have a couple of example queries in `alignments/rec/README.md` to show how Brick/REC integration is useful.

gtfierro:
> I looked at REC, but they don't have rec-device:Boiler. Are we creating classes for representing DeviceFunctionType? Shouldn't REC define those terms on their side as well? Are they going to do it? Otherwise, it doesn't make sense to use their namespace.

It's defined [here](https://github.com/RealEstateCore/rec/blob/master/device.rdf#L932); it may not be in the latest documentation online but it is part of REC and I expect it will be published on their documentation page soon. We aren't creating classes other than the anonymous owl:Restriction classes. Does that address your other questions here?

> I see that we use Boiler for both Device and DeviceFunctionType at the same time even though they are under different namespaces. Their names are equivalent, which means that there are overlapping concepts that we did not clarify between Equipment and DeviceFunctionType. Could you explain why such differentiation is necessary?

This is because of how REC models equipment. A `rec-device:Boiler` is a DeviceFunctionType, not an equipment. A `rec-core:Device` is a broad class for many points/equipment in REC. They differentiate between different pieces of equipment in terms of their function. Hence, the equivalency here

> I would like to put the REC generation part into a separate script outside generate_brick.py. We will have other integrations in the future and cannot put all of them into the single file.

This is already the case. I haven't merged them into `Brick.ttl` because its not clear we want the alignments in the base distribution by default. I'd rather merge the alignments in now and figure out how to package it up later

> It'd be useful to have a couple of example queries in alignments/rec/README.md to show how Brick/REC integration is useful.

A great idea, and maybe the subject of a future PR? We'd need models that cross both ontologies, which is in the pipeline, but not immediately ready. Having the alignment done is the first step

jbkoh:
> This is because of how REC models equipment. A `rec-device:Boiler` is a DeviceFunctionType, not an equipment. A `rec-core:Device` is a broad class for many points/equipment in REC. They differentiate between different pieces of equipment in terms of their function. Hence, the equivalency here

`rec-core:Device` is defined as "A thing made or adapted for a particular purpose, specifically a piece of electronic equipment". It reads as Boiler should be an instance or a subclass of Device, which has `BoilingFunction` or `BoilerFunction`. I have no control over their design, but this will cause a lot of confusion later.

Another example: they give an example subclass of Device as `Sensor` too. The current names for Boiler are like Sensor hasDeviceFunction Sensor.

> A great idea, and maybe the subject of a future PR? We'd need models that cross both ontologies, which is in the pipeline, but not immediately ready. Having the alignment done is the first step

Though I can imagine that this is going to be useful, I'd rather have a general procedure to understand the use cases first before accepting anything in our core repo. However, I also have no strong opinion on generally hosting any ontology integrations if you are thinking in that way.

gtfierro:
> rec-core:Device is defined as "A thing made or adapted for a particular purpose, specifically a piece of electronic equipment". It reads as Boiler should be an instance or a subclass of Device, which has BoilingFunction or BoilerFunction. I have no control over their design, but this will cause a lot of confusion later.

You should take it up with the REC folks if you really have issues with it, but keep in mind that their model just has a different perspective on these systems than we do. The intuition isn't too bad though: A Brick Boiler is equivalent to a REC Device whose function is that of a boiler. I don't think there's a simpler way of encoding that equivalence than what I have here.

> Though I can imagine that this is going to be useful, I'd rather have a general procedure to understand the use cases first before accepting anything in our core repo. However, I also have no strong opinion on generally hosting any ontology integrations if you are thinking in that way.

Are you arguing that including an example is a prerequisite for including ontology integrations in the repo? I agree they're important, and we will definitely add them, but I want to get the alignment merged first. Developing the examples will take time.

jbkoh:
> You should take it up with the REC folks if you really have issues with it, but keep in mind that their model just has a different perspective on these systems than we do. The intuition isn't too bad though: A Brick Boiler is equivalent to a REC Device whose function is that of a boiler. I don't think there's a simpler way of encoding that equivalence than what I have here.

I thought you hinted them at using Brick Boiler as DeviceFunction there. If `rec:Boiler` is a `rec:Device`, we can simply add 
```turtle
brick:Boiler rdfs:subClassOf rec:Boiler
```
Or, more aggressively, though we have to check whether there are conflicting properties or not,
```
brick:Boiler owl:equivalentClass rec:Boiler
```

Anyway, if you don't have any control over their vocabulary, please don't bother. I am fine with the current one.


> Are you arguing that including an example is a prerequisite for including ontology integrations in the repo? I agree they're important, and we will definitely add them, but I want to get the alignment merged first. Developing the examples will take time.

I could easily agree on your PR, but why is this so urgent?

gtfierro:
Correct: if they had direct class corollaries then the mapping would be much more direct. This is the case for some of the classes in the mapping, but not in the case of devices.

I'm trying to get this merged  (a) as a show of solidarity with the REC folks putting their own mapping up, and (b) because I want to move on to other tasks that can potentially build off of this. Keeping this PR open achieves neither of these things

jbkoh:
okay, I am fine with this PR.

gtfierro:
Thanks! It's definitely important to include examples of REC + Brick together. I hope to assemble some in the coming months

gtfierro:
@jbkoh can you hit "Approve" on the PR when you get a chance? It doesn't complete the Travis merge tests until that's done, I think

PR #111, opened 2020-02-12
Adding Louver and Standby_CRAC

gtfierro:
Weren't added in #108 due to oversight.

Louver is a subclass of Shade equipment and Standby_CRAC is a subclass of CRAC


jbkoh:
LGTM

PR #112, opened 2020-02-13
Clean up test utilities

gtfierro:
-  Remove unused code from the test harness: We are now using the `brickschema` package to implement reasoning, so we can remove all the old code that used to handle the reasoning for us. Now that our utility functions are much simpler, we can embed the only function in `util.py` in the tests directory
- resort to a safer inference method for `brickschema` (requires fewer external dependencies): see https://github.com/BrickSchema/py-brickschema/pull/8 for more context


gtfierro:
We can; what this PR does is use the generic "OWLRLInferenceSession" object which under the covers automatically chooses the fastest implementation rather than mandating a specific one. See [the brickschema docs](https://github.com/BrickSchema/py-brickschema#owlrl-inference) for more info. We can definitely continue to use Allegro, but I didn't want users to have to install/manage docker just to get the tests to run. This way, we can "opt in" to using Allegro for our testing environment

PR #113, opened 2020-02-18
Fix compare_tools.py

jbkoh:
- remove unnecessary imports
- new version style
- add README


tools/compare_versions/compare_versions.py
```diff
@@ -6,18 +6,26 @@
 import argparse
 import semver
 
+import pdb
```
gtfierro: Can you remove the `pdb` import?

gtfierro:
We should probably also add `semver` to the `requirements.txt` for the Brick repo

tools/compare_versions/README.md
```diff
@@ -0,0 +1,7 @@
+Brick Version Comparison
+------------------------
+
+# How to use it?
+- `python tools/compare_versions/compare_versions.py --help` for getting help
+- `python --oldbrick 1.0.3 https://github.com/BrickSchema/Brick/releases/download/v1.0.3/Brick.ttl --newbrick 1.1.0 ./Brick.ttl
```
gtfierro: Does this really work with URLs as well as local files? If so, that's really cool :)

tools/compare_versions/compare_versions.py
jbkoh: done.

jbkoh:
> We should probably also add `semver` to the `requirements.txt` for the Brick repo

It's in this PR.

tools/compare_versions/README.md
jbkoh: Yeah `rdflib` handles both in the same way. It's a good interface.

jbkoh:
Thanks! I will wait for travis and then merge, though there shouldn't be any problem...

PR #114, opened 2020-02-19
Fix parameter subclasses

shreyasnagare:
Remove brick:Discharge_Water_Differential_Pressure_Proportional_Band_Parameter as a subclass of:
- itself
- brick:Supply_Water_Differential_Pressure_Proportional_Band_Parameter


gtfierro:
Good find! I'll approve this as soon as the tests pass

PR #115, opened 2020-02-19
Add subclass hierarchy test

shreyasnagare:
This test can help us find cycles in the subclass hierarchy.

### DIFF
--- a/tests/test_subclass_hierarchy.py
+++ b/tests/test_subclass_hierarchy.py
@@ -1,0 +1,40 @@
+import sys
+from rdflib import RDFS, Graph
+
+sys.path.append("..")
+from bricksrc.namespaces import BRICK
+
+g = Graph()
+g.parse("Brick.ttl", format="turtle")
+
+g.bind("rdfs", RDFS)
+g.bind("brick", BRICK)
+
+
+def minify(node):
+    return node.split("#")[-1]
+
+
+def test_cycles():
+    # Find all the classes that make the hierarchy cyclic.
+    classes_in_cycles = []
+    res1 = g.query(
+        "SELECT ?brick_class WHERE { ?brick_class rdfs:subClassOf+ ?brick_class .}"
+    )
+    for row in res1:
+        classes_in_cycles.append(row[0])
+
+    # Find all the triples involved.
+    loops = set()
+    for brick_class in classes_in_cycles:
+        res2 = g.query(
+            """SELECT DISTINCT ?s ?o WHERE {
+                              ?s rdfs:subClassOf ?o .
+                              brick:%s rdfs:subClassOf+ ?s .
+                              ?o rdfs:subClassOf+ brick:%s .
+                        }"""
+            % (minify(brick_class), minify(brick_class))
+        )
+        for s, o in res2:
+            loops.add("%s -> subClassOf -> %s" % (minify(s), minify(o)))
+    assert not loops, f"Loops found in the class hierarchy!"
--- a/tests/test_subclass_hierarchy.py
+++ b/tests/test_subclass_hierarchy.py
@@ -2,7 +2,7 @@ import sys
 from rdflib import RDFS, Graph
 
 sys.path.append("..")
-from bricksrc.namespaces import BRICK
+from bricksrc.namespaces import BRICK  # noqa: E402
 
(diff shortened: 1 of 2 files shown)


gtfierro:
Good find! I'll approve this as soon as the tests pass

tests/test_subclass_hierarchy.py
```diff
@@ -0,0 +1,40 @@
+import sys
+from rdflib import RDFS, Graph
+
+sys.path.append("..")
+from bricksrc.namespaces import BRICK
```
gtfierro: This should have been flagged by the style checker. Can you make sure `pre-commit` is installed (see instructions @ https://github.com/BrickSchema/Brick/blob/master/CONTRIBUTING.md) and see if this gets flagged?

tests/test_subclass_hierarchy.py
```diff
@@ -0,0 +1,40 @@
...
+        "SELECT ?brick_class WHERE { ?brick_class rdfs:subClassOf+ ?brick_class .}"
+    )
+    for row in res1:
+        classes_in_cycles.append(row[0])
+
+    # Find all the triples involved.
+    loops = set()
+    for brick_class in classes_in_cycles:
+        res2 = g.query(
+            """SELECT DISTINCT ?s ?o WHERE {
```
gtfierro: Stylistically, can you fold this string back to the line above?

```python
res2 = g.query("""SELECT # and so on
```

tests/test_subclass_hierarchy.py
```diff
@@ -0,0 +1,40 @@
...
+                              ?s rdfs:subClassOf ?o .
+                              brick:%s rdfs:subClassOf+ ?s .
+                              ?o rdfs:subClassOf+ brick:%s .
+                        }"""
+            % (minify(brick_class), minify(brick_class))
+        )
+        for s, o in res2:
+            loops.add("%s -> subClassOf -> %s" % (minify(s), minify(o)))
+    assert not loops, f"Loops found in the class hierarchy!"
```
gtfierro: Can you make the test explicit? `assert len(loops) == 0` so its more obvious what the test is checking?

tests/test_subclass_hierarchy.py
shreyasnagare: `Black` formats it like that, I guess. Similar to queries on other files: https://github.com/BrickSchema/Brick/blob/f8c9a4d4ae2cb0206e8719b97b29245bab0239ec/tests/test_no_inference.py#L112

gtfierro: Ok! Far be it from me to question Black's output :)

tests/test_subclass_hierarchy.py
shreyasnagare: You're right, I forgot to install pre-commit on my new clone.

shreyasnagare:
It should pass all the tests once #114 is merged in.

gtfierro:
One thing I'm realizing with my suggestion of `assert len(loops) == 0` is we don't get what the loops are. Can you add the loops themselves to the error message? Something like 

```python
     assert len(loops) == 0, f"Loops found in the class hierarchy for classes: {loops}"
```

Open to any ideas you have for how to make the loops themselves obvious

gtfierro:
I merged in #114 so these tests should pass now

shreyasnagare:
> One thing I'm realizing with my suggestion of `assert len(loops) == 0` is we don't get what the loops are. Can you add the loops themselves to the error message? Something like
> 
> ```python
>      assert len(loops) == 0, f"Loops found in the class hierarchy for classes: {loops}"
> ```
> 
> Open to any ideas you have for how to make the loops themselves obvious

I still think your suggestion is okay. When I try to add the triple `brick:HVAC rdfs:subClassOf brick:RTU .`, the test throws the following error:
```
E       AssertionError: Loops found in the class hierarchy!
E       assert 3 == 0
E        +  where 3 = len({'AHU -> subClassOf -> HVAC', 'HVAC -> subClassOf -> RTU', 'RTU -> subClassOf -> AHU'})
```

I think this output is good enough to identify the issue.

jbkoh:
A great addition. LGTM. Thanks. I am merging this one.

PR #120, opened 2020-03-05
Update location hierachy and add address

jbkoh:
This PR updates Brick's location hierarchy to be logically more aligned to existing models (e.g., IFC, BOT, ISO-12006)
A large part of this update is attributed to a survey by Ambuj Shatdal at Johnson Controls.

### Update in High-level Concepts in Location
- `Basement` and `Roof` now become subclasses of `Floor`.
- Introduce `Site`: An area containing one or more buildings (adopted from BOT). This is distinguished from `haystack:site` referring to `brick:Building`.
    - `Site` can `hasPart` `Building`
- `Room` is now a subclass of `Space`. Room is not defined in either BOT or IFC, but they use `Space` to represent such concepts.
- Removing `City`. I don't know when it came in, but it's too large to model in Brick.

### Address
- Introduce `brick:hasAddress`, a subclass of `vcard:hasAddress`, as a way to associate `vcard:Address` to `brick:Location`.

### Usage of spaces.
- This requires a more extensive discussion. I will create an Issue to discuss this separately.


gtfierro:
I think this is a good start.


1. `Site`s and `Building`s: I'd like to generalize the `Site` concept to be "an area containing zero or more built structures". This distinction is important because not all built things are buildings (and not all sites have anything built) and we would like to reserve the ability to capture other structures in the built environment that may have their own telemetry and metadata.

2. `Floor` vs `Storey`: the latter is generally thought of a as a more general term for any level part of a building, which can encapsulate concepts like "roof" and "garage" and "basement" that people may not normally think of as "Floors".

3. `City`: Interesting discussion as to what it means to be "too big" to model in Brick :), but I'm ok with leaving this off scope for now.

4. `brick:hasAddress`: You are proposing a separate property is so that we can add additional restrictions onto its usage, right? I think that's a good decision.

jbkoh:
>     1. `Site`s and `Building`s: I'd like to generalize the `Site` concept to be "an area containing zero or more built structures". This distinction is important because not all built things are buildings (and not all sites have anything built) and we would like to reserve the ability to capture other structures in the built environment that may have their own telemetry and metadata.

I like the intention but I want to clarify the definition. Referring to https://www.thefreedictionary.com/site, a site means an "the place where a structure or group of structures was, is, or is to be located", so basically "zero" doesn't make sense though we can change the terminology of "building" into "structure". Calling a place without nothing as site sounds weird to me (and not sure if it's useful.). What about "area containing one or more structures"?

If you'd like to model traffic data in the future where there is no structure other than the road, it might make sense but not sure even the concept of site can be useful in that case.

>     2. `Floor` vs `Storey`: the latter is generally thought of a as a more general term for any level part of a building, which can encapsulate concepts like "roof" and "garage" and "basement" that people may not normally think of as "Floors".
Interesting. I wouldn't like to challenge a language issue with a native speaker, but I think I should. I looked up several places but they are interchangeably used. 

Floor is a story or level of a building: https://www.thefreedictionary.com/floor
Storey is a floor or level of a building: https://www.thefreedictionary.com/storey

There are some contextual difference: https://english.stackexchange.com/a/238411
but their architectural meanings are the same.

Basement is a floor (or a set of floors): https://en.wikipedia.org/wiki/Basement

I'd say Roof is neither a floor nor a storey but Rooftop is either a floor or a storey (which I should change in the current PR.)

Could you share a source of the argument? I understand it'd feel too natural to you, but a documented source of the argument would be helpful.

>     3. `City`: Interesting discussion as to what it means to be "too big" to model in Brick :), but I'm ok with leaving this off scope for now.

Yeah, we are including city as a part of the address for now anyway.

gtfierro:
> I like the intention but I want to clarify the definition. Referring to https://www.thefreedictionary.com/site, a site means an "the place where a structure or group of structures was, is, or is to be located", so basically "zero" doesn't make sense though we can change the terminology of "building" into "structure". Calling a place without nothing as site sounds weird to me (and not sure if it's useful.). What about "area containing one or more structures"?

I think it's important to capture both (1) structures in addition to buildings, and (2) not require there to be any buildings in a site. For the first, there are many built structures that may have telemetry associated that are not necessarily "buildings": parking structures, toll booths, parking lots, and so on.

To the second point, your Site definition includes "was" and "is to be", both of which capture the notion of "zero". The Site concept should be able to be instantiated without having any built structure there. One may still have sensors, meters deployed in a site without having anything built there.

---

I'm going to question your methodology of appealing to general English dictionaries for defining technical terms :). We really want to be using technical definitions such as from the [ASHRAE dictionary](https://xp20.ashrae.org/terminology/) or another *technical* resource

ASHRAE defines "floor area":

> the horizontal area within the inside perimeter of the exterior walls of the building space. Also, the sum of the areas of several floors of the building including basements, mezzanine, intermediate floored tiers, and penthouses of headroom height, measured from the exterior faces of exterior walls or from the centerline of walls separating buildings but excluding covered walkways, open roofed-over areas, porches and similar spaces, pipe trenches, exterior terraces or steps, chimneys, roof overhangs, and similar features.

IFC defines "Building storey":

> The building storey has an elevation and typically represents a (nearly) horizontal aggregation of spaces that are vertically bound. 

So from a technical perspective it sounds like they are nearly interchangeable. What gets left out of those definitions is what practitioners tend to use: merits conversation with JCI, most likely.

jbkoh:
> To the second point, your Site definition includes "was" and "is to be", both of which capture the notion of "zero". The Site concept should be able to be instantiated without having any built structure there. One may still have sensors, meters deployed in a site without having anything built there.

"Was" and "is to be" mean different things than "zero". Of course, sensors and meters can be deployed anywhere, but they don't always need to be associated with a certain symbolic "site" regardless. Why do we even want to call someplace without any/to-be/once structures? Any use cases?
- IFC: A site is a defined area of land, possibly covered with water, on which the project construction is to be completed. A site may be used to erect, retrofit or turn down building(s), or for other construction related developments.
- ISO6707-1: area of land or water where construction work or other development is undertaken.

> 
> I'm going to question your methodology of appealing to general English dictionaries for defining technical terms :). We really want to be using technical definitions such as from the [ASHRAE dictionary](https://xp20.ashrae.org/terminology/) or another _technical_ resource
> 
> ASHRAE defines "floor area":
> 
> > the horizontal area within the inside perimeter of the exterior walls of the building space. Also, the sum of the areas of several floors of the building including basements, mezzanine, intermediate floored tiers, and penthouses of headroom height, measured from the exterior faces of exterior walls or from the centerline of walls separating buildings but excluding covered walkways, open roofed-over areas, porches and similar spaces, pipe trenches, exterior terraces or steps, chimneys, roof overhangs, and similar features.
> 
> IFC defines "Building storey":
> 
> > The building storey has an elevation and typically represents a (nearly) horizontal aggregation of spaces that are vertically bound.
> 
> So from a technical perspective it sounds like they are nearly interchangeable. What gets left out of those definitions is what practitioners tend to use: merits conversation with JCI, most likely.

It is always nice to have good references. The only difference between ashrae:floor_area and ifc:storey seems to be whether the entity is surrounded by walls or not. With the difference, terraces are not a part of floors in ashrae:floor. Other than that, a major distinction would  be "roof" (differentiated from rooftop).

I think Storey is more generalizable (regardless of whether we want the generalizability or not) and Floor is more commonly used in HVAC systems. I'd be happy to change Floor to Storey if other people agree. If we stick to Floor, the only difference we need to make is to remove Roof from the Floor subtree.

gtfierro:
> "Was" and "is to be" mean different things than "zero"

True! What I said was they *capture* the notion of there not being anything there at a point in time. Think about how the structures/things associated with a Site may change over time. In the lifecycle of real estate, construction, property management, building management, sensor network management, and so on I don't think it's a very controversial decision to model a Site without necessarily having anything physically built or even anything included in the model.

What if I have some land and I raze the building that's there? Can I no longer model that as a site? What if I am using a Site to plan out *future* buildings and want to model the buildings and supporting infrastructure as that site is developed? Representing "there is nothing here (yet)" is a very useful concept to model

jbkoh:
Sure, I was fine with saying "the place where a structure or group of structures was, is, or is to be located" but not just "zero structures". I will put the definition in the next PR after we discuss Floor vs. Storey tomorrow.

Thanks for the comments!

JoelBender:
Cornell has "facilities" rather than "buildings" and "floors."  A "site" is usually a geographic region that is used to describe the scope of a project which may or many not include one or more facilities.  For example, our three solar farms are sites but they don't have any facilities.

JGD-Sistrol:
Sorry to be late, but Rooftop as a location will clash with Rooftop HVAC equipment. 
Could it be better to leave Roof as Floor subclass, although there are sloped Roofs that are not Floors?

jbkoh:
Hi @JGD-Sistrol. We have [Rooftop_Unit](https://brickschema.org/ontology/1.1/classes/Rooftop_Unit) or RTU for representing what you wanted to describe. Would it solve your problem?

JGD-Sistrol:
RTU or Rootyp_Unit vs Rooftop: Fully solved, many thanks

gtfierro:
@jbkoh it looks like the tests are failing due to the shapes still not being up to date. I checked out the `update-locations` branch and it doesn't look like it's fully up to date with the master branch. 

Fixing the build should be as simple as merging in the latest from master branch. Let me know when you get a chance to run that and push any last changes. Hopefully we can merge this early next week?

bricksrc/properties.py
```diff
@@ -146,4 +146,10 @@
         RDFS.domain: BRICK.Tag,
         RDFS.range: OWL.Class,
     },
+    "hasAddress": {
+        RDFS.subPropertyOf: VCARD.hasAddress,
+        SKOS.definition: Literal("To specify the address of a location."),
+        RDFS.domain: BRICK.Location,
```
gtfierro, line 152: Do we want the domain of `hasAddress` to be a `brick:Building`? I guess if we agree that multiple Brick locations (e.g. it might make sense to have `hasAddress` apply to `brick:Site` as well), then we need to use `brick:Location` instead of a more specific class. I'm open either way, though I do like the idea of saying that an address *must* be associated with a Building

jbkoh: Addresses can be in different granularities, I think. For example, a room can have an address. Unless we can have an exhaustive list of location types that can be associated with an address, I would like to allow any location to have an address. Let me know what you think.

jbkoh:
> One other question: with the room use designation modelling we've discussed, we may end up removing the different room subclasses that are currently defined in the file. Do we want to remove those now or do you think it is better to wait until the room use modelling is fully thought through?

The current PR is mostly incremental. We can remove the usage-related locations when we have a concrete plan later.

bricksrc/properties.py
gtfierro: According to [the vCard documentation](https://www.w3.org/TR/vcard-rdf/#Delivery_Addressing_Properties) it looks like an address refers to a street address. I'm ok with opening up the idea of an address to more than just buildings, but this would require expanding the definition of address beyond what is provided by the vCard ontology

jbkoh: Good catch. I just looked up a bit more and there was a corresponding concept for that: [extended address](https://www.w3.org/TR/vcard-rdf/#d4e860), as described here: https://tools.ietf.org/id/draft-ietf-vcarddav-vcardrev-02.html#ADR
The docs do not explain why it has been deprecated. If this is the case, I agree with your suggestion. I will change the domain of hasAddress to Building only.

gtfierro: I'll keep an eye out for an extended address format/standard that we might be able to apply. Might be an interesting place to use a building/geographic coordinate system? Maybe there are standards for indoor localization or mapping?

jbkoh:
@gtfierro There is an additional change I just pushed in this PR because I thought it was minor, but maybe not. Can you quickly look at the last two commits?

The issue is that generate_brick.py and generate_shacl.py add a newline at the end of the turtle files. I updated the serialization function to strip the spaces. However, somehow, pre-commit detects the generated files are different. I checked those files at the binary level, and they were equivalent.

If it takes time to fix this, I will revert the last two commits.

gtfierro:
Ok I'll take a look tonight!

jbkoh:
@gtfierro Thanks. I updated the above comment with more details.

gtfierro:
I was able to get it to work with the following serialization code:

```python
# serialize to output
with open("Brick.ttl", "wb") as fp:
    fp.write(G.serialize(format="turtle").rstrip())
    fp.write(b"\n")
```

Not quite sure why it ends in LF (which is preceded by `0x00` when I look at it through hexdump), which is really weird. Finishing it off with the newline character after manually stripping the right-sided whitespace seems to work ok though

jbkoh:
That works. I just pushed the fix. Thanks for the help!
I will merge this once it passes the tests.

PR #121, opened 2020-03-10
Add basic security devices and a few example devices. 

epaulson:
This adds a new 'Security Device' to the equipment class hierarchy, and a few example devices. We have more devices but I wanted to start simple. 

This is leaving off relationships/constraints for now, we can talk about that next.

### DIFF
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -179,7 +179,7 @@ equipment_subclasses = {
     },
     "Elevator": {"tags": [TAG.Elevator, TAG.Equipment]},
     "Security_Equipment": {"tags": [TAG.Security, TAG.Equipment]},
-    "Camera_Equipment": {"tags": [TAG.Camera, TAG.Equipment]},
+    "Camera": {"tags": [TAG.Camera, TAG.Equipment]},
 }
 
 
@@ -468,7 +468,7 @@ security_subclasses = {
     "Access_Control_Equipment": {
         "tags": [TAG.Equipment, TAG.Security, TAG.Access, TAG.Control],
         "subclasses": {
-            "Access_Reader_Equipment": {
+            "Access_Reader": {
                 "tags": [
                     TAG.Equipment,
                     TAG.Security,
@@ -515,7 +515,7 @@ security_subclasses = {
                 SKOS.definition: Literal(
                     "An optical instrument to capture still images or record moving images, which are stored on a physical or digital medium."
                 ),
-                "parents": [BRICK.Camera_Equipment]
+                "parents": [BRICK.Camera]
                 # TODO: subclass of PTZ (Pan/Tilt/Zoom) cameras?
             },
             "NVR": {
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -179,6 +179,7 @@ equipment_subclasses = {
     },
     "Elevator": {"tags": [TAG.Elevator, TAG.Equipment]},
     "Security_Equipment": {"tags": [TAG.Security, TAG.Equipment]},
+    "Camera_Equipment": {"tags": [TAG.Camera, TAG.Equipment]},
 }
 
 
@@ -464,10 +465,10 @@ valve_subclasses = {
 }
 
 security_subclasses = {
-    "Access_Control": {
+    "Access_Control_Equipment": {
         "tags": [TAG.Equipment, TAG.Security, TAG.Access, TAG.Control],
         "subclasses": {
-            "Access_Reader": {
+            "Access_Reader_Equipment": {
                 "tags": [
                     TAG.Equipment,
                     TAG.Security,
@@ -500,10 +501,10 @@ security_subclasses = {
         # Door_Release: An electronic input device used to immediately unlock specififed doors that are equipped with electronic locks.
         # Badge Station: A kiosk or checkpoint that requires the use of a badge in order to verify credentials and to grant access.
     },
-    "Video_Surveillance": {
+    "Video_Surveillance_Equipment": {
         "tags": [TAG.Equipment, TAG.Security, TAG.Video, TAG.Surveillance],
         "subclasses": {
-            "Camera": {
+            "Surveillance_Camera": {
                 "tags": [
                     TAG.Equipment,
                     TAG.Security,
@@ -514,6 +515,7 @@ security_subclasses = {
                 SKOS.definition: Literal(
                     "An optical instrument to capture still images or record moving images, which are stored on a physical or digital medium."
                 ),
+                "parents": [BRICK.Camera_Equipment]
                 # TODO: subclass of PTZ (Pan/Tilt/Zoom) cameras?
             },
             "NVR": {
@@ -528,7 +530,6 @@ security_subclasses = {
                 SKOS.definition: Literal("A Network Video Recorder."),
             },
             "Network_Video_Recorder": {
-                OWL.equivalentClass: BRICK["NVR"],
                 "tags": [
                     TAG.NVR,
                     TAG.Equipment,
@@ -537,6 +538,8 @@ security_subclasses = {
                     TAG.Recorder,
                     TAG.Network,
                 ],
+                OWL.equivalentClass: BRICK["NVR"],
+                SKOS.definition: Literal("A Network Video Recorder."),
             },
         },
         # TODO
@@ -545,46 +548,10 @@ security_subclasses = {
         # - is any specific kind of switch e.g. PoESwitch implied here?
         # Video_Wall (or should this be in a separate classification with displays and monitors?)
     },
-    "Safety": {
-        "tags": [TAG.Equipment, TAG.Security, TAG.Safety],
-        "subclasses": {
-            "Automated_External_Defibrillator": {
-                OWL.equivalentClass: BRICK["AED"],
-                "tags": [TAG.AED],
-                SKOS.definition: Literal(
-                    "Automated External Defibrillator. Used by trained people to help those experiencing cardiac issues."
-                ),
-            },
-            "AED": {
-                "tags": [
-                    TAG.Equipment,
-                    TAG.Security,
-                    TAG.Safety,
-                    TAG.AED,
-                    TAG.Defibrillator,
-                ],
-                SKOS.definition: Literal(
-                    "Automated External Defibrillator. Used by trained people to help those experiencing cardiac issues."
-                ),
-            },
-        },
-    },
-    "Intrusion": {
+    "Intrusion_Equipment": {
         "tags": [TAG.Equipment, TAG.Security, TAG.Intrusion],
-        "subclasses": {
-            "Motion_Sensor": {
-                "tags": [
-                    TAG.Equipment,
-                    TAG.Security,
-                    TAG.Intrusion,
-                    TAG.Motion,
-                    TAG.Sensor,
-                ],
-                SKOS.definition: Literal(
-                    "Sensor to indicate motion detected. Similar to Occupancy Sensor"
-                ),
-            },
             # TODO
+            # Motion sensor - but maybe to Points, but still need a way to represent security motion sensors
             # Security Control Panel: The central hub of a security system. All devices are connected to the security panel for easy
             #    and efficient access for different security protocols (i.e. Intrusion security) and events. Question: How’s this different from
             #    Access Panel? Is this specific to Intrusion detection system or more general?
@@ -592,9 +559,8 @@ security_subclasses = {
             # Duress_Button: Panic button, an electronic input device used to help alerting someone in emergency situations.
             # Door_Contacts: Door contact sensor, a peripheral security sensor that lets an alarm system know whether a door is
             # open or closed.
-        },
     },
-    "Intercom": {
+    "Intercom_Equipment": {
         "tags": [TAG.Equipment, TAG.Security, TAG.Intercom],
         "subclasses": {
             "Emergency_Phone": {
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -178,37 +178,7 @@ equipment_subclasses = {
         },
     },
     "Elevator": {"tags": [TAG.Elevator, TAG.Equipment]},
-    "Security_Device": {
-        "tags": [TAG.Equipment, TAG.Security, TAG.Safety],
-        "subclasses": {
-            "Camera": {"tags": [TAG.Equipment, TAG.Security, TAG.Safety, TAG.Camera]},
-            "NVR": {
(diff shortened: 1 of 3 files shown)


jbkoh:
@epaulson Thanks for the PR! Leaving off constraints is fine for now and the proposed new classes are under Equipment, so they will be subjective to the parent class anyway, so no problem at all.

 I have several questions (or suggestions):
- Isn't NVR a subclass Camera then? I was thinking a video recorder is a type of camera. 
- I'd avoid using abbreviated terminologies (`AED` and `NVR`)as much as possible. Are they 
- Can you explain why the tag `Safety` is necessary, especially along with `Security` tag?
- Is `AED` a subclass of `Security Device`? I think it's rather Saftey Equipment. I'd recommend introducing `Safety Equipment` in parallel to `Security Device` too.
- Would you be fine with `Security Equipment` instead of `Security Device`? I would like the top-level concept to be more flexible to accommodate more than standalone devices.
- Both NVC and AED could have more abstracted concepts like `Defibrillator`. `Video Recorder`. Could you add those too for completeness and future extensions?

epaulson:
Jason - 

Thanks for the feedback, we're on-board with much of it. We'll push some changes next week with a bit more abstraction, and then we can have some discussions about what to do re: safety devices

epaulson:
@jbkoh I have pushed new changes to the branch, with a richer hierarchy.

One thing to note - I disabled the pre-commit hook. flake8 kept rejecting a change that black was making, and I couldn't make them both work together so I gave up. Hopefully someone smarter than me can figure out what I'm doing wrong.

gtfierro:
@epaulson I checked out your branch to investigate the linting issue. Is the following error what you get?

```
bricksrc/equipment.py:613:78: E231 missing whitespace after ','
```

If not, can you paste what the error you are seeing is?

In this case, the fix is to go from

```python
            "Video_Intercom": {
                "tags": [TAG.Equipment, TAG.Security, TAG.Intercom, TAG.Video,],
                SKOS.definition: Literal(
                    "An intercom device that has video capabilites as well as voice capabilities"
                ),
            },
```

to 

```python
            "Video_Intercom": {
                "tags": [TAG.Equipment, TAG.Security, TAG.Intercom, TAG.Video],   # <---- no trailng comma after "Video"
                SKOS.definition: Literal(
                    "An intercom device that has video capabilites as well as voice capabilities"
                ),
```
            },

epaulson:
Fixed. I was thrown by this message from flake8:
`(brickvenv) (base) jci-mbp18:Brick cpaulser$ flake8 --config=.flake8 bricksrc/equipment.py 
bricksrc/equipment.py:613:78: E231 missing whitespace after ','`

So I added whitespace, which black then removed. Anyway, I suppose removing the comma removes the need for whitespace, and both black and flake8 were happy.

gtfierro:
Definitely a non-obvious fix there! I only know the fix because I struggled with the same issue when we started using flake8+black

jbkoh:
@epaulson I looked up the update and found a couple of interesting issues worth more discussions.

- Access_Control, Video_Surveilance, and Safety are 'functions' but not equipment. Should we use call them XXX_Equipment? Otherwise, can they be interchangeably used? 
- Isn't NVR a subclass of Camera?
- Cameras may be used for different purposes other than security too (e.g., general motion detection, general occupancy counting, teleconference, etc.). Should Camera be a "subclass" of Security still? If there needs to be something under the security subtree, we could introduce Security_Camera to subclass both Video_Surveilance and Camera. 
- Same for Motion_Sensor. Motion_Sensor is a generic concept not only for security. And the name indicates it is not Equipment but a Point. 
- Is there a special reason to have security.py outside equipment.py in the generation code?

epaulson:
I don't have a strong opinion if we want to stick an _Equipment suffix on the upper levels of the hierarchy, though it seems a bit verbose. 

NVR (Network Video Recorder) is most certainly not a subclass of Camera. Think of it as the server that cameras can store their video in. 

A camera is going to be one of the most common security devices in a built environment. If your query is "find all security devices in this building", you're going to expect to find a bunch of cameras. I don't think you'd ever want to search a building for all cameras and get a mix of the teleconference cameras and the security cameras - the use cases are very different for each camera. We could rename 'camera' in this hierarchy to be 'security_camera' if that'd be better. 

There isn't a security.py file but there is a separate dictionary for security in equipment.py, just like HVAC and valves.

jbkoh:
> I don't have a strong opinion if we want to stick an _Equipment suffix on the upper levels of the hierarchy, though it seems a bit verbose.

Yeah, I feel those verbose too, but a long time ago there was a discussion of `Temperature` vs `Temperature_Sensor` too. I argued it was more important to make the meaning clear than looking good. We can ask for other people's opinions at tomorrow's meeting.

> NVR (Network Video Recorder) is most certainly not a subclass of Camera. Think of it as the server that cameras can store their video in.

Got it. Shall we add a definition like this? "A network video recorder (NVR) is a specialized computer system that includes a software program that records video in a digital format to a disk drive, USB flash drive, SD memory card or other mass storage device" [from wiki](https://en.wikipedia.org/wiki/Network_video_recorder)

> 
> A camera is going to be one of the most common security devices in a built environment. If your query is "find all security devices in this building", you're going to expect to find a bunch of cameras. I don't think you'd ever want to search a building for all cameras and get a mix of the teleconference cameras and the security cameras - the use cases are very different for each camera. We could rename 'camera' in this hierarchy to be 'security_camera' if that'd be better.

I think adding Security_Camera will be better for both usability and completeness. We can also finalize this discussion tomorrow.

> 
> There isn't a security.py file but there is a separate dictionary for security in equipment.py, just like HVAC and valves.

Sorry for the confusion. I wanted to ask if it's necessary to add Security as a top object in the generation code.

gtfierro:
> Sorry for the confusion. I wanted to ask if it's necessary to add Security as a top object in the generation code.

I think this is fine and a construct we use elsewhere in `equipment.py`

> I don't have a strong opinion if we want to stick an _Equipment suffix on the upper levels of the hierarchy, though it seems a bit verbose.

I like including the `_Equipment` suffix to denote the class of equipment serving a function; e.g. `Security_Equipment` is the class of equipment that serves functions for security, `Access_Control_Equipment` is the class of equipment that handles access control, etc. We elide the suffix when we start talking about specific types of equipment (though this isn't a hard and fast rule with a clear boundary)

jbkoh:
Just to summarize our previous resolutions:

- Access_Control -> Access_Control_Equipment
- Video_Surveillance -> Video_Surveillance_Equipment
- Add Equipment tag to Access_Reader. (I am curious how it passed the test actually.)
- Take out Camera out of Security_Equipment. More specifically
    - Introduce Camera as a top-class Equipment.
    - Introduce Surveillance_Camera under Video_Surveillance_Equipment. Surveillance_Camera subclasses both Video_Surveillance_Equipment and Camera.
- Add a definition for NVR.
- Safety -> Safety_Equipment
- Take Safety_Equipment out of Security_Equipment
- Intrusion -> Intrusion_Equipment
- Remove Motion_Sensor from Intrusion. (I just realized Motion_Sensor is already defined as Point.)

A new observation:
- Is Intercom used only for security?

@epaulson Let me know if I forgot anything.

epaulson:
I'll take a look at this next week. 

A dumb question about the 'parents' entry in the python dictionaries: is there an implicit 'parents' entry on anything in the nested 'subclasses'? e.g. when I create a 'survelliance_camera' under 'video_surveillance_equipment' if I was only subclassing from one parent, I wouldn't have needed to say anything, right? Now that I need two parents do I need to list them both or is the parent from the dictionary automatically included?

gtfierro:
@epaulson there is an implicit `parents` entry to the enclosing class (i.e. if you list a bunch of things under `subclasses`, you are automatically the parent). The `parents` key is for pointers to other parts of the class structure that aren't immediately implied by nesting.

bricksrc/equipment.py
```diff
@@ -179,6 +179,7 @@
     },
     "Elevator": {"tags": [TAG.Elevator, TAG.Equipment]},
     "Security_Equipment": {"tags": [TAG.Security, TAG.Equipment]},
+    "Camera_Equipment": {"tags": [TAG.Camera, TAG.Equipment]},
```
jbkoh, line 182: @epaulson We don't have to say it's `Camera_Equpiment` though we need to add that in the Tag level. Could you change this to like this?: `"Camera": {"tags": [TAG.Camera, TAG.Equipment]},`

bricksrc/equipment.py
```diff
@@ -464,10 +465,10 @@
 }
 
 security_subclasses = {
-    "Access_Control": {
+    "Access_Control_Equipment": {
         "tags": [TAG.Equipment, TAG.Security, TAG.Access, TAG.Control],
         "subclasses": {
-            "Access_Reader": {
+            "Access_Reader_Equipment": {
```
jbkoh, line 471: Sorry for my ignorance in the previous comment. Could you take out `_Equipment` postfix here? I was meaning `Equipment` as a Tag but it's already there.

bricksrc/equipment.py
```diff
@@ -514,6 +515,7 @@
                 SKOS.definition: Literal(
                     "An optical instrument to capture still images or record moving images, which are stored on a physical or digital medium."
                 ),
+                "parents": [BRICK.Camera_Equipment]
```
jbkoh, line 518: Change this to `BRICK.Camera` based on the other comment

bricksrc/equipment.py
```diff
@@ -545,56 +548,19 @@
...
-                    TAG.Defibrillator,
-                ],
-                SKOS.definition: Literal(
-                    "Automated External Defibrillator. Used by trained people to help those experiencing cardiac issues."
-                ),
-            },
-        },
-    },
-    "Intrusion": {
+    "Intrusion_Equipment": {
```
jbkoh, line 551: I guess it should be something like `Intrusion_Detection_Equipment`? Can this be correct without `Detection` tag?

bricksrc/equipment.py
```diff
@@ -545,56 +548,19 @@
         # - is any specific kind of switch e.g. PoESwitch implied here?
         # Video_Wall (or should this be in a separate classification with displays and monitors?)
     },
-    "Safety": {
```
jbkoh: @epaulson I didn't mean to "remove" `Safety` here. I want this subtree to be independent to `Security_Equipment` as we discussed last time. I believe this subtree is still useful.

jbkoh:
@epaulson I left several comments again. I think I misspoke some of my thoughts last time. These should be the final request of change. Thanks!

bricksrc/equipment.py
epaulson: Done

bricksrc/equipment.py
epaulson: Done

bricksrc/equipment.py
epaulson: Done

bricksrc/equipment.py
epaulson: I'm not quite sure that we'd want to add 'Detection' this high in the tree, since it's perfectly reasonable to call it 'Intrusion Control Equipment' or 'Intrusion management'. I'm not deadset against adding detection, just not convinced it's necessary.

bricksrc/equipment.py
epaulson: Yes, I'll add it in a new PR once this one is finished.

epaulson:
BTW, I mentioned a while back that the tests failed and I didn't understand why - I think what's happening is that in my (conda) environment, when I created a fresh environment and installed requirements.txt, it picks up a copy the brickschema package and puts it into site-packages and 'make test' seems to use at least some of that? 

My 'make test' run this afternoon failed with Docker (!) problems which was completely unexpected, but I think it's picking that up from something trying to use Allegrograph. I decided not to try to debug that, since I think my few-line changes are probably not the source of that failed test

(Also I haven't been checking in the regenerated .ttl files, since I figured that will just be a merge nightmare)

epaulson:
@jbkoh @gtfierro  any thoughts on the 'detection' name for the intrusion_equipment. I'm fine with changing it to 'Intrustion_Detection_Equipment' if that's better - I'm just a little unclear what that does for tagging, e.g. if I had a device in my building and I was only doing tags, not Brick, I'm not sure that if I was putting tags onto it in my BMS that I'd use the tags 'Intrusion', 'Equipment' and 'Detection' - I'd probably omit the 'Detection' tag - but if the point is to have the union of all possible tags, then I have absolutely no issue with including it in the tags and even putting it in the name.

jbkoh:
Sorry, I should've followed this up earlier. 

@epaulson Regarding the tag "Detection" tag issue. Allowing tagging does not mean that we will allow arbitrary tagging. Here being arbitrary means both something random and a set of tags familiar to certain users out than than being complete. For example, one may argue that she just wants to put only "zone" and "temperature" to a ZNT point, whereas only two is partial to any of the Classes in Brick. It could be either Zone_Air_Temperature_Sensor or Zone_Air_Temperature_Setpoint. So our approach is to be complete and exact as much as possible at the Brick schema side. For allowing "zone" and "temperature" tagging to an entity, an external tool can infer that the closest Class to the set of tags is Zone_Air_Temperature_Sensor (with some heuristics). The same thing applies to the Intrusion_Detection_Equipment.

All that being said, I think it's fine to have Intrusion_Management_Equipment (or Control) if you think its subclasses will have more diversity than just detection. Its my limited imagination of what its subclasses could be. I just think Intrusion_Equipment alone means its Equipment "for" Intrusion, which is weird.

jbkoh:
@epaulson Regarding the development workflow, I will try to run everything from this PR once you push all the changes. I haven't experienced such a docker issue. If available, please put the exact message you saw.  I agree that debugging the development environment can be too much hassle to push small changes. Let the Travis instance do the job for now. I will debug it when this is ready.

epaulson:
I just went ahead and added detection to the name, i'm fine with it. Tyco is too - see here for more example devices: https://www.tycois.com/insights-and-opinions/security-devices/intrusion-detection-devices 

I think this is ready.

bricksrc/equipment.py
jbkoh: @epaulson I don't see Safety_Equipment in the recent commit yet. Do you want to merge this without it now?

jbkoh:
@epaulson I just ran the test after merging it with the master branch, and confirmed it worked fine. If you want, you can try to check it yourself too, but I think this is good to be merged. The only thing is `Safety_Equipment` that you originally wanted to include. I support the idea, and please let me know if you'd like to still add it. I can do it for you too if you are busy. Please let me know your thought.

epaulson:
@jbkoh Let's do safety equipment in a new PR

jbkoh:
@epaulson I am finally merging this one. Thanks for all the hard work!

PR #125, opened 2020-04-03
Fix: properties inside restrictions are not converted

shreyasnagare:
```
:myclass rdfs:subClassOf [ a owl:Restriction ;
            owl:onProperty brickframe:hasPoint ;
            owl:someValuesFrom brick:Change_Filter_Alarm ] .
```
should be converted to 
```
:myclass rdfs:subClassOf [ a owl:Restriction ;
            owl:onProperty brick:hasPoint ;
            owl:someValuesFrom brick:Change_Filter_Alarm ] .
```

PR #126, opened 2020-04-03
Fix: properties inside restrictions are not converted

shreyasnagare:
```
:myclass rdfs:subClassOf [ a owl:Restriction ;
            owl:onProperty brickframe:hasPoint ;
            owl:someValuesFrom brick:Change_Filter_Alarm ] .
```
should be converted to 
```
:myclass rdfs:subClassOf [ a owl:Restriction ;
            owl:onProperty brick:hasPoint ;
            owl:someValuesFrom brick:Change_Filter_Alarm ] .
```
Ref: #125

### DIFF
--- a/tools/convert/conversions/1.0.2-1.1.json
+++ b/tools/convert/conversions/1.0.2-1.1.json
@@ -12,64 +12,136 @@
       "query": "DELETE{ ?subject brickframe:contains ?object . } INSERT{ ?subject brick:isLocationOf ?object . } WHERE { ?subject brickframe:contains ?object . }"
     },
     {
+      "description": "contains => isLocationOf",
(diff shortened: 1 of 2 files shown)


shreyasnagare:
@gtfierro I'm sure both branches have all the latest changes. I think this has something to do with the changes made to `py-brickschema`.

gtfierro:
Hmmmm ok. Changes to py-brickschema *shouldn't* have affected these tests but it looks like they are. Good catch! I'll dig into this

gtfierro:
@shreyasnagare looks like the tests are failing because `brickschema` incorporated the latest release of Brick, for which the Brick unit tests were not updated. This is a dependency loop in our development process that we'll need to smooth over. I'll submit the changes to Brick in a separate PR, you can merge that in to this branch (assuming my PR fixes the issue) and that should solve these issues.

gtfierro:
@shreyasnagare can you merge from master and see if your tests are passing now?

shreyasnagare:
@gtfierro Tests are passing now.

PR #127, opened 2020-04-03
Fixing broken unit tests

gtfierro:
Unit tests were broken in master due to a couple reasons:

1.  Unit tests in Brick repo were not updated to incorporate the necessary tags for inferring the required classes. this has been fixed by updating the tags in the Brick unit tests
2. The unit tests use the `brickschema` package to perform inference, but didn't use the correct flag to use the local development version of Brick rather than the version of Brick packaged with the `brickschema` library. The Brick unit tests here now rebuild the necessary indices on their own so that we don't hit this issue in the future (i.e. if `brickschema` is out of date)

### DIFF
--- a/Brick.ttl
+++ b/Brick.ttl
@@ -1,9 +1,11 @@
 @prefix brick: <https://brickschema.org/schema/1.1/Brick#> .
+@prefix bsh: <https://brickschema.org/schema/1.1/BrickShape#> .
 @prefix dcterms: <http://purl.org/dc/terms#> .
 @prefix owl: <http://www.w3.org/2002/07/owl#> .
 @prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
 @prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
 @prefix sdo: <http://schema.org#> .
+@prefix sh: <http://www.w3.org/ns/shacl#> .
 @prefix skos: <http://www.w3.org/2004/02/skos/core#> .
 @prefix sosa: <http://www.w3.org/ns/sosa#> .
 @prefix tag: <https://brickschema.org/schema/1.1/BrickTag#> .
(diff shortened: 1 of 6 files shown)


gtfierro:
@shreyasnagare @jbkoh this should fix the unit test issues you are seeing in your pending PRs. Take a look through to see the changes. I'm not 100% sure on the ontology changes -- seems like there were some weird structures in there before -- so I'm open to a different fix there. I think the rest of the fixes are fairly straightforward though

gtfierro:
@jbkoh any feedback on the ontology edits? A few PRs are waiting on this being merged

Brick.ttl
```diff
@@ -3066,6 +3058,17 @@ brick:Max_Discharge_Air_Static_Pressure_Setpoint_Limit a owl:Class ;
         tag:Setpoint,
         tag:Static .
 
+brick:Max_Discharge_Air_Temperature_Setpoint a owl:Class ;
```
jbkoh, line 3061: What is this?

Brick.ttl
```diff
@@ -3463,6 +3466,17 @@ brick:Min_Discharge_Air_Static_Pressure_Setpoint_Limit a owl:Class ;
         tag:Setpoint,
         tag:Static .
 
+brick:Min_Discharge_Air_Temperature_Setpoint a owl:Class ;
```
jbkoh, line 3469: Same as above. Why do we have this one without `_Limit`?

Brick.ttl
gtfierro: These were technically already in the ontology under `setpoint.py`; all I did was adjust what they were a subclass of (which didn't make sense). They should be `Parameter` instead, so I can make that fix

Brick.ttl
gtfierro: Same as above

Brick.ttl
jbkoh: I think they should have been removed before, but not just moved to `Parameter`. Our choice was to use `_Limit` for such classes, if I recall correctly. But in this PR, I think your change makes sense. We can resolve this issue later if you prefer.

gtfierro: Addressed in the most recent commit

gtfierro: They weren't removed before, and their `Parameter`ized selves do exist in the `Parameter` hierarchy with the expected `_Limit` suffix

gtfierro:
My comment at the top of the PR documents how we are decoupling this development from the `brickschema` packaging, so while we've exported the functionality of reasoning to the package, we are no longer dependent upon the version of Brick that is packaged with `brickschema`. This should reduce if not eliminate these issues. I've added the versioning bound to the requirements.txt so we know we're using a functional version of `brickschema`. Semver should handle the rest.

Are there any specific changes you are requesting?

jbkoh:
Sorry I missed the version number included in the file changes. This looks good to me and thanks!

gtfierro:
Thanks! Can you mark the PR as approved so I can merge it? I'll wait until tests pass of course

PR #128, opened 2020-04-10
Autogen SHACL

gtfierro:
Fixing up some loose ends in the SHACL implementation that cause future PRs to fail if they introduce new classes with properties that are validated by SHACL

- SHACL generation, tests should be properly formatted now
- BrickShape.ttl file is automatically generated when its dependencies change

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,13 +3,17 @@
 Brick.ttl: bricksrc/*.py generate_brick.py
 	python generate_brick.py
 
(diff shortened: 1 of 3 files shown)


gtfierro:
@jbkoh let me know when you've gotten a chance to look over this and then I'll merge it into master

gtfierro:
Thanks! Let me know if you have any trouble getting the fix incorporated on the Location PR

PR #131, opened 2020-04-23
Units prototype with QUDT

gtfierro:
# Units Prototype

Basic model structure is as follows

```
                brick:measures          brick:associatedQuantityKind
  ╔════════════╗      |      ╔════════════════╗      |      +-------------------+
  ║Brick Point ║════════════>║ Brick Quantity ║════════════>| QUDT QuantityKind |
  ║   Class    ║             ╚══════╦═════════╝             +---------+---------+
  ╚════════════╝                    ║                                 |
        |                           ║ qudt:applicableUnit             |
        | rdf:type                  ║ (compiled into Brick def)       |
        v                           ║                                 |
  ┌─────────┐                       v                                 |
  │Point    │ brick:hasUnit   +---------+     qudt:applicableUnit     |
  │Instance │────────────────>|QUDT Unit| <---------------------------+
  └─────────┘                 +---------+

```

- Bolded (double-line) boxes, edges are defined in the Brick schema 
- Smooth (single-line) boxes, edges are defined in a Brick model (an instance of a building)
- Dashed boxes, edges are defined in QUDT and are *not* included in the Brick distribution

The high-level idea is thus:

- Brick Point classes are associated with a Brick Quantity:
  - currently the association is done for most Sensors (this implementation was done in Brick v1.1)
  - association is done via the `brick:measures` relationship
  - the idea is that knowing the Quantity associated with a Point class should help with:
    - a process can verify that the units associated with an instance of a point are valid/consistent with the Quantity of that point's class
    - user interfaces/libraries can look up associated/applicable units to suggest for instances of the point
- `bricksrc/quantity.py` associates Brick Quantity classes (arranged in a hierarchy) with QUDT QuantityKind classes (a flat namespace)
    - the association between a Quantity and a QuantityKind is done through the `brick:associatedQuantityKind` property
    - the actual property here can change depending on the answer to the below issues
- `generate_brick.py` uses the `brick:associatedQuantityKind` property to pull in the `qudt:applicableUnit` properties to the Brick quantity classes
    - this is arguably brittle because our mapping between quantities, quantitykinds and units might diverge from QUDT if it were to change
    - however, by pulling the `qudt:applicableUnit` property into the Brick ontology definition, we don't need to import the QUDT graph

## Implementation Issues

**Issue**: QuantityKinds are a flat namespace and the Brick Quantity namespace is hierarchical

First question is: does the Brick Quantity namespace have to be hierarchical?
It is helpful to group quantities by general subject (e.g. "Luminous Flux" and
"Luminous Intensity" are related to "Luminance"). It may be that a more
appropriate organization uses `skos:broader`/`skos:narrower` or `rdfs:seeAlso`
or something else. I'm open here! It is safe to say that we are not currently
making use of the existing subclass relationship between Quantities.  The
choice to make them subclasses was driven by the Haystack v4 quantity/substance
organization.

We cannot use `owl:equivalentClass` to associate Brick Quantities with QUDT
QuantityKinds because QuantityKinds are not classes!

Because we are using `rdfs:subClassOf`, we can't use `owl:sameAs` to associate
Brick Quantities with QUDT QuantityKinds because all of the Quantity subclasses
will be inferred to be the same entity! This is bad for many reasons.

**Question**: why the indirection of Brick Quantity? Why not use QUDT QuantityKinds directly?

QUDT QuantityKinds do not all have applicable units; there is some relation of
concepts in QUDT that isn't super clean (e.g. QUDT ElectricPower vs QUDT
ActivePower, ReactivePower, etc; the latter concepts are "independent" of the
electrical context but there is no way to compose them). I argue it is helpful
for Brick to provide an organization of quantities that is more specialized for
the built environment domain. We have the most control over this by defining our
own concepts (Brick Quantities) and relating them to the QUDT ontology.

## Reflection and Next Steps

This implementation is provided as a record of the current attempt. Moving
forward, I think the best choice is to use `skos:broader`, `skos:narrower` (or
an analogous pair of properties) to relate Brick Quantities to one another.
Then, we can use `owl:sameAs` to relate Brick Quantities with QUDT
QuantityKinds, which makes the 'inheritance' of properties from QUDT QuantityKinds
more automatic and easier to maintain. 

We might still want to replicate `qudt:applicableUnit` in the Brick ontology
definition (or develop our own analogous property) in order to avoid needing
the QUDT ontology loaded in.

### DIFF
--- a/bricksrc/properties.py
+++ b/bricksrc/properties.py
@@ -1,5 +1,5 @@
 from rdflib import Literal
-from .namespaces import A, OWL, RDFS, SKOS, BRICK, QUDTUNIT
+from .namespaces import A, OWL, RDFS, SKOS, BRICK, QUDTUNIT, QUDT
 
 """
 Defining properties
@@ -151,5 +151,11 @@ properties = {
         SKOS.definition: Literal("The QUDT unit associated with this Point"),
         RDFS.domain: BRICK.Point,
         RDFS.range: QUDTUNIT.Unit,
+    },
+    "associatedQuantityKind": {
+        A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+        SKOS.definition: Literal("The QUDT QuantityKind associated with this Quantity"),
+        RDFS.domain: BRICK.Quantity,
+        RDFS.range: QUDT.QuantityKind,
     },
 }
--- a/bricksrc/properties.py
+++ b/bricksrc/properties.py
@@ -1,5 +1,5 @@
 from rdflib import Literal
-from .namespaces import A, OWL, RDFS, SKOS, BRICK
+from .namespaces import A, OWL, RDFS, SKOS, BRICK, QUDTUNIT
 
 """
 Defining properties
@@ -145,5 +145,11 @@ properties = {
         SKOS.definition: Literal("The tag is associated with the given class"),
         RDFS.domain: BRICK.Tag,
         RDFS.range: OWL.Class,
+    },
+    "hasUnit": {
+        A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+        SKOS.definition: Literal("The QUDT unit associated with this Point"),
+        RDFS.domain: BRICK.Point,
+        RDFS.range: QUDTUNIT.Unit,
     },
 }
--- a/bricksrc/properties.py
+++ b/bricksrc/properties.py
@@ -1,5 +1,5 @@
 from rdflib import Literal
-from .namespaces import A, OWL, RDFS, SKOS, BRICK, VCARD, QUDTUNIT, QUDT
+from .namespaces import A, OWL, RDFS, SKOS, BRICK, VCARD, UNIT, QUDT
 
 """
 Defining properties
@@ -156,7 +156,7 @@ properties = {
         A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
         SKOS.definition: Literal("The QUDT unit associated with this Point"),
         RDFS.domain: BRICK.Point,
-        RDFS.range: QUDTUNIT.Unit,
+        RDFS.range: UNIT.Unit,
     },
     "associatedQuantityKind": {
         A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
--- a/bricksrc/properties.py
+++ b/bricksrc/properties.py
@@ -158,10 +158,4 @@ properties = {
         RDFS.domain: BRICK.Point,
         RDFS.range: UNIT.Unit,
     },
-    "associatedQuantityKind": {
-        A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
-        SKOS.definition: Literal("The QUDT QuantityKind associated with this Quantity"),
-        RDFS.domain: BRICK.Quantity,
-        RDFS.range: QUDT.QuantityKind,
-    },
 }
--- a/bricksrc/quantities.py
+++ b/bricksrc/quantities.py
@@ -47,7 +47,7 @@ quantitykind_extensions = {
 
 quantity_definitions = {
     "Air_Quality": {
-        "subclasses": {
+        "subconcepts": {
             "CO2_Level": {OWL.sameAs: QUDTQK["Concentration"]},
             "PM10_Level": {OWL.sameAs: QUDTQK["Concentration"]},
             "PM25_Level": {OWL.sameAs: QUDTQK["Concentration"]},
@@ -66,10 +66,10 @@ quantity_definitions = {
     "Grains": {},
     "Power": {
         OWL.sameAs: QUDTQK["Power"],
-        "subclasses": {
+        "subconcepts": {
             "Electric_Power": {
                 OWL.sameAs: QUDTQK["ElectricPower"],
-                "subclasses": {
+                "subconcepts": {
                     "Apparent_Power": {OWL.sameAs: QUDTQK["ApparentPower"]},
                     "Active_Power": {
                         OWL.sameAs: [QUDTQK["ActivePower"], BRICK["Real_Power"]],
@@ -85,15 +85,15 @@ quantity_definitions = {
                     "Tracks the highest (peak) observed power in some interval"
                 ),
             },
-            "Thermal_Power": {},
+            "Thermal_Power": {OWL.sameAs: QUDTQK["ThermalPower"]},
         },
     },
     "Cloudage": {},
     "Current": {
-        "subclasses": {
+        "subconcepts": {
             "Electric_Current": {
                 OWL.sameAs: QUDTQK["ElectricCurrent"],
-                "subclasses": {
+                "subconcepts": {
                     "Current_Angle": {},
                     "Current_Magnitude": {},
                     "Current_Imbalance": {},
@@ -104,7 +104,7 @@ quantity_definitions = {
         },
     },
     "Voltage": {
-        "subclasses": {
+        "subconcepts": {
             "Voltage_Magnitude": {OWL.sameAs: QUDTQK["Voltage"]},
             "Voltage_Angle": {},
             "Voltage_Imbalance": {},
@@ -112,20 +112,20 @@ quantity_definitions = {
     },
     "Daytime": {},
     "Dewpoint": {OWL.sameAs: QUDTQK["DewPointTemperature"]},
-    "Direction": {"subclasses": {"Wind_Direction": {}}},
+    "Direction": {"subconcepts": {"Wind_Direction": {}}},
     "Energy": {
-        "subclasses": {
+        "subconcepts": {
             "Electric_Energy": {OWL.sameAs: QUDTQK["Energy"]},
             "Thermal_Energy": {OWL.sameAs: QUDTQK["ThermalEnergy"]},
         },
     },
-    "Flow": {OWL.sameAs: QUDTQK["VolumeFlowRate"], "subclasses": {"Flow_Loss": {}}},
+    "Flow": {OWL.sameAs: QUDTQK["VolumeFlowRate"], "subconcepts": {"Flow_Loss": {}}},
     "Frequency": {
         OWL.sameAs: QUDTQK["Frequency"],
-        "subclasses": {"Alternating_Current_Frequency": {}},
+        "subconcepts": {"Alternating_Current_Frequency": {}},
     },
     "Humidity": {
-        "subclasses": {
+        "subconcepts": {
             "Relative_Humidity": {OWL.sameAs: QUDTQK["RelativeHumidity"]},
             "Absolute_Humidity": {OWL.sameAs: QUDTQK["AbsoluteHumidity"]},
         }
@@ -133,10 +133,10 @@ quantity_definitions = {
     "Illuminance": {OWL.sameAs: QUDTQK["Illuminance"]},
     "Irradiance": {
         OWL.sameAs: QUDTQK["Irradiance"],
-        "subclasses": {"Solar_Irradiance": {}},
+        "subconcepts": {"Solar_Irradiance": {}},
     },
     "Level": {
-        "subclasses": {
+        "subconcepts": {
             "CO2_Level": {},
             "PM10_Level": {},
             "PM25_Level": {},
@@ -145,18 +145,18 @@ quantity_definitions = {
     },
     "Luminance": {
         OWL.sameAs: QUDTQK["Luminance"],
-        "subclasses": {
+        "subconcepts": {
             "Luminous_Flux": {OWL.sameAs: QUDTQK["LuminousFlux"]},
             "Luminous_Intensity": {OWL.sameAs: QUDTQK["LuminousIntensity"]},
         },
     },
-    "Occupancy": {"subclasses": {"Occupancy_Count": {}, "Occupancy_Percentage": {}}},
+    "Occupancy": {"subconcepts": {"Occupancy_Count": {}, "Occupancy_Percentage": {}}},
     "Position": {},
     "Power_Factor": {OWL.sameAs: QUDTQK["PowerFactor"]},
     "Precipitation": {},
     "Pressure": {
         OWL.sameAs: QUDTQK["Pressure"],
-        "subclasses": {
+        "subconcepts": {
             "Atmospheric_Pressure": {OWL.sameAs: QUDTQK["AtmosphericPressure"]},
             "Dynamic_Pressure": {},
             "Static_Pressure": {OWL.sameAs: QUDTQK["StaticPressure"]},
@@ -165,11 +165,11 @@ quantity_definitions = {
             },
         },
     },
-    "Radiance": {OWL.sameAs: QUDTQK["Radiance"], "subclasses": {"Solar_Radiance": {}}},
-    "Speed": {OWL.sameAs: QUDTQK["Speed"], "subclasses": {"Wind_Speed": {}}},
+    "Radiance": {OWL.sameAs: QUDTQK["Radiance"], "subconcepts": {"Solar_Radiance": {}}},
+    "Speed": {OWL.sameAs: QUDTQK["Speed"], "subconcepts": {"Wind_Speed": {}}},
     "Temperature": {
         OWL.sameAs: QUDTQK["ThermodynamicTemperature"],
-        "subclasses": {
+        "subconcepts": {
             "Operative_Temperature": {},
             "Radiant_Temperature": {},
             "Dry_Bulb_Temperature": {},
@@ -178,7 +178,7 @@ quantity_definitions = {
     },
     "Time": {
         OWL.sameAs: QUDTQK["Time"],
-        "subclasses": {"Acceleration_Time": {}, "Deceleration_Time": {}},
+        "subconcepts": {"Acceleration_Time": {}, "Deceleration_Time": {}},
     },
     "Torque": {OWL.sameAs: QUDTQK["Torque"]},
     "Weather_Condition": {},
--- a/bricksrc/quantities.py
+++ b/bricksrc/quantities.py
@@ -1,5 +1,6 @@
+import brickschema
 from rdflib import Literal
-from .namespaces import SKOS, OWL, BRICK, QUDTQK
+from .namespaces import SKOS, OWL, RDFS, BRICK, QUDTQK, QUDT
 
 
 quantity_definitions = {
@@ -65,13 +66,9 @@ quantity_definitions = {
     "Voltage": {
         OWL.equivalentClass: QUDTQK["Voltage"],
         "subclasses": {
-            "Electric_Voltage": {
-                "subclasses": {
-                    "Voltage_Magnitude": {},
-                    "Voltage_Angle": {},
-                    "Voltage_Imbalance": {},
-                },
-            },
+            "Voltage_Magnitude": {},
+            "Voltage_Angle": {},
+            "Voltage_Imbalance": {},
         },
     },
     "Daytime": {},
@@ -158,3 +155,18 @@ quantity_definitions = {
     "Torque": {OWL.equivalentClass: QUDTQK["Torque"]},
     "Weather_Condition": {},
 }
+
+
+def associate_units(outputG):
+    g = brickschema.graph.Graph()
+    g.g.bind("qudt", QUDT)
+    for triple in outputG:
+        g.add(triple)
+    g.load_file("support/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl")
+    unit_map = g.query(
+        "SELECT ?quant ?unit WHERE {\
+        ?quant a brick:Quantity .\
+        ?quant owl:equivalentClass/qudt:applicableUnit ?unit . }"
+    )
+    for brickquant, unit in unit_map:
+        outputG.add((brickquant, QUDT.applicableUnit, unit))
--- a/bricksrc/quantities.py
+++ b/bricksrc/quantities.py
@@ -11,32 +11,30 @@ quantity_definitions = {
             "TVOC_Level": {},
         },
     },
-    "Angle": {OWL.equivalentClass: QUDTQK["Angle"]},
-    "Conductivity": {OWL.equivalentClass: QUDTQK["Conductivity"]},
-    "Capacity": {OWL.equivalentClass: QUDTQK["Capacity"]},
+    "Angle": {OWL.sameAs: QUDTQK["Angle"]},
+    "Conductivity": {OWL.sameAs: QUDTQK["Conductivity"]},
+    "Capacity": {OWL.sameAs: QUDTQK["Capacity"]},
     "Enthalpy": {
         SKOS.definition: Literal(
             "(also known as heat content), thermodynamic quantity equal to the sum of the internal energy of a system plus the product of the pressure volume work done on the system. H = E + pv, where H = enthalpy or total heat content, E = internal energy of the system, p = pressure, and v = volume. (Compare to [[specific enthalpy]].)"
         ),
-        OWL.equivalentClass: QUDTQK["Enthalpy"],
+        OWL.sameAs: QUDTQK["Enthalpy"],
     },
     "Grains": {},
     "Power": {
-        OWL.equivalentClass: QUDTQK["Power"],
+        OWL.sameAs: QUDTQK["Power"],
         "subclasses": {
             "Electric_Power": {
-                OWL.equivalentClass: QUDTQK["ElectricPower"],
+                OWL.sameAs: QUDTQK["ElectricPower"],
                 "subclasses": {
-                    "Apparent_Power": {OWL.equivalentClass: QUDTQK["ApparentPower"]},
+                    "Apparent_Power": {OWL.sameAs: QUDTQK["ApparentPower"]},
                     "Active_Power": {
-                        OWL.equivalentClass: [
-                            BRICK["Real_Power"],
-                            QUDTQK["ActivePower"],
-                        ]
+                        OWL.sameAs: QUDTQK["ActivePower"],
+                        OWL.equivalentClass: BRICK["Real_Power"],
                     },
                     "Real_Power": {},
-                    "Reactive_Power": {OWL.equivalentClass: QUDTQK["ReactivePower"]},
-                    "Complex_Power": {OWL.equivalentClass: QUDTQK["ComplexPower"]},
+                    "Reactive_Power": {OWL.sameAs: QUDTQK["ReactivePower"]},
+                    "Complex_Power": {OWL.sameAs: QUDTQK["ComplexPower"]},
                 },
             },
             "Peak_Power": {
@@ -51,7 +49,7 @@ quantity_definitions = {
     "Current": {
         "subclasses": {
             "Electric_Current": {
-                OWL.equivalentClass: QUDTQK["ElectricCurrent"],
+                OWL.sameAs: QUDTQK["ElectricCurrent"],
                 "subclasses": {
                     "Current_Angle": {},
                     "Current_Magnitude": {},
@@ -63,7 +61,7 @@ quantity_definitions = {
         },
     },
     "Voltage": {
-        OWL.equivalentClass: QUDTQK["Voltage"],
+        OWL.sameAs: QUDTQK["Voltage"],
         "subclasses": {
             "Electric_Voltage": {
                 "subclasses": {
@@ -75,32 +73,29 @@ quantity_definitions = {
         },
     },
     "Daytime": {},
-    "Dewpoint": {OWL.equivalentClass: QUDTQK["DewPointTemperature"]},
+    "Dewpoint": {OWL.sameAs: QUDTQK["DewPointTemperature"]},
     "Direction": {"subclasses": {"Wind_Direction": {}}},
     "Energy": {
-        OWL.equivalentClass: QUDTQK["Energy"],
+        OWL.sameAs: QUDTQK["Energy"],
         "subclasses": {
             "Electric_Energy": {},
-            "Thermal_Energy": {OWL.equivalentClass: QUDTQK["ThermalEnergy"]},
+            "Thermal_Energy": {OWL.sameAs: QUDTQK["ThermalEnergy"]},
         },
     },
-    "Flow": {
-        OWL.equivalentClass: QUDTQK["VolumeFlowRate"],
-        "subclasses": {"Flow_Loss": {}},
-    },
+    "Flow": {OWL.sameAs: QUDTQK["VolumeFlowRate"], "subclasses": {"Flow_Loss": {}}},
     "Frequency": {
-        OWL.equivalentClass: QUDTQK["Frequency"],
+        OWL.sameAs: QUDTQK["Frequency"],
         "subclasses": {"Alternating_Current_Frequency": {}},
     },
     "Humidity": {
         "subclasses": {
-            "Relative_Humidity": {OWL.equivalentClass: QUDTQK["RelativeHumidity"]},
-            "Absolute_Humidity": {OWL.equivalentClass: QUDTQK["AbsoluteHumidity"]},
+            "Relative_Humidity": {OWL.sameAs: QUDTQK["RelativeHumidity"]},
+            "Absolute_Humidity": {OWL.sameAs: QUDTQK["AbsoluteHumidity"]},
         }
     },
-    "Illuminance": {OWL.equivalentClass: QUDTQK["Illuminance"]},
+    "Illuminance": {OWL.sameAs: QUDTQK["Illuminance"]},
     "Irradiance": {
-        OWL.equivalentClass: QUDTQK["Irradiance"],
+        OWL.sameAs: QUDTQK["Irradiance"],
         "subclasses": {"Solar_Irradiance": {}},
     },
     "Level": {
@@ -112,39 +107,32 @@ quantity_definitions = {
         },
     },
     "Luminance": {
-        OWL.equivalentClass: QUDTQK["Luminance"],
+        OWL.sameAs: QUDTQK["Luminance"],
         "subclasses": {
-            "Luminous_Flux": {OWL.equivalentClass: QUDTQK["LuminousFlux"]},
-            "Luminous_Intensity": {OWL.equivalentClass: QUDTQK["LuminousIntensity"]},
+            "Luminous_Flux": {OWL.sameAs: QUDTQK["LuminousFlux"]},
+            "Luminous_Intensity": {OWL.sameAs: QUDTQK["LuminousIntensity"]},
         },
     },
     "Occupancy": {"subclasses": {"Occupancy_Count": {}, "Occupancy_Percentage": {}}},
     "Position": {},
-    "Power_Factor": {OWL.equivalentClass: QUDTQK["PowerFactor"]},
+    "Power_Factor": {OWL.sameAs: QUDTQK["PowerFactor"]},
     "Precipitation": {},
     "Pressure": {
-        OWL.equivalentClass: QUDTQK["Pressure"],
+        OWL.sameAs: QUDTQK["Pressure"],
         "subclasses": {
-            "Atmospheric_Pressure": {
-                OWL.equivalentClass: QUDTQK["AtmosphericPressure"]
-            },
+            "Atmospheric_Pressure": {OWL.sameAs: QUDTQK["AtmosphericPressure"]},
             "Dynamic_Pressure": {},
-            "Static_Pressure": {OWL.equivalentClass: QUDTQK["StaticPressure"]},
+            "Static_Pressure": {OWL.sameAs: QUDTQK["StaticPressure"]},
             "Velocity_Pressure": {
-                OWL.equivalentClass: [
-                    QUDTQK["DynamicPressure"],
-                    BRICK["Dynamic_Pressure"],
-                ]
+                OWL.sameAs: QUDTQK["DynamicPressure"],
+                OWL.equivalentClass: BRICK["Dynamic_Pressure"],
             },
         },
     },
-    "Radiance": {
-        OWL.equivalentClass: QUDTQK["Radiance"],
-        "subclasses": {"Solar_Radiance": {}},
-    },
-    "Speed": {OWL.equivalentClass: QUDTQK["Speed"], "subclasses": {"Wind_Speed": {}}},
+    "Radiance": {OWL.sameAs: QUDTQK["Radiance"], "subclasses": {"Solar_Radiance": {}}},
+    "Speed": {OWL.sameAs: QUDTQK["Speed"], "subclasses": {"Wind_Speed": {}}},
     "Temperature": {
-        OWL.equivalentClass: QUDTQK["Temperature"],
+        OWL.sameAs: QUDTQK["Temperature"],
         "subclasses": {
             "Operative_Temperature": {},
             "Radiant_Temperature": {},
@@ -153,9 +141,9 @@ quantity_definitions = {
         },
     },
     "Time": {
-        OWL.equivalentClass: QUDTQK["Time"],
+        OWL.sameAs: QUDTQK["Time"],
         "subclasses": {"Acceleration_Time": {}, "Deceleration_Time": {}},
     },
-    "Torque": {OWL.equivalentClass: QUDTQK["Torque"]},
+    "Torque": {OWL.sameAs: QUDTQK["Torque"]},
     "Weather_Condition": {},
 }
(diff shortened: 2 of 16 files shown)


gtfierro:
@jbkoh @JoelBender would be great to get your input/feedback on this!

jbkoh:
Hi Gabe,

- `brick:measures` is for instances. Could you modify the diagram? Btw, nice markdown diagram!

- I believe QuantityKinds are instances, aren't they? (which I think complicates the integration process. I will discuss it in the next bullet point.)

- I would like to have a more direct relation between `brick:Quantity` and `qudt:QuantityKind`. The current edge, `Quantity--applicatbleUnit-->Unit`, is not useful for rules in QUDT because QUDT rules cannot recognize Brick Quantities. However, I understand there is no easy solution for this problem because QuantityKinds are instances. I am curious if `rdf:type` can function like `rdfs:subClassOf` for instances in OWL rules. I will check the specification later.
    - If QuantityKinds are classes, we can use `rdfs:subClassOf` instaed of `brick:associatedQuantityKind`.

- Brick Quantity should have a hierarchy if we want to keep things like `Chilled_Water` for example. I am not entirely sure why QuantityKinds being flat affects our hierarchy decision.


I will review actual mapping after we finish design discussions like this.

gtfierro:
I think I've answered a few of your questions in the text already; I would recommend investigating the mapping implementation itself because it brings a few of these issues to light in a concrete matter.

> brick:measures is for instances. Could you modify the diagram? Btw, nice markdown diagram!

`measures` is currently used to parameterize the Point classes, e.g. Sensor. This design is documented in the House of Sticks paper. `brick:measures` is not associated with Point instances (though we'll need to adjust the `rdfs:domain` property to reflect this; the current annotation is a mistake).

> I believe QuantityKinds are instances, aren't they?  I would like to have a more direct relation between brick:Quantity and qudt:QuantityKind. The current edge, Quantity--applicatbleUnit-->Unit, is not useful for rules in QUDT because QUDT rules cannot recognize Brick Quantities. However, I understand there is no easy solution for this problem because QuantityKinds are instances. I am curious if rdf:type can function like rdfs:subClassOf for instances in OWL rules. I will check the specification later.

QuantityKinds are indeed instances; again, see the text above. A more "direct" association e.g. through `owl:sameAs` is hard to accomplish with the hierarchical organization for Brick Quantities because the combination of `rdfs:subClassOf`, `owl:sameAs` implies that all of the subclasses of the same Quantity are the same class. I *think* this gets solved if we only associate QUDT quantitykinds with only the 'leaf' Brick quantities.

Agree that we want to maintain the hierarchical nature of Brick Quantities and to have a closer association between Quantities and QUDT QuantityKinds. The issue is how to accomplish this

jbkoh:
One "possible" tweak is to add:
```turtle
brick:hasQuantityKind    rdfs:subPropertyOf    qudt:hasQuantityKind; 
                         a                     owl:TransitiveProperty.
brick:measures           rdfs:subPropertyOf    brick:hasQuantityKind. 
brick:associatedQuantityKind    rdfs:subPropertyOf    brick:hasQuantityKind.
```
This way we can say a point instance `qudt:hasQuantityKind` a `qudt:QuantityKind` while we keep everything as is. Though it could solve our problem, there are two problems. First, "hasQuantityKind" is not logically transitive (in terms its meaning in the language). And it means that `brick:measures` should be able to accept qudt:QuantityKind too, though I don't mind it.

I am not sure about the skos approach because skos is usually for annotation and assertion but not inference. For example, existing inference engines cannot do "give me any sensors measuring Temperature" out-of-box, though we can add inference rules for skos to emulate rdfs.

gtfierro:
I like the idea of using a transitive `hasQuantityKind` relationship to provide a smoother link to the QUDT classes, but we may not have to handle this through OWL. What are our intended use cases? I can think of 2 closely related ones:

1. Validate the choice of QUDT units associated with an instance of a Brick Point class, given the Quantity/QuantityKind associated with the particular Point class.
2. Enable simple queries that list valid/possible units for a particular class

(1) is probably better handled through SHACL shapes that can validate a Brick model. (2) relies on a sensible implementation in OWL, but doesn't need to rely on any fancy inference. Providing straightforward documentation on how the associations are made and how those queries work might suffice here.

---

I've solicited some outside opinions on the general modelling approach here, and I'll report back when I've gathered that.

gtfierro:
# Units Prototype

Basic model structure is as follows

```
                brick:measures                  owl:sameAs
  ╔════════════╗      |      ╔════════════════╗      |      +-------------------+
  ║Brick Point ║════════════>║ Brick Quantity ║════════════>| QUDT QuantityKind |
  ║   Class    ║             ╚══════╦═════════╝             +---------+---------+
  ╚════════════╝                    ║                                 |
        |                           ║ qudt:applicableUnit             |
        | rdf:type                  ║ (compiled into Brick def)       |
        v                           ║                                 |
  ┌─────────┐                       v                                 |
  │Point    │ brick:hasUnit   +---------+     qudt:applicableUnit     |
  │Instance │────────────────>|QUDT Unit| <---------------------------+
  └─────────┘                 +---------+

```

- Bolded (double-line) boxes, edges are defined in the Brick schema 
- Smooth (single-line) boxes, edges are defined in a Brick model (an instance of a building)
- Dashed boxes, edges are defined in QUDT and are *not* included in the Brick distribution


## Model Overview

- Brick `Point` classes are arranged into an OWL-based class hierarchy; this is
  exactly as is modeled in Brick v1.1
  - Brick `Point` classes are associated through the `brick:measures` property with a Brick `Quantity`
- Brick `Quantity` is an OWL Class and a subclass of SKOS Concept:
  - instances of `Quantity` represent the kinds of quantities that can be measured and have a numerical value associated with them
  - instances of Brick `Quantity` are equated with QUDT `QuantityKind`s through the `owl:sameAs` property; this means they should be considered the same entity
      - as a result, Brick `Quantity` instances inherit the `qudt:applicableUnit` properties from QUDT
  - instead of being organized into an OWL-based class hierarchy, as was done in a previous draft of this model, the Brick `Quantity` instances are related through the use of `skos:broader` and `skos:narrower`
      - this preserves the generality/specificity of the original hierarchical association but without the logical complications of combining `rdfs:subClassOf` and `owl:sameAs`
  - **breaking changes**:
    - concepts like `Power` and `Temperature`, which were previously modeled as OWL classes, are no longer classes; this is not too bad of a breaking change because Brick `Quantity` concepts were punned and the "instance" flavor of the pun was used. It is very unlikely that anyone or anything was using the Brick `Quantity` concepts as classes. Brick `Substance`s are still punned and are modeled as a Brick class hierarchy as before.
- the `brick:hasUnit` property associates instances of Brick `Point`s with QUDT `Unit`s:
    - QUDT `Unit`s are full URIs, but have human-friendly labels indicated through the `qudt:symbol` property

## Example (Turtle)

Partial definition of `Quantity` and the `Power` and `Electric_Power` instances

```turtle
brick:Quantity a owl:Class ;
    rdfs:subClassOf skos:Concept,
        sosa:ObservableProperty,
        brick:Measurable .

brick:Power a brick:Quantity ;
    rdfs:label "Power" ;
    qudt:applicableUnit unit:HP,
        unit:HP_Water,
        unit:KiloW,
        unit:W ;
    owl:sameAs qudtqk:Power .

brick:Electric_Power a brick:Quantity ;
    rdfs:label "Electric Power" ;
    qudt:applicableUnit unit:KiloW ;
    owl:sameAs qudtqk:ElectricPower ;
    skos:broader brick:Power .

unit:HP qudt:symbol "HP" .
unit:HP_Water qudt:symbol "hp_water" .
unit:KiloW qudt:symbol "kW" .
unit:W qudt:symbol "W" .


brick:Power_Sensor a owl:Class ;
    rdfs:label "Power Sensor" ;
    rdfs:subClassOf [ owl:intersectionOf ( _:has_Point _:has_Sensor _:has_Power ) ],
        brick:Sensor ;
    owl:equivalentClass [ owl:intersectionOf ( [ a owl:Restriction ;
                        owl:hasValue brick:Power ;
                        owl:onProperty brick:measures ] ) ] ;
    brick:hasAssociatedTag tag:Point,
        tag:Power,
        tag:Sensor .
```

Use w/n a Brick model

```turtle
# provided
ex:power_sensor_1    a   brick:Power_Sensor ;
                    brick:hasUnit unit:KiloW .
```

signal-2020-05-25-192003

signal-2020-05-25-192233

bricksrc/properties.py
```diff
@@ -152,4 +152,16 @@
         RDFS.domain: BRICK.Building,
         RDFS.range: VCARD.Address,
     },
+    "hasUnit": {
+        A: [OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+        SKOS.definition: Literal("The QUDT unit associated with this Point"),
+        RDFS.domain: BRICK.Point,
+        RDFS.range: UNIT.Unit,
+    },
+    "associatedQuantityKind": {
```
jbkoh, line 161: This should be removed as it's no longer used I believe.

bricksrc/quantities.py
```diff
@@ -1,5 +1,32 @@
...
+g = Graph()
+g.load_file("support/VOCAB_QUDT-QUANTITY-KINDS-ALL-v2.1.ttl")
+g.load_file("support/VOCAB_QUDT-UNITS-ALL-v2.1.ttl")
+g.g.bind("qudt", QUDT)
+g.g.bind("qudtqk", QUDTQK)
+sess = BrickInferenceSession()
+g = sess.expand(g)
+
+
+def get_units(brick_quantity):
```
jbkoh, line 16: It might be useful to include unit definitions too. I have no strong opinion on this, but Brick could become a single point for metadata schema with complete definitions.

On the other hand, I am not fully convinced to include the qudt symbols. It will be useful for some use cases, but including them inside Brick schema would confuse users IMHO.

generate_brick.py
```diff
@@ -268,21 +331,38 @@ def define_properties(definitions, superprop=None):
 G.add((BRICK.Quantity, RDFS.subClassOf, SOSA.ObservableProperty))
 G.add((BRICK.Quantity, RDFS.subClassOf, BRICK.Measurable))
 G.add((BRICK.Quantity, A, OWL.Class))
+G.add((BRICK.Quantity, RDFS.subClassOf, SKOS.Concept))
 # set up Substance definition
 G.add((BRICK.Substance, RDFS.subClassOf, SOSA.FeatureOfInterest))
 G.add((BRICK.Substance, RDFS.subClassOf, BRICK.Measurable))
 G.add((BRICK.Substance, A, OWL.Class))
+G.add((BRICK.Substance, RDFS.subClassOf, SKOS.Concept))
```
jbkoh, line 339: Is Substance a skos:Concept too? I can't find a description of this change.

shacl/BrickShape.ttl
```diff
@@ -1,17 +1,21 @@
...
 @prefix sh: <http://www.w3.org/ns/shacl#> .
-@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
-@prefix sosa: <http://www.w3.org/ns/sosa#> .
-@prefix tag: <https://brickschema.org/schema/1.1/BrickTag#> .
+@prefix unit: <http://qudt.org/vocab/unit/> .
 @prefix vcard: <http://www.w3.org/2006/vcard/ns#> .
-@prefix xml: <http://www.w3.org/XML/1998/namespace> .
-@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+bsh:associatedQuantityKindDomainShape a sh:NodeShape ;
```
jbkoh, line 9: Those two shapes should be removed too, I think.

tests/test_units.py
```diff
@@ -0,0 +1,108 @@
...
+    """
+    In the current implementation, using owl:equivalentClass to align Quantity
+    with QUDT QuantityKinds, we need to make sure that through RDFS subclassing
+    that a Brick Quantity does not end up with more than 1 QuantityKind
+    """
+    g = brickschema.graph.Graph()
+    g.load_file("Brick.ttl")
+    g.g.bind("qudt", QUDT)
+    g = brickschema.inference.OWLRLInferenceSession(load_brick=False).expand(g)
+    quantity_qk = g.query(
```
jbkoh, line 23: Can this cover Brick Quantities not associated with any QUDT QuantityKinds?

tests/test_units.py
```diff
@@ -0,0 +1,108 @@
...
+    g = brickschema.inference.OWLRLInferenceSession(load_brick=False).expand(g)
+
+    # test the definitions by making sure that some quantities have applicable
+    # units
+    quantities_with_units = g.query(
+        "SELECT ?q WHERE { \
+             ?q rdf:type brick:Quantity .\
+             ?q qudt:applicableUnit ?unit}"
+    )
+    assert len(quantities_with_units) > 0
```
jbkoh, line 108: I think this test should compare the number of quantities to the number of quantities with units.

tests/test_units.py
```diff
@@ -0,0 +1,108 @@
...
+            continue
+        if "Brick" in quant and "qudt" in quantkind:
+            seen[quant].append(quantkind)
+    for quant, kindlist in seen.items():
+        assert (
+            len(kindlist) == 1
+        ), f"Quantity {quant} has more than one associated QuantityKind! {kindlist}"
+
+
+def test_instances_measure_correct_units():
```
jbkoh, line 41: Could you clarify what this test does once more? Instance--hasUnit-->qudt:Unit is generated from the class--...-->qudt:Unit, so you are comparing A to B while A is logically equivalent to B in the generation code. Could you give an example of what kind of error this can detect?

bricksrc/quantities.py
gtfierro: We can discuss this more on the call today: I think we want to minimize the amount of external information we include in the Brick definition that is replicated elsewhere -- it is easy for the number of definitions we include to grow and grow. If we start with QuantityKinds, then its Units, then its the dimensions and before we know it we have pulled in all of QUDT. This is not how the semweb world usually works -- it is more appropriate to pull in external graphs when they are needed. 

QUDT symbols are nice for knowing how to render a unit, but the symbols chosen by QUDT are not always the most obvious. We can explore including our own symbols or removing this altogether

generate_brick.py
gtfierro: I think I added this by mistake --- only looking at Brick Quantity for now. I'll remove it

tests/test_units.py
gtfierro: No. This is only testing that each Brick Quantity is only associated with a single QUDT QuantityKind.

tests/test_units.py
gtfierro: That would be a "completeness" test; this is verifying that the association worked at all

tests/test_units.py
gtfierro: Good catch! This test is a carry over from when it was class-based, so it's not  verifying anything anymore

gtfierro: I guess it provides an example of how `hasUnit` is used with Point instances -- I'll keep the implementation for now so we can build it into something more useful later

gtfierro:
> How to guide users what units they need to use? For now, they need to understand the indirection from Point class to Quantity to Unit. It'd be helpful to introduce the same qudt:applicableUnit at Point class level too. Though it's redundant, it will help users while not affecting instance-level models.

Do you mean that `qudt:applicableUnit` is applied to the Point classes instead of the Quantities? I suppose we could infer that through the `brick:measures` relationship, but I think the answer here is *documentation* rather than redundant properties

> Is skos:Concept similar to rdfs:Class in terms of rule inheritance? (It seems quite loosely defined from the documentation though.) For example, if a parent has a set of applicable units, can its child strictly have a subset of the units? If it is, I'd suggest to add a test to check whether the Concept hierarchy is valid or not in terms of applicable units. QUDT's QuantityKinds are flat, so if we introduce some hierarchy, we should verify whether it makes sense or not.

No, there is no "inheritance" with skos:Concept and skos:broader/skos:narrower, there is no larger/smaller sets of applicable units, and thus no hierarchy here.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
-        "subclasses": {
-            "CO2_Level": {},
-            "PM10_Level": {},
-            "PM25_Level": {},
-            "TVOC_Level": {},
+        "subconcepts": {
+            "CO2_Level": {
+                QUDT.applicableUnit: [UNIT.PPM],
+                QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                SKOS.definition: Literal("The concentration ratio of some substance"),
```
jbkoh, line 41: The substance should be explicitly `CO2` in the definition.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("Concentration"),
```
jbkoh, line 43: Why is it just Concentration?

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
-            "CO2_Level": {},
-            "PM10_Level": {},
-            "PM25_Level": {},
-            "TVOC_Level": {},
+        "subconcepts": {
+            "CO2_Level": {
+                QUDT.applicableUnit: [UNIT.PPM],
+                QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                SKOS.definition: Literal("The concentration ratio of some substance"),
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
```
jbkoh, line 42: What is this for? If this is useful, we should add this to everywhere but not just Quantities for consistency..

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+"""
 quantity_definitions = {
     "Air_Quality": {
-        "subclasses": {
-            "CO2_Level": {},
-            "PM10_Level": {},
-            "PM25_Level": {},
-            "TVOC_Level": {},
+        "subconcepts": {
+            "CO2_Level": {
```
jbkoh, line 38: I suggest changing `Level` to `Concentration` for all the current Level quantities. Level is vague, and Concentration is more explicit as in each of the documentations.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("Concentration"),
+                SKOS.broader: QUDTQK.Dimensionless,
+            },
+            "PM10_Level": {
+                QUDT.applicableUnit: [UNIT.PPM],
+                QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                SKOS.definition: Literal("The concentration ratio of some substance"),
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("Concentration"),
```
jbkoh, line 51: label should be updated.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                SKOS.broader: QUDTQK.Dimensionless,
+            },
+            "PM10_Level": {
+                QUDT.applicableUnit: [UNIT.PPM],
+                QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                SKOS.definition: Literal("The concentration ratio of some substance"),
```
jbkoh, line 49: definition should be updated.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
     "Enthalpy": {
         SKOS.definition: Literal(
             "(also known as heat content), thermodynamic quantity equal to the sum of the internal energy of a system plus the product of the pressure volume work done on the system. H = E + pv, where H = enthalpy or total heat content, E = internal energy of the system, p = pressure, and v = volume. (Compare to [[specific enthalpy]].)"
         ),
+        OWL.sameAs: QUDTQK["Enthalpy"],
+    },
+    "Mass": {
+        OWL.sameAs: QUDTQK["Mass"],
+        "subconcepts": {
+            "Grains": {
```
jbkoh, line 84: definition missing.

jbkoh, line 84: I feel Grains is a unit of mass. Any source of defininig it as Quantity/Property?

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M1H0T0D0"],
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("Grains"),
+                SKOS.broader: QUDTQK.Mass,
+            }
+        },
+    },
+    "Phasor": {
+        "subconcepts": {
+            "PhasorAngle": {
```
jbkoh, line 95: Should be separated by hyphen for consistency. I don't see a good reason of having an exception here.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                    UNIT.MilliARCSEC,
+                    UNIT.REV,
+                ],
+                QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                SKOS.definition: Literal("Angle component of a phasor"),
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("PhasorAngle"),
+                SKOS.broader: QUDTQK.PlaneAngle,
+            },
+            "PhasorMagnitude": {
```
jbkoh, line 115: Should be separated by a hyphen.

bricksrc/quantities.py
jbkoh, line 95: PhasorAngle is a subcomponent of Phasor. It does not mean that it is a subconcept of Phasor.

bricksrc/quantities.py
jbkoh, line 115: Do we need this? There is no specific concept of PhasorMagnitude. Just Magnitude is a subcomponent to represent a Phasor.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                        OWL.sameAs: [QUDTQK["ActivePower"], BRICK["Real_Power"]],
+                    },
+                    "Real_Power": {
+                        OWL.sameAs: [QUDTQK["ActivePower"], BRICK["Active_Power"]],
+                    },
+                    "Reactive_Power": {OWL.sameAs: QUDTQK["ReactivePower"]},
+                    "Complex_Power": {OWL.sameAs: QUDTQK["ComplexPower"]},
                 },
             },
             "Peak_Power": {
```
jbkoh, line 153: It might be a separate discussion, but I'd suggest removing vague definitions like this and related Points. (Still, Peak_Power may mean Power_Rating, which has a specific definition)

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                QUDT.hasDimensionVector: QUDTDV["A0E0L2I0M1H0T-3D0"],
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("PeakPower"),
                 SKOS.definition: Literal(
                     "Tracks the highest (peak) observed power in some interval"
                 ),
             },
-            "Thermal_Power": {},
+            "Thermal_Power": {
+                QUDT.applicableUnit: [UNIT.MilliW, UNIT.W, UNIT.KiloW, UNIT.MegaW],
```
jbkoh, line 164: MBTU per hour is also common, unfortunately.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
             "Electric_Current": {
-                "subclasses": {
-                    "Current_Angle": {},
-                    "Current_Magnitude": {},
-                    "Current_Imbalance": {},
-                    "Current_Total_Harmonic_Distortion": {},
-                    "Alternating_Current_Frequency": {},
+                OWL.sameAs: QUDTQK["ElectricCurrent"],
+                "subconcepts": {
+                    "Current_Angle": {
```
jbkoh, line 187: Current_Angle is a subcomponent of Eletric_Current, but it is not a subconcept of Current.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                            UNIT.MilliARCSEC,
+                            UNIT.REV,
+                        ],
+                        QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                        SKOS.definition: Literal("Angle of current phasor"),
+                        RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                        RDFS.label: Literal("CurrentAngle"),
+                        SKOS.broader: BRICK.PhasorAngle,
+                    },
+                    "Current_Magnitude": {
```
jbkoh, line 210: Same like Current_Angle.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                            UNIT.MilliA,
+                            UNIT.MicroA,
+                            UNIT.KiloA,
+                        ],
+                        SKOS.definition: Literal("Magnitude of current phasor"),
+                        RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                        RDFS.label: Literal("CurrentMagnitude"),
+                        SKOS.broader: BRICK.PhasorMagnitude,
+                    },
+                    "Current_Imbalance": {
```
jbkoh, line 225: This is related to Current but not a subconcept of Current.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                        SKOS.definition: Literal(
+                            "The percent deviation from average current"
+                        ),
+                        QUDT.applicableUnit: [UNIT.PERCENT],
+                        QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                        RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                        RDFS.label: Literal("CurrentImbalance"),
+                        SKOS.broader: BRICK.Dimensionless,
+                    },
+                    "Current_Total_Harmonic_Distortion": {
```
jbkoh, line 235: This is not a subconcept of `Current` but rather `Total_Harmonic_Distortion`, if exists.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                        SKOS.definition: Literal(
+                            "Measurement of harmonic distortion present in a signal defined as the sum of the powers of all harmonic components to the power of the fundamental frequency. (https://en.wikipedia.org/wiki/Total_harmonic_distortion)"
+                        ),
+                        QUDT.applicableUnit: [UNIT.PERCENT, UNIT.DeciB_M],
+                        QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                        RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                        RDFS.label: Literal("CurrentTotalHarmonicDistortion"),
+                        SKOS.broader: BRICK.Dimensionless,
+                    },
+                    "Alternating_Current_Frequency": {
```
jbkoh, line 245: This is not a subconcept of Current.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
     "Voltage": {
-        "subclasses": {
-            "Electric_Voltage": {
-                "subclasses": {
-                    "Voltage_Magnitude": {},
-                    "Voltage_Angle": {},
-                    "Voltage_Imbalance": {},
-                },
+        "subconcepts": {
+            "Voltage_Magnitude": {OWL.sameAs: QUDTQK["Voltage"]},
```
jbkoh, line 266: Voltage Magnitude and Voltage Angle are not subconcepts of Voltage.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                    UNIT.MilliARCSEC,
+                    UNIT.REV,
+                ],
+                QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                SKOS.definition: Literal("Angle of voltage phasor"),
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("VoltageAngle"),
+                SKOS.broader: BRICK.PhasorAngle,
+            },
+            "Voltage_Imbalance": {
```
jbkoh, line 290: Not a subconcept of Voltage.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                QUDT.hasDimensionVector: QUDTDV["A0E0L0I0M0H0T0D1"],
+                SKOS.definition: Literal("Direction of wind relative to North"),
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("Wind_Direction"),
+            }
+        }
+    },
+    "Energy": {
+        "subconcepts": {
+            "Electric_Energy": {OWL.sameAs: QUDTQK["Energy"]},
```
jbkoh, line 326: QUDT energy is not limited to Eletric_Energy: http://qudt.org/vocab/quantitykind/Energy.html

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+            "Thermal_Energy": {OWL.sameAs: QUDTQK["ThermalEnergy"]},
+        },
+    },
+    "Flow": {
+        OWL.sameAs: QUDTQK["VolumeFlowRate"],
+        "subconcepts": {
+            "Flow_Loss": {
```
jbkoh, line 333: Flow_Loss is not a subconcept of Flow.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                SKOS.definition: Literal(
+                    "The power per unit area of solar electromagnetic radiation incident on a surface"
+                ),
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("SolarIrradiance"),
+                SKOS.broader: BRICK.Irradiance,
+            }
+        },
+    },
     "Level": {
```
jbkoh, line 386: As in the other comment, I propose to change this to Concentration.

jbkoh, line 386: Do we need this?

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
     },
-    "Luminance": {"subclasses": {"Luminous_Flux": {}, "Luminous_Intensity": {}}},
-    "Occupancy": {"subclasses": {"Occupancy_Count": {}, "Occupancy_Percentage": {}}},
-    "Position": {},
-    "Power_Factor": {},
-    "Precipitation": {},
+    "Luminance": {
+        OWL.sameAs: QUDTQK["Luminance"],
+        "subconcepts": {
+            "Luminous_Flux": {OWL.sameAs: QUDTQK["LuminousFlux"]},
```
jbkoh, line 428: Luminous_Flux Luminous_Intensity are not a subconcept of Luminance but components

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
+                SKOS.definition: Literal(
+                    "Percent of total occupancy of space that is occupied"
+                ),
+                RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),
+                RDFS.label: Literal("Occupancy_Percentage"),
+                SKOS.broader: QUDTQK.Dimensionless,
+            },
+        }
+    },
+    "Position": {
```
jbkoh, line 453: Do we have other use cases of this than Damper_Position? The term "Position" is very generic and can mislead users to a different meaning (a person's position) than what's intended here. I'd rather just use Angle for Damper Position and remove this.

bricksrc/quantities.py
```diff
@@ -1,109 +1,565 @@
...
     "Temperature": {
-        "subclasses": {
-            "Operative_Temperature": {},
-            "Radiant_Temperature": {},
-            "Dry_Bulb_Temperature": {},
-            "Wet_Bulb_Temperature": {},
+        OWL.sameAs: QUDTQK["ThermodynamicTemperature"],
+        "subconcepts": {
+            "Operative_Temperature": {
+                QUDT.applicableUnit: [UNIT["DEG_F"], UNIT["DEG_C"], UNIT["K"]],
```
jbkoh, line 516: Can we use more generic methodology for this like Operative_Temperature's applicableUnits are same as Temperature's? (Not sure if it's true for this though.)

bricksrc/quantities.py
gtfierro: I'm not sure what this means, but it seems perhaps more appropriate to separate out the `CO2`/`PM2.5` aspects of the quantity into substances

bricksrc/quantities.py
gtfierro: When there is a `OWL.sameAs` property, the definition is provided by QUDT. Is that what you're referring to?

bricksrc/quantities.py
gtfierro: It is consistent with QUDT, which uses capitalized camel-case.  Also, do you mean hyphen (`-`) or underscore (`_`) here?

bricksrc/quantities.py
gtfierro: `Peak_Power` and `Power_Rating` are different things

bricksrc/quantities.py
gtfierro: Why not? It's a "flavor" of flow that refers to the amount of flow that is lost rather than the amount of flow that is delivered

bricksrc/quantities.py
gtfierro: What part are you referring to? It is ambiguous from the code snippet

bricksrc/quantities.py
gtfierro: There's a definition attached that clarifies that this is not for a person's position: do you think the definition needs work? This is *definitely* not a "rate". Position may be used for other moving parts such as valves

bricksrc/quantities.py
gtfierro: What do you mean by a more generic methodology? Operative Temperature needs its own definition and is a more specific concept than just regular Temperature

bricksrc/quantities.py
gtfierro: Clarified as "GrainsOfMoisture"

bricksrc/quantities.py
gtfierro: Why "unfortunately"? QUDT has support for BTU; adding MBTU could be done here or in QUDT

bricksrc/quantities.py
gtfierro: Are you sure? Total Harmonic Distortion is a quantity of electrical signals, e.g. https://en.wikipedia.org/wiki/Total_harmonic_distortion

gtfierro:
I've left questions where your comments were unclear or where GitHub's choice of lines left it ambiguous as to which specific component of the file you were referring to.

You seem to be taking issues with calling components "subconcepts" of their larger whole (e.g. is a Phasor Angle a subconcept of a Phasor). I actually haven't provided actual encoded relationships between the "subconcepts" within the Brick Quantity tree beyond what is explicitly annotated as SKOS narrower/broader, so this is still left for us to define. We had considered `skos:narrower` for moving down the hierarchy, but maybe `skos:related` would be a better option. I don't think `skos:narrower`/`skos:broader` is inappropriate either though.

The two main patterns we have are (1) a component of a larger concept (phasor magnitude, angle are components of a phasor) and (2) more specific "flavors" of a quantity (operative temperature is a more specific idea than thermodynamic temperature). The second case definitely fits the `skos:narrower`/`skos:broader` paradigm. Maybe `skos:related` would be more appropriate for the former?

bricksrc/quantities.py
jbkoh: The word Position is not only used for moving parts represented by a percentage but location too.

bricksrc/quantities.py
jbkoh: I meant `Level`.

bricksrc/quantities.py
gtfierro: Ok, but that can be a distinct term, e.g. "Location" or "coordinates". We are providing a definition here of what we mean (and what practitioners commonly mean) as Position

bricksrc/quantities.py
gtfierro: I've moved "Concentration"-oriented quantities to the Concentration part. Level is used for things like water levels, precipitation levels, etc

bricksrc/quantities.py
gtfierro: This has been confirmed by the JCI folks: Flow Loss is also measured as volume flow rate

bricksrc/quantities.py
jbkoh: You are right in that it's a quality of electrical signals, but not a type of electric current (i.e., flow rate).

bricksrc/quantities.py
jbkoh: Yes, please add MBTU-per-hour here or in QUDT.

bricksrc/quantities.py
gtfierro: Flow rate has nothing to do with electric current, so I'm not sure where that confusion is coming from. What's the suggested edit here?

bricksrc/quantities.py
jbkoh: I was referring to this: https://en.wikipedia.org/wiki/Peak_power.

bricksrc/quantities.py
jbkoh: Sorry for the confusion. You are right I meant underscores. You used underscores somewhere but here you don't. E.g., `Thermal_Power`.

bricksrc/quantities.py
gtfierro: Ok, so it sounds like it should stay? We've provided a definition that I think is clear and matches how the term is used in the built environment domain

bricksrc/quantities.py
jbkoh: I was referring to `Grains`. I don't see it is owl.sameAs something in QUDT.

bricksrc/quantities.py
gtfierro: I'm still confused what you are referring to: all of the labels follow the QUDT convention but the class names follow the Brick convention here. It looks consistent to me

bricksrc/quantities.py
jbkoh: I was referring to `RDFS.isDefinedBy: URIRef(str(BRICK).strip("#")),`

bricksrc/quantities.py
gtfierro: It wasn't clear what you were referring to. In the future, please make sure you're explicit about exactly which term/aspect you're talking about so that we don't have to have this unnecessary back and forth

bricksrc/quantities.py
jbkoh: I was referring to the exact line of `SKOS.definition: Literal("The concentration ratio of some substance"),` The definitions should say "The concentration ratio of CO2" instead of some substance.

bricksrc/quantities.py
gtfierro: This reflects the structure in the QUDT definitions; we can add it elsewhere if we think it is useful but here we are using it to distinguish between the quantities defined in Brick and those in QUDT

bricksrc/quantities.py
gtfierro: This is already done

jbkoh:
`skos:related` looks like a good option.

At the same time, I don't see enough use cases of `skos` relations here. Are they for the categorization of quantities for humans?

gtfierro:
Can you be more specific by what you mean by "use cases"? Do you mean why we are using `skos` properties to model Quantity concepts?

jbkoh:
In other words, why do we need relations between Quantities compared with the flat structure in QUDT?

gtfierro:
The idea is that related quantities are grouped by the more general concepts that they are associated with: e.g. what are the components of Voltage that can be measured? I like the use of `skos:related` here to point different Quantities toward one another, which would be more consistent than trying to impose a hierarchical structure

gtfierro:
@jbkoh I've used SKOS.related and SKOS.broader/narrower together to redefine the Quantity structure here. It should be more consistent with regards to the semantics you have in mind. Can you review this soon so that we can merge?

gtfierro:
Naming is fixed

We did have a discussion about `Level` vs `Height`: `Level` captures the context of things that are measured w/n containers, which is more specific than the general notion of `Height`. It makes more sense to measure the level of precipitation than it does to measure the "height" of precipitation

Having a test to ensure tat units exist is a good idea, but it's not even fully true for QUDT itself. The test I've provided raises a warning for all of the quantities that do not have associated units, but it is a much longer task to implement units for each of the quantities here. I think we are safe leaving the same amount of coverage as QUDT

gtfierro:
@jbkoh can you approve the PR if you're satisfied with the implementation? thanks!

Issue #130, opened 2020-04-22
Resolve On/Off, Start/Stop points


gtfierro:
Is there a significant semantic difference between them? If so, what is it? If not, can we combine these two families of points into a single one that is easier to use and make sense of?

fstinner:
Start/Stop can also be the start and stop of a control sequence. It depends on the context. On/Off is not as specific as Start/Stop, although both are often used in the same semantic way. 

In short: related, but not the same

JoelBender:
Other old school terms are Trip/Close referring to the relay connected to some operating equipment like fans and pumps.  This is why BACnet uses specific enumerations active and inactive for Binary Input/Output/Value objects because True/False is also related, but not the same.

gtfierro:
To help resolve this I'm drafting some documentation for start/stop vs on/off and also command vs status. Let me know what you think:

- `brick:Command`: A Command is a control point offering read/write functionality that directly affect the state of equipment
- (subclass of Command) `brick:On_Off_Command`: An On/Off Command controls or reports the binary status of a control loop, relay or equipment activity
- (subclass of On Off Command) `brick:Off_Command`: An Off Command controls or reports the binary 'off' status of a control loop, relay or equipment activity. It can only be used to stop/terminate/deactivate an associated equipment or process, or determine that the related entity is 'off'
- similar story for `On_Command`
- (subclass of On Off Command) `brick:Start_Stop_Command`: A Start/Stop Command controls or reports the active/inactive status of a control sequence

The `Status` variants of this are "read-only".

Does this match your understanding/expectation?

praftery:
I understand the semantic difference between start/stop and on/off is the distinction between representing a command or instruction to do something (start the motor) or reporting the state of something (the motor is on). In terms of what the controller sees, that's a binary output (e.g. energize a relay controlling a fan motor) or a binary input (measure whether a contact is open or closed). Often On/Off could be either a command (a binary output) or a state measurement (binary input), but I think Start/Stop is rarely used when it is a measurement.

I think Gabe's proposed documentation resolves this. The only grey area is as Joel notes - values of True/False can correspond to the equipment being On/Off _or vice versa_ depending on the equipment - if it is normally open or normally closed. If reversed, then the bacnet Polarity property indicates this. I think the timeseries we work with should be stored with that inversion already handled such that a value of true always represents on/start. Maybe add some clarification to the docs to that effect?

weyvogel:
[...]

Issue #34, opened 2017-10-01
RFC: Definition of High-level TagSets of Points.


jbkoh:
Based on the discussion at: https://github.com/BuildSysUniformMetadata/Brick/issues/32

### Motivation
There were no clear definitions of below highlevel Point TagSets.

### Add definitions to following TagSets
1. **Status**: State, position, or condition of an item. (adopted from ASHRAE dictionary) Technically, this value should not be arbitrarily written by users or other systems as it is an observation.
2. **Alarm**: Signal, either audible or visual, or both, that alerts an operator to an off-normal condition which requires some form of corrective action. (adopted from ASHRAE dictionary)
3. **Setpoint**: Point at which the desired property is set. (modified from ASHRAE dictionary)
4. **Sensor**: device or instrument designed to detect and measure a variable.
5. **Command**: output point that determines equipment's behavior directly and/or affects relevant operational points.

### Move Meter to Equipment
Reasoning: Meter's meaning as Point is same as Sensor. We also need the concept of Equipment measuring multiple types of values. Meter can present such things as Equipment having multiple Sensors.

jbkoh:
I am adding the abvoe definitions but the Meter issue is more complicated than I expected as there are many dependent Points named as Meter already. I would like to make the below changes.

1. Change ``Meter, Water Meter, and Steam Meter`` as ``Equipment``. Those are measuring *something* related to the target ``Resources``.
    1. Remove ``Energy Meter`` and ``Power Meter``.
    2. Introduce ``Electricity Meter`` instead.

2. Change some meters into sensors: ``Gas Usage Sensor, Power Sensor, Energy Usage Sensor, Peak Energy Today Sensor, Hot Water Flow Sensor, Steam Usage Sensor, Thermal Power Sensor``, etc.

gtfierro:
I think `Meter` should be a subclass of `Sensor`. `Meter` implies that the quantity is being measured for some sort of billing or accounting purpose. So I say lets keep those `Meter` names as-is, but make sure that `Meter` is a subclass of `Sensor`.

I like having `Energy Meter` and `Power Meter` because these unambiguously describe what they are measuring, whereas `Electricity Meter` doesn't.

Thoughts?

jbkoh:
I know that we are more used to meters as points because people have used it more than using sensors. There are a few reasons why I would like to make ``Meter`` an ``Equipment``.

[...]

PR #132, opened 2020-04-23
Clarifying Point definitions

gtfierro:
Concentrating on On/Off, Start/Stop.

This draft PR holds implementations of the suggestions captured in the discussion at https://github.com/BrickSchema/Brick/issues/130

### DIFF
--- a/bricksrc/alarm.py
+++ b/bricksrc/alarm.py
@@ -1,10 +1,13 @@
 from rdflib import Literal
-from .namespaces import BRICK, TAG, SKOS
+from .namespaces import BRICK, TAG, SKOS, RDFS
 
 alarm_definitions = {
     "Alarm": {
         SKOS.definition: Literal(
-            "Alarm points are high-priority indicators conveying non-nominal behavior"
+            "Alarm points are signals (either audible or visual) that alert an operator to an off-normal condition which requires some form of corrective action"
+        ),
+        RDFS.seeAlso: Literal(
+            "https://xp20.ashrae.org/terminology/index.php?term=alarm"
         ),
         "tags": [TAG.Point, TAG.Alarm],
         "subclasses": {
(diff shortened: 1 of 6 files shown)


jbkoh:
We already have a consensus for the Point top-classes: https://github.com/BrickSchema/Brick/issues/34. It was dropped somewhere. This PR should start from there unless we found any problem with the previous ones, which I am open to discuss.

gtfierro:
I knew we had done definitions of the top level points before! Thanks for finding those. I incorporated those definitions and the ASHRAE dictionary definitions into some new text. Let me know what you think! I tried to be explicit about input/output nature of the points. It didn't make sense any more to me to adopt the ASHRAE definitions w/o alterations because some of the ASHRAE definitions do not directly address logical 'points' vs physical things e.g. sensors

gtfierro:
@jbkoh any last edits on this before I merge it in?

jbkoh:
I was adding top classes' definitions here: https://github.com/BrickSchema/Brick/commit/cec01b0cd22a7f9590d6089998437e7a16263d2a
but it would be easier to just add them here.

Other than that, I can't see the result of our discussion that we need to introduce more explicit class for Control_Loop_On_Off_Command == Start_Stop_Command. I have no strong opinion on keeping Start_Stop_Command or not. Do we wait for Ambuj to get back with a better idea on it?

gtfierro:
We're not introducing new classes but rather clarifying the definitions and use of the ones that are already there. We can wait for additional input but we already have a few experts saying that they are fine with this classification

jbkoh:
Our previous conclusion was to be explicit as much as possible. The "experts" might be fine with it because they internalized the difference between start_stop and on_off, which might not be generally true. I can still bear with it, if you prefer the implicit way,

gtfierro:
I think we're in agreement here: `Start_Stop_{Command, Status}` is a term that is in use, hence why it appears in Brick. This PR is simply clarifying the definition of what that means in relation to existing concepts by (a) structuring the class hierarchy around what experts have agreed is a sensible interpretation of the terms, and (b) making both the structure and the actual definitions explicit in the definition of the Brick ontology. If there's other steps we can take, please let me know what you think they are and we can take them

gtfierro:
Merging as of the approval on the Brick dev meeting today

PR #135, opened 2020-05-19
Update conversion tool to use rdflib 5.0.0

gtfierro:
Should not affect any functionality in the tool. @shreyasnagare can you verify? I don't recall where we were on validating the conversion process with tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,5 @@ script:
   - python generate_brick.py
   # -s flag: do not capture output (avoids killing test after 10 min)
   - pytest -s -vvvv
+  # do integration tests
+  - cd tests/integration && bash run_integration_tests.sh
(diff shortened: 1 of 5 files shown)


shreyasnagare:
I manually tested the conversion tool, and it works fine with rdflib==5.0.0 as expected.

[This test](https://github.com/BrickSchema/Brick/blob/master/tests/test_conversions.py) runs all the SPARQL queries to make sure they aren't malformed.

I think adding a something like `cd tools/convert && python convert.py --source 1.0.2 --target 1.1 examples/ebu3b_brick.ttl` in the Travis CI test config would be enough to test the functionality of the tool.

gtfierro:
@shreyasnagare I added an integration test script that verifies that the requirements install correctly and that all of the conversions run properly

PR #138, opened 2020-06-03
add requests dependency to fix integration tests

gtfierro:
Should be straightforward to fix, but I don't know how the Travis tests weren't failing if this dependency was an issue. Don't want to merge until I know why

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ install:
   - pip install -r requirements.txt
 
 script:
+  - pip freeze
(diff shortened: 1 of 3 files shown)


gtfierro:
It looks like `requests` gets installed in the Travis environment as a dependency of one of the packages in Brick's main `requirements.txt`; there must be some special handling of dependencies in Travis so that this environment 'leaks' into the integration tests (probably because then `requests` is installed "globally"?)

Issue #139, opened 2020-06-10
Finish Brick definitions


gtfierro:
Tracking progress of definitions in @shreyasnagare's spreadsheet:

- [x] Equipment
- [x] Location
- [x] Alarm
- [x] Command
- [x] Parameter
- [x] Sensor
- [x] Setpoint
- [x] Status
- [x] Substances
- [x] Quantitites (nearly done: https://github.com/BrickSchema/Brick/pull/131)

gtfierro:
@shreyasnagare @jbkoh we're closing in on being done with the definitions here. The two outstanding elements are the discussion at https://github.com/BrickSchema/Brick/issues/142 on parameter/limit definitions and the PR at https://github.com/BrickSchema/Brick/pull/131 for the quantity definitions. Given that both of these are under discussion, how would you both feel about closing this issue and considering the draft of definitions as ready for the next stage?

I will collate the changes we proposed/suggested in the spreadsheet into the pending PR at https://github.com/BrickSchema/Brick/pull/140 as a next step. In the meantime, please take a look over #142 and #131 so we can move those forward

gtfierro:
@shreyasnagare do you have time this next week to put together a draft PR for incorporating the definitions into the Brick compilation framework? We discussed a little on the call what this might look like, but I wanted to hash the details out with you a bit more. 

I have in mind something like the following:
- `bricksrc/definitions.csv`: contains a list of all of the classes (with the full Brick namespace in front), their definitions, the 'seeAlso', etc from the spreadsheet. Maybe sorted alphabetically?
- `brick/*.py`: remove the `skos:definition` fields from the class definitions and make sure these are reflected in the spreadsheet (I think this latter part is already true)
- `generate_brick.py`: we will need to define a small function that consumes the CSV file and associates the definitions (using the `skos:definition` property) with each Brick class
- add a test that throws a warning/error if a subclass of `brick:Class` doesn't have a value for `skos:definition`

I can handle the last two items, but it would be a huge help if you were able to take the first two items if you have time. If you're slammed this week, or if you'd like to take a pass at all of them, let me know!

gtfierro:
Merged https://github.com/BrickSchema/Brick/pull/143, containing the definitions!

PR #140, opened 2020-06-11
Refining class structure and names

gtfierro:
Reflects discussion, structures, renaming being done in the definitions spreadsheet (see https://github.com/BrickSchema/Brick/issues/139).

This is a staging area for the removals, renamings and other changes.

### Changes

- `Lockout` "sensors" are actually parameters and have been moved to that hierarchy
- all air humidity sensors have been clarified as measuring *relative humidity*
- rename emergency power off, emergency air flow statuses to reflect the fact that they are statuses of systems
- moved the emergency power off, emergency air flow system status points under the `System Status` class
- reorganize start/stop and on/off status points to reflect the definitions that start/stop is for control loops and on/off can be used for equipment
- Frost sensor is also a temperature sensor

### DIFF
--- a/bricksrc/alarm.py
+++ b/bricksrc/alarm.py
@@ -86,8 +86,8 @@ alarm_definitions = {
                     },
                 },
             },
-            "Liquid_Detected_Alarm": {
-                "tags": [TAG.Point, TAG.Liquid, TAG.Detected, TAG.Alarm],
+            "Liquid_Detection_Alarm": {
+                "tags": [TAG.Point, TAG.Liquid, TAG.Detection, TAG.Alarm],
             },
             "Luminance_Alarm": {"tags": [TAG.Point, TAG.Luminance, TAG.Alarm]},
             "Maintenance_Required_Alarm": {
@@ -233,16 +233,16 @@ alarm_definitions = {
             "Smoke_Alarm": {
                 "tags": [TAG.Point, TAG.Smoke, TAG.Alarm],
                 "subclasses": {
-                    "Smoke_Detected_Alarm": {
-                        "tags": [TAG.Point, TAG.Smoke, TAG.Detected, TAG.Alarm],
+                    "Smoke_Detection_Alarm": {
+                        "tags": [TAG.Point, TAG.Smoke, TAG.Detection, TAG.Alarm],
                         "subclasses": {
-                            "Discharge_Air_Smoke_Detected_Alarm": {
+                            "Discharge_Air_Smoke_Detection_Alarm": {
                                 "tags": [
                                     TAG.Point,
                                     TAG.Discharge,
                                     TAG.Air,
                                     TAG.Smoke,
-                                    TAG.Detected,
+                                    TAG.Detection,
                                     TAG.Alarm,
                                 ],
                                 "parents": [BRICK.Air_Alarm],
--- a/bricksrc/alarm.py
+++ b/bricksrc/alarm.py
@@ -47,16 +47,6 @@ alarm_definitions = {
                     "Emergency_Generator_Alarm": {
                         "tags": [TAG.Point, TAG.Generator, TAG.Emergency, TAG.Alarm],
                     },
-                    "Emergency_Power_Loss_Alarm": {
-                        "tags": [
-                            TAG.Point,
-                            TAG.Power,
-                            TAG.Loss,
-                            TAG.Emergency,
-                            TAG.Alarm,
-                        ],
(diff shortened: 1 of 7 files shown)


gtfierro:
Removed following equipment:
- Preheat valve VFD

Removed following alarms:
- Emergency Power Loss Alarm

Removed following commands:
- highest zone cooling command
- exhaust fan fire control panel on command
- exhaust fan fire control panel off command

Removed follow sensors:
- temperature adjust sensor

Removed following setpoints:
- Thermal Energy Storage Discharge Water Differential Pressure Deadband Setpoint
- Thermal Energy Storage Supply Water Differential Pressure Deadband Setpoint
- Fan air flow setpoint
- mode setpoint
- dual band mode setpoint
- unoccupied mode setpoint
- occupied mode setpoint
- Request Setpoint
- Cooling Request Setpoint
- Cooling Request Percent Setpoint
- Heating Request Setpoint
- Heating Request Percent Setpoint
- Discharge Fan Differential Speed Setpoint
- Return Fan Differential Speed Setpoint
- Supply Fan Differential Speed Setpoint
- Max return air CO2 setpoint

gtfierro:
@shreyasnagare I think this is ready for review! Take a look through and let me know if there any errors/edits you have on the changes contained here. This should align us with the changes in the document

gtfierro:
@shreyasnagare I think this is ready for review! Take a look through and let me know if there any errors/edits you have on the changes contained here. This should align us with the changes in the document

shreyasnagare:
> @shreyasnagare I think this is ready for review! Take a look through and let me know if there any errors/edits you have on the changes contained here. This should align us with the changes in the document

This looks great. Here are some thoughts:
- Instead of removing `Max_Return_Air_CO2_Setpoint`, we can make it `Max_Return_Air_CO2_Setpoint_Limit`. Looks like we forgot to migrate this class when the `Parameter` hierarchy was created.
- We can remove `Fan_Speed_Reset_Command` since we already have `Fan` + `Speed_Reset_Command`.
- Remove other classes that are redundant/not needed. Here are a few (with comments from the sheet):
  - `Lockout_Command` - probably covered by the lockout parameters we already have.
  - `Steam_On_Off_Command` - not clear what this controls (I think this one might be useful now that we have a `Steam_System`)
  - `Curtailment_Override_Command` - not clear what this is: maybe it overrides a curtailment control strategy?
  - `Run_Request_Command` -  in favor of on command
  - `Shutdown_Command` - in favor of off command
  - `Hot_Water_Shutdown_Command` -  in favor of off command
  - `Unoccupied_Hot_Water_Shutdown_Command` -  in favor of off command

gtfierro:
Thanks for the quick turnaround and good feedback!

Removed `Fan_Speed_Reset_Command`, `Lockout_Command`, `Run_Request_Command` and `Shutdown_Command` and its children. Let's keep `Steam_On_Off_Command` and `Curtailment_Override_Command` for now to get feedback from the domain experts -- if they're redundant or nonsensical then we can remove them.

I'll wait on the discussion in #142 to figure out how to handle `Max_Return_Air_CO2_Setpoint`, but I agree it should be a parameter.

jbkoh:
Hey guys, thanks for moving this forward. I am happy with those changes. Two comments:

- I think Frost_Sensor is not a type of Temperature_Sensor, but rather it measures some phenomena related to temperature and other properties. So, it shouldn't be a subclass of Temperature_Sensor and should not measure Temperature ([reference](http://ictinternational.com/products/sf-110/sf-110-frost-sensor/)
- Regarding `Emergency_Air_Flow_System_Status`, it measures `Emergency_Air_Flow_System`'s `Status`, but `Emergency_Air_Flow_System` is never defined. Is it going to be included in another PR?

gtfierro:
@jbkoh your reference says that the Frost Sensor measures temperature --- what do you envision as a better definition?

We have no `measures` relationship for `Emergency_Air_Flow_System_Status`, and statuses do not measure anything anyway. 

@shreyasnagare can you include an `Emergency_Air_Flow_System` class in #141 ? It would likely be a sibling of `Emergency_Power_Off_System`

gtfierro:
@jbkoh any ideas for how to handle Frost? Otherwise, would you be ok with me merging this

gtfierro:
Merging for now to support the definitions work in https://github.com/BrickSchema/Brick/pull/143

jbkoh:
@gtfierro Sorry that I missed your message. I will continue this discussion at the definition thread.

PR #141, opened 2020-06-12
Add more electrical and fluid system equipments

shreyasnagare:
Adds the following equipment classes:
- Electrical_System:
    - Disconnect_Switch
    - Switchgear
    - Bus_Riser
    - Transformer
    - Motor_Control_Center
    - Breaker_Panel
- Gas_System:
    - Gas_Valve
    - Gas_Distribution
- Water_System:
    - Water_Valve
    - Water_Distribution
- Steam_System
    - Steam_Valve
    - Steam_Distribution

**Note:** All `*_Valve` classes are subclasses of `brick:Valve`


shreyasnagare:
> If we were to use the subclass approach, say for example a `Steam_Valve` is a subclass of `Steam_System`, then this would imply that instances of `Steam_Valve` are also instances of `Steam_System`, which I don't think is the semantics that we'd like to promote. One approach might be that the system classes are a somewhat distinct hierarchy:

Good catch! I have removed the subclass relationship for the fluid system components. Doesn't this translate to - instances of `Touchpanel` being instances of `Lighting_System` or instances of `Battery` being instances of `Electrical_Systems`?

> * make the `System` classes part of their own hierarchy rooted in a generic `System` class that may be a sibling or child of `Equipment` (probably sibling?)

It would be great to have a generic `System` class to organize all the `*_System` classes like `Electrical_System`, `Shading_System`, `Water_System`, etc. I have created a list of all the `*_System` classes below, in a hierarchy that made sense to me. I'm unsure about most items here, feel free to edit this.

- brick:Class
    - brick:System
        - brick:Electrical_System
            - brick:Emergency_Power_Off_System
            - brick:Lighting_System
        - brick:Fluid_System
            - brick:Steam_System
            - brick:Gas_System
            - brick:Water_System
                - brick:Chilled_Water_System
                - brick:Hot_Water_System
                    - brick:Domestic_Hot_Water_System
        - brick:Shading_System
        - brick:Heating_Ventilation_Air_Conditioning_System
        - brick:Fire_Safety_System

gtfierro:
> Good catch! I have removed the subclass relationship for the fluid system components. Doesn't this translate to - instances of Touchpanel being instances of Lighting_System or instances of Battery being instances of Electrical_Systems?

Yes, this carries over into other parts of the class structure as well. It's been bugging me for awhile (a touchpanel is part of a lighting system, it isn't a type of lighting system!), so this feels like a good opportunity to fix it :)

As another cut at the System hierarchy:

- brick:Class:
  - brick:System:
    - brick:Electrical_System
    - brick:Lighting_System
    - brick:Gas_System
    - brick:Heating_Ventilation_Air_Conditioning_System:
      - brick:Steam_System
      - brick:Water_System
        - brick:Chilled_Water_System
        - brick:Hot_Water_System:
          - brick:Domestic_Hot_Water_System
    - brick:Shading_System
    - brick:Safety_System:
      - brick:Fire_Safety_System
      - brick:Emergency_Power_Off_System

This organizes systems around what they are used for rather than what they primarily operate on -- it's a little iffy with Domestic Hot Water being there, admittedly. Given that there are not many kinds of systems, we don't have a lot to lose by making the hierarchy flatter -- there isn't a huge amount of complexity that can be mitigated by organizing the concepts hierarchically.

JoelBender:
This aligns precisely with the direction I'm taking in the 223 proposal (think sub-classes of a connection), so a huge +1 from me.

As for water systems, there is a big distinction in composition between water used for HVAC applications and domestic applications so those should be disjoint.  Hot water for HVAC is subdivided into preheat, reheat, radiation and heat recovery, water used for ground source/geothermal applications is something else again.  I'm not sure where steam condensate fits, it's also hot water.

gtfierro:
Great to hear! We'll be discussing this on the Brick development call tomorrow so I'll update with some of our thoughts.

An adjusted hierarchy:

- brick:Class:
    - brick:System:
        - brick:Domestic_Hot_Water_System
		- brick:Electrical_System
		- brick:Gas_System
		- brick:Heating_Ventilation_Air_Conditioning_System:
            - brick:Steam_System
            - brick:Water_System
                - brick:Chilled_Water_System
				- brick:Hot_Water_System:
                    - brick:Preheat_Hot_Water_System
                    - brick:Reheat_Hot_Water_System
                    - brick:Radiation_Hot_Water_System
                    - brick:Heat_Recovery_Hot_Water_System
		- brick:Lighting_System
		- brick:Safety_System:
            - brick:Fire_Safety_System
			- brick:Emergency_Power_Off_System
		- brick:Shading_System

bricksrc/equipment.py
```diff
@@ -6,32 +6,61 @@
...
-        "tags": [TAG.Heat, TAG.Ventilation, TAG.Air, TAG.Conditioning, TAG.System],
-    },
-    "Heating_Ventilation_Air_Conditioning_System": {
-        OWL.equivalentClass: BRICK["HVAC"],
-        "tags": [TAG.HVAC],
+        "tags": [TAG.HVAC]
     },
     "Weather": {"tags": [TAG.Weather]},
-    "Electrical_System": {
-        "tags": [TAG.Electrical, TAG.System],
```
gtfierro, line 18: I think we should replace `Electrical System` with `Electrical Equipment`, or something similar to group the more specific equipment types

bricksrc/equipment.py
```diff
@@ -126,57 +147,55 @@
...
+        SKOS.definition: Literal(
+            "Utilize a water distribution source to represent how water is distributed across multiple destinations (pipes)"
+        ),
+    },
+    "Steam_Distribution": {
+        "tags": [TAG.Steam, TAG.Distribution, TAG.Equipment],
+        SKOS.definition: Literal(
+            "Utilize a steam distribution source to represent how steam is distributed across multiple destinations"
+        ),
+    },
```
gtfierro, line 161: Where do `Water Distribution` and `Steam Distribution` come from? These might have more to do with the System modeling than with the equipment

bricksrc/equipment.py
```diff
@@ -126,57 +147,55 @@
...
+    "Lighting": {
+        "subclasses": {
+            "Luminaire": {"tags": [TAG.Luminaire, TAG.Equipment]},
+            "Luminaire_Driver": {
+                "tags": [TAG.Luminaire, TAG.Driver, TAG.Equipment],
+            },
+        },
     },
-    "Lighting_System": {
-        "tags": [TAG.Lighting, TAG.Equipment],
```
gtfierro, line 136: Similar story here: we should have a `Lighting Equpiment` class to hold all of the lighting-related types

bricksrc/system.py
```diff
@@ -0,0 +1,74 @@
...
+                        }
+                    },
+                    "CWS": {
+                        OWL.equivalentClass: BRICK["Chilled_Water_System"],
+                        "tags": [TAG.CWS],
+                    },
+                    "HWS": {
+                        OWL.equivalentClass: BRICK["Hot_Water_System"],
+                        "tags": [TAG.HWS]
+                    },
```
gtfierro, line 52: I feel like we should remove these: now that we're defining out own class structure here, it's less important to provide the abbreviations. When we were modelling the equipment classes before, we wanted to preserve some naming conventions, but I think that's less important for the System stuff (because these are not necessarily explicitly named in the BMS)

gtfierro:
@shreyasnagare will you have cycles to finish this up this week? I think there's just a few remaining changes

bricksrc/system.py
shreyasnagare: Do you think we should remove `brick:FCP` as well?

bricksrc/equipment.py
shreyasnagare: Do you think we should remove them for now?

bricksrc/system.py
gtfierro: Yeah, good idea

bricksrc/equipment.py
gtfierro: Actually, let's keep for now, but we should consider what possible subclasses would be

bricksrc/equipment.py
```diff
@@ -448,21 +468,35 @@
...
-                "parents": [BRICK.Hot_Water_System, BRICK.Water_Valve],
+                "parents": [BRICK.Water_Valve],
             },
         },
     },
     "Cooling_Valve": {"tags": [TAG.Valve, TAG.Cool, TAG.Equipment]},
     "Water_Valve": {
         "tags": [TAG.Valve, TAG.Water, TAG.Equipment],
+        SKOS.definition: Literal(
+            "Building water is provided by utility company and sourced through a master building feed."
```
gtfierro, line 476: This seems more appropriate for a definition of the substance itself rather than the valve; the valve mediates the flow of water

bricksrc/equipment.py
```diff
@@ -448,21 +468,35 @@
...
             "Chilled_Water_Valve": {
                 "tags": [TAG.Chilled, TAG.Valve, TAG.Water, TAG.Equipment],
-                "parents": [BRICK.Chilled_Water_System],
             },
         },
     },
+    "Gas_Valve": {
+        "tags": [TAG.Gas, TAG.Valve, TAG.Equipment],
+        SKOS.definition: Literal(
+            "Building gas is provided by utility company and sourced through a master building feed."
```
gtfierro, line 487: Same thing here: this is a definition for Gas, not a Gas Valve

bricksrc/equipment.py
```diff
@@ -448,21 +468,35 @@
...
+        ),
+    },
     "Isolation_Valve": {"tags": [TAG.Isolation, TAG.Valve, TAG.Equipment]},
+    "Steam_Valve": {
+        "tags": [TAG.Steam, TAG.Valve, TAG.Equipment],
+        SKOS.definition: Literal(
+            "Building steam is provided by the utility company and sourced through a master building feed."
```
gtfierro, line 494: Same thing here: this is a definition for Steam, not a Steam Valve. In general we also can't assume that steam is provided by the utility company (it is often produced on site), so I would just remove this.

shreyasnagare:
> Couple small changes but this is close!

I've removed the incorrect definitions for now, we can find the correct ones and add them to the CSV file later. Do you have any other changes in mind? I believe this PR also closes #100.

gtfierro:
That should be it! Let me know if you need any help doing the merge or getting the tests to pass; I can merge once that's finished

shreyasnagare:
> That should be it! Let me know if you need any help doing the merge or getting the tests to pass; I can merge once that's finished

@gtfierro I have resolved all the merge conflicts already. However, I won't be able to get the tests to pass atm. Unfortunately, I don't have access to a unix-based operating system or a decent internet connection right now which is why I won't be able to generate brick.ttl, run the tests locally or look at the test dumps.

I have tried to run the tests on an OpenShift playground anonymously. For some reason, this required me to install `brickschema[reasonable]`, after which, I was able to generate_brick and run the `test_definitions.py` tests successfully (no errors). I don't understand why these won't work with the default reasoner. Can you take look at the failing tests?

gtfierro:
For sure! I'll take a look. The `reasonable` reasoner is *mostly* compliant but not 100% (still a few left to implement: https://github.com/gtfierro/reasonable#owl-2-rules) which might explain the issues. I'll fix up the tests and perform the rest of the merge.

Thanks for digging into this even though you don't have access to the usual dev environment!

shreyasnagare:
Thanks for understanding. I'm undergoing an institutional quarantine right now.

I have simplified the test query for finding the obsolete definitions. Now the only requirement is - "for `?x` to have a definition/external reference, there must exist a triple `?x a ?y .` in `Brick.ttl`." If this is too much work, we can convert this AssertionError to a UserWarning for now.

PR #143, opened 2020-06-26
Incorporate the definitions into the Brick compilation framework

### LINKED ISSUE #139: Finish Brick definitions

shreyasnagare:
This adds a unified `definitions.csv` file that contains definitions for Brick terms which is then parsed and compiled into `Brick.ttl`.

 - [x] Adds definitions as `Literals` with `@en` language tags through `skos:definition`. 
 - [x] Adds related source information through `rdfs:seeAlso` (as `URIRef`).
 - [x] Adds the following tests for definitions:
   - [x] All Brick classes **should** have definitions. The list of classes is saved as `./tests/classes_no_definitions.json`.
   - [x] All Brick relationships **should** have definitions. The list of classes is saved as `./tests/relationships_no_definitions.json`.
   - [x] Obsolete definitions **mustn't** be present in the ontology. The list of terms is saved as `./tests/obsolete_definitions.json`.
 - [x] Removes definitions from `./bricksrc/*.py` files.
 - [x] Verify definitions in `definitions.csv`.
 - [x] Add definitions for properties.

Resolves #139.


jbkoh:
@shreyasnagare I added logic to automatically add Limit definitions. Please review it and remove hand-written definitions for those. (I wanted to do it myself but don't know how `definitions.csv` was generated.)

generate_brick.py
```diff
@@ -242,6 +242,53 @@ def add_definitions():
...
+    select ?param where {
+      ?param rdfs:subClassOf* brick:Parameter.
+    }
+    """
+    params = [row["param"] for row in G.query(qstr)]
+    for param in params:
+        words = param.split("#")[-1].split("_")
+        prefix = words[0]
+        suffix = words[-1]
+        limit_def_template = "A parameter that places {direction} bound on the range of permitted values of a {setpoint}."
```
gtfierro: @jbkoh you should be querying for subclasses of `brick:Limit` rather than doing string parsing. In general, all code that operates on the class hierarchy should be working with one of the structured forms (the Python dicts or the RDF model) and not the class names

generate_brick.py
```diff
@@ -242,6 +242,53 @@ def add_definitions():
...
+                setpoint = "Setpoint"
+            else:
+                if prefix:
+                    setpoint = "_".join(words[1:-1])
+                else:
+                    setpoint = "_".join(words[:-1])
+
+            if setpoint.split("_")[-1] != "Setpoint":
+                setpoint = setpoint + "_Setpoint"
+                print(f"Created setpoint: {setpoint}")
```
gtfierro: Can you leave some comments clarifying what this is doing? I'm not sure why we are parsing out the word "Setpoint" from the class name. Also in general `print` statements should be replaced with calls to `logging`

generate_brick.py
jbkoh: I think you are looking at an outdated commit. I am using `Limit`.

generate_brick.py
jbkoh: Fixed this in the current commit.

`Limits` are associated with Setpoints, but some class names do not explicitly have associated `Setpoints` in their names. Thus, I manually add them here as a first attempt to fix such issues. I am open to suggestions. E.g., I say the definition of `Close_Limit` is "a boundary of `Close_Setpoint`", and throws a warning if `Close_Setpoint` does not exist.

gtfierro: Isn't this backwards? I thought the idea was to take the set of `Setpoint` classes that are defined and then generate the `Limit` classes for those --- rather than attempting to align the two sets of classes. This way, the set of `Limit` classes is always aligned to the `Setpoint` classes we have defined.

jbkoh: I intended to automatically generate definitions for existing Limits. I think not all the Setpoints should be explicitly associated with Limits.

JoelBender:
I'm confused about what is being constrained, the measured variable (supply air temperature) or its setpoint (supply air temperature setpoint).  The former (supply air temperature high limit) is an alarm condition parameter that might indicate a fault in a process (steam valve failed wide open) and the latter (supply air temperature setpoint high limit) seems like a constraint, maybe an alarm condition, but most likely a system design consideration.

The (zone temperature high limit) in my house is the hottest I'm willing to accept as comfortable, the (zone temperature setpoint high limit) is the highest value that I will accept the thermostat to be set before yelling at the kids to not crank it up so high.

generate_brick.py
gtfierro: Okay, sounds good. We'll need the autogeneration of definitions either way

gtfierro:
Good question, @JoelBender . The intention with the `Limit` classes is to model the constraint on the **setpoint**; the `Alarm` classes would configure the values used in monitoring/alarming on the measured variable. Therefore, the names of classes referencing the limits of setpoints should be called something like `...._Setpoint_Limit`

gtfierro:
@shreyasnagare is this ready for review? This is the last item before the v1.1 release :)

shreyasnagare:
> @shreyasnagare is this ready for review? This is the last item before the v1.1 release :)

Yes, it is.

jbkoh:
Continuing [a discussion](https://github.com/BrickSchema/Brick/pull/140#issuecomment-650360635) in a previous thread.

@gtfierro 
- Regarding `Frost_Sensor`, if you read [the description](http://ictinternational.com/products/sf-110/sf-110-frost-sensor/) carefully, it is measuring two temperatures to infer/predict frost. Frost_Sensor detects frost.
- Have we had `Emergency_Air_Flow_System` before? What's the source of `Emergency_Air_Flow_System_Status`? I am not sure if it'd ever existed. I am fine with adding this class, though not sure if it's real.

gtfierro:
> Regarding Frost_Sensor, if you read the description carefully, it is measuring two temperatures to infer/predict frost. Frost_Sensor detects frost.

If you're in the business of reading the description carefully, then you'll note that the output of a Frost sensor is indeed two voltages corresponding to the two temperatures. It may be the right approach to have two subclasses which clarify which of the two temperature streams the Frost sensor refers to, but it still seems that measuring Temperature is the right way to model the sensors

jbkoh:
The sensor uses the two temperature values to let users "predict" or approximate frost. Maybe we can name it `Temperature-based_Frost_Sensor` and make it subclass both Frost_Sensor and Temperature_Sensor. 

Here's a different example not using temperature: https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwjHudDfkbLqAhXRu54KHekJDiMQFjAAegQIBBAB&url=https%3A%2F%2Fwww.ti.com%2Flit%2Fpdf%2Ftidud79&usg=AOvVaw1tN4LDrVtcndV9XRfOTVG1
Some of the related texts from the above link:

- "Other techniques use temperature-sensitive devices to measure the temperature of the refrigerating medium and signal the defrost cycle upon a predetermined rise in the temperature as detected by the device. These temperature-sensitive devices are generally inaccurate because they must be extremely sensitive to detect a small change in the temperature of the refrigerating medium, which corresponds to a wide variation in the thickness of ice or frost accumulation."
- "The FDC2214 capacitive-to-digital converter from Texas Instruments enables the sensing of frost and ice. By mounting a capacitive sensor around the tube or fin of a cooling body, the system can sense the thickness of frost and ice by measuring the capacitance change between the sensor and tube/fin."

Obviously, I don't argue one is better than the other, but temperature is not the only measure for frost detection.

gtfierro:
Let's just keep it at "measures Frost" for now, then

PR #159, opened 2020-07-20
update Heating_Command definition

corymosiman12:


gtfierro:
Looks like I forgot the global flag on the regex! Good catch and thanks for the PR

PR #160, opened 2020-07-20
Update definitions for heat-exchanger, coil

gtfierro:
Heat exchanger definition updated to: `A heat exchanger is a piece of equipment built for efficient heat transfer from one medium to another. The media may be separated by a solid wall to prevent mixing or they may be in direct contact` (BEDES)

Coil definition updated to: `Cooling or heating element made of pipe or tube that may or may not be finned and formed into helical or serpentine shape` (ASHRAE)

Also made coils a subclass of heat-exchanger to reflect how they are typically classified


corymosiman12:
@gtfierro do you have a formatter to run your files through?  This commit seems to have reformatted a bunch of things unrelated to the PR.  The definitions seem good to me.

gtfierro:
Yep! Check out the [CONTRIBUTING](https://github.com/BrickSchema/Brick/blob/master/CONTRIBUTING.md) document for instructions on how to set up the development environment. If it's suggesting formatting changes, I usually just include those in the commit anyway.

PR #161, opened 2020-07-20
Weather -> Weather_Station with associated definition

gtfierro:
Weather Station makes more sense as an Equipment than old "Weather" did, which was inherited from Brick v1.0.3. Using `A dedicated weather measurement station` as the definition for now, but open to other more descriptive definitions


corymosiman12:
Looks good

PR #162, opened 2020-07-22
Move Booster fan to be sibling of Supply fan rather than child

gtfierro:
May also handle supply/discharge fan classes, depending on how we want to handle that.


gtfierro:
@jbkoh do you remember the outcome of older conversations about Discharge vs Supply Fans? Did we come down one way or the other on whether or not to mark them as equivalent classes?

jbkoh:
@gtfierro We decided to have them equivalent. I guess we didn't include them in the 1.1 release.

gtfierro:
@jbkoh Some things were bound to fall through the cracks given that we essentially rebuilt the whole thing from the ground up. I'll put in an issue for handling supply vs discharge equivalencies where they make sense

Issue #152, opened 2020-07-20
Filter hierarchy and definition


corymosiman12:
Your definition should probably also include the (3) definition from ASHRAE - `device to remove solid material from a fluid`.  I would then probably subclass an `Air_Filter`, then finally a `Mixed_Air_Filter`.

Might want to also consider the concepts of `Pre_Filter` and `Final_Filter`, as these are also important concepts / distinctions.  Unfortunately I didn't find a good definition for `Final_Filter`...  

Prefilter:
- https://qr.ae/pNs25y

gtfierro:
Thanks for the issue! Took a stab at the definitions for those concepts in https://github.com/BrickSchema/Brick/pull/163

gtfierro:
Fixed in https://github.com/BrickSchema/Brick/pull/163

PR #163, opened 2020-07-22
Add prefilter, final filter and bolster filter definition

gtfierro:
Addressing discussion in https://github.com/BrickSchema/Brick/issues/152


corymosiman12:
Seems reasonable to me :) 👍

PR #164, opened 2020-07-22
Fix VFD definition

gtfierro:
- [X] Wrong definition was copied in for VFD
- [X] added citation to ASHRAE data dictionary
- [x] move VFD out of HVAC Equipment hierarchy as a subclass of a more generic Motor class


corymosiman12:
Hey Gabe - my understanding is that a VFD is actually not a type of motor.  It is instead a motor controller.  Whether or not we want to get that technical is another question, but just wanted to put it out there.

gtfierro:
Haystack currently models a VFD as a kind of Motor, so I was just being consistent with that choice. I think for the kind of analytics we are targeting, conflating the driver with the motor itself seems to be a common design decision

corymosiman12:
Ah - I see.  Thanks for the clarification

Issue #157, opened 2020-07-20
Direction_Command


corymosiman12:
Currently this is the same as `Disable_Command`, which should be changed.

PR #168, opened 2020-08-06
Small Fixes

gtfierro:
Collection of fixes that don't necessarily merit their own PRs:

- [X] add return air differential pressure sensor
~~- [ ] add other differential pressure sensor "flavors"~~ can't find the earlier reference I was going to use
- [x] constant air volume box (cav)
- [x] carbon monoxide (co)
- [X] add cooling tower
- [X] define heat wheel
- [X] add "air handling unit' as more common term (the existing "air handler unit" class isn't quite right, but we aren't deprecating it)
- [X] fix direction command  (https://github.com/BrickSchema/Brick/issues/157)
- [x] supply/discharge equivalencies:
    - [X] alarm
    - [x] substances
    - [x] sensor
    - [x] status
    - [x] setpoint
    - [x] parameter


s0301132:
Hi @gtfierro, so you will add a cooling tower tag in Brick.ttl later?

gtfierro:
Yes, for sure! It is a long-standing oversight that I finally had time to fix. Are there properties/types of cooling towers that you would like to see added to Brick as well?

s0301132:
@gtfierro for now I just need cooling tower tag so I can set relationship between cooling tower fan and cooling tower, Thanks!

gtfierro:
Wonderful! Feel free to use `brick:Cooling_Tower` as that class for now.

jbkoh (approved):
All look good. A collection of small changes is not small at all. Great!

PR #174, opened 2020-09-02
VBIS alignment

gtfierro:
Defines the alignment TTL file from the exported 1st sheet of the alignment document.

@jbkoh would be great to get your input on this

TODOs:
- [x] remove VBIS version information (they have none yet)


jbkoh:
@gtfierro Sorry for the late response, and thanks for porting the CSV file.

1. 
I think it looks good in general, but have a meta point to discuss: Class-level modeling vs Instance-level modeling. I agree with the current way of alignment, but want to briefly discuss this for the record.

The current alignment is at the Class level; any instances won't be directly related to VBIS tags, but Classes are annotated with VBIS tags. Any querier interested in VBIS tag should look up the class and its associated VBIS tag. So, technically we don't use the power of graph traversal or "reasoning" for VBIS tags.

On the other hand, if we model VBIS tag at instance level similar to Brick Tags, we can know what VBIS tags are associated with actual instances and they are directly queriable with VBIS tags. However, it requires users to use "reasoning" to get the instances' association with VBIS tags.

That being said, VBIS tag does not reside in the RDF world, and we'd want to use this alignment as a reference/annotation instead of querying (e.g., please return all the instances with `ME-ATU-VAV-SDHC` VBIS tag.)

2. 
The compatibility check in terms of the class hierarchy. Though it's obvious in many cases, it'd be safe to have a test case to check whether the VBIS specialization has the same structure as Brick's subclass hierarchy. For example,
```turtle
brick:Variable_Air_Volume_Box vbisalign:hasVBISTag "ME-ATU-VAV*" .

brick:Variable_Air_Volume_Box_with_Reheat vbisalign:hasVBISTag "ME-ATU-VAV-SDEH",
        "ME-ATU-VAV-SDHC" .
```
This needs to be better checked. (If we model VBIS tags at instance level, we can do the check automatically within the OWL reasoning framework, but I'm fine with custom logic in Python.)


4. 
I see that `*` is in several places but only for suffixes. Is there any specific reason for not using it for prefixes? E.g., *-me-*.

5.
While I was writing those thoughts, I got curious whether each level in VBIS has a specific meaning other than just specialization level. If level has some meaning, we'd better separate them out into multiple tags in this alignment, but I think it's not the case. Please confirm if my understanding is correct.

6. 
`brick:Supply_Fan vbisalign:hasVBISTag "ME-Fa*" .` --> Is this correct? I understand VBIS doesn't model `Supply_` but only asset types, but if we want to mix two worlds, we should be consistent in either 1) not explicitly associating them or 2) completely associating them.

Rearticulating 1) and 2)
1) `ME-FA*` should not be associated with `brick:Supply_Fan` but only `brick:Fan`. Then, users should judge by themselves whether they want to look up a target Class's parents for VBIS tags or not.
OR
2) `ME-FA*` should be applied to all the subclasses of the target class for completeness, which should be considered in the generation code. So, if there is some hypothetical class called `brick:Special_Supply_Fan`, it should be also associated with `ME-FA*` even though it may not be in the spreadsheet. 

I'm fine either way, but prefer the second option for ease of use. 


Actual mapping looks good, and I'm willing to discuss the above agenda further. Thanks!

JoelBender:
SHACL supports regular expression matching (`sh:pattern` based on SPARQL REGEX function, in turn based on [XPath](https://www.w3.org/TR/xpath-functions/#regex-syntax)), so maybe a SHACL engine could help with this mapping as a separate rule/shape file.

gtfierro:
Trying out using SHACL patterns for this. Current model now encodes Brick classes as SHACL shapes whose values of `hasVBISTag` now must match the given patterns:

Simple example
```ttl
brick:Variable_Air_Volume_Box a sh:NodeShape ;
    sh:property [ a sh:PropertyShape ;
            sh:path vbisalign:hasVBISTag ;
            sh:pattern "ME-ATU-VAV*" ] .
```

Example where there are multiple compatible patterns:
```ttl
brick:Variable_Air_Volume_Box_with_Reheat a sh:NodeShape ;
    sh:or ( [ a sh:PropertyShape ;
                sh:path vbisalign:hasVBISTag ;
                sh:pattern "ME-ATU-VAV-SDHC" ] [ a sh:PropertyShape ;
                sh:path vbisalign:hasVBISTag ;
                sh:pattern "ME-ATU-VAV-SDEH" ] ) .
```

TODOs:
- [ ] is it ok to have a mapping that does not infer/generate the Brick class or VBIS tag from one representation vs another? This current implementation just validates
- [ ] make sure that the correct patterns (e.g. `*`) are added

JoelBender:
I believe that (1) it is preferable to not generate Brick classes or tags, alignment belongs as it's own thing and I would move these alignment shapes to their own namespace and work with VBIS to create and publish the model as RDF Linked Data.  For example:
```
brick:Water_Distribution a sh:NodeShape ;
    sh:property [ a sh:PropertyShape ;
            sh:path vbisalign:hasVBISTag ;
            sh:pattern "HY*" ] .
```
Becomes:
```
vbisalign:Water_Distribution a sh:NodeShape ;
    sh:targetClass brick:Water_Distribution ;
    sh:property [ a sh:PropertyShape ;
            sh:path vbis:hasTag ;
            sh:pattern "HY*" ] .
```
The idea that there can be some `?thing vbis:hasTag "ME-ATU-VAV-SDHC"` and have some engine generate the appropriate `?thing a brick:Variable_Air_Volume_Box_with_Reheat` is intriguing!

gtfierro:
@JoelBender I can get on board with making sure that we generate different shapes for the alignment rather than attaching them directly to the Brick namespace, and I've amended the prototype to match your suggestion (modulo the changes required to make VBIS "patterns" align with XML Schema-style regular expressions as required by SHACL)

My concern now is that the SHACL shapes themselves are not a convenient structure to drive a tool that can perform the required VBIS -> Brick inference. If we have a fully-qualified VBIS tag, then we can perform inference in OWL as follows:

```ttl
brick:Variable_Air_Volume_Box_with_Reheat   owl:equivalentClass [
    a owl:Restriction ;
    owl:hasValue  "ME-ATU-VAV-SDHC" ;
    owl:onProperty vbisalign:hasVBISTag
] .
```

but in the case where it is just the regular expression patterns, it would be convenient to have a `vbisalign:possibleVBISTag` property (or an equivalent but better named alternative) that just stored the relation between Brick classes and VBIS patterns w/o validating them:

```ttl
brick:Meter vbisalign:possibleVBISTag
  "EP-Me*", "BMC-ME*", "HY-Me*", "BMC-FM*" .
```

then that could be picked up by another tool (say `brickschema`) that uses the VBIS patterns to infer the Brick class, which could then be verified by the SHACL shapes

gtfierro:
This looks to be in a state where we are good to merge. There will be an upcoming PR on https://github.com/BrickSchema/py-brickschema using this alignment file to drive bi-directional Brick/VBIS inference

PR #180, opened 2020-09-18
Add hasSite and isSiteOf relationships

### LINKED ISSUE #177: Add isSiteOf/hasSite
We prototyped this in earlier papers (e.g. Mortar).

Goal is to include a relationship that captures which individuals in a graph are associated with a Site.

shreyasnagare:
The new properties can be used to capture relationships between individuals in a graph and sites associated with them.

Closes #177

### DIFF
--- a/shacl/BrickShape.ttl
+++ b/shacl/BrickShape.ttl
@@ -71,13 +71,19 @@ bsh:hasPointRangeShape a sh:NodeShape ;
 
 bsh:hasSiteHasLocationShape a sh:NodeShape ;
     sh:property [ sh:class brick:Site ;
-            sh:message "Subject of property hasSite has object with incorrect type" ;
+            sh:message "Subject of property isSiteOf has object with incorrect type" ;
             sh:path brick:hasLocation ],
         [ sh:class brick:Site ;
-            sh:message "Subject of property isSiteOf has object with incorrect type" ;
+            sh:message "Subject of property hasSite has object with incorrect type" ;
             sh:path brick:hasLocation ] ;
     sh:targetSubjectsOf brick:hasSite,
         brick:isSiteOf .
+
+bsh:hasSiteRangeShape a sh:NodeShape ;
+    sh:property [ sh:class brick:Site ;
+            sh:message "Property hasSite has object with incorrect type" ;
+            sh:path brick:hasSite ] ;
+    sh:targetSubjectsOf brick:hasSite .
 
 bsh:hasTagRangeShape a sh:NodeShape ;
     sh:property [ sh:class brick:Tag ;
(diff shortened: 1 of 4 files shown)


shacl/BrickShape.ttl
```diff
@@ -69,6 +69,16 @@ bsh:hasPointRangeShape a sh:NodeShape ;
             sh:path brick:hasPoint ] ;
     sh:targetSubjectsOf brick:hasPoint .
 
+bsh:hasSiteHasLocationShape a sh:NodeShape ;
+    sh:property [ sh:class brick:Site ;
+            sh:message "Subject of property hasSite has object with incorrect type" ;
+            sh:path brick:hasLocation ],
+        [ sh:class brick:Site ;
+            sh:message "Subject of property isSiteOf has object with incorrect type" ;
+            sh:path brick:hasLocation ] ;
```
gtfierro, line 78: It looks like these are getting added twice when they should just be there once --- can you investigate? It also looks like it is targeting subjects of both hasSite and isSiteOf, which is incorrect

gtfierro:
I was looking through the SHACL generation script and the subproperty handling is pretty clunky. I don't think it needs to be this complex for what we're doing. A simple recursive approach to handling subproperties is probably sufficient. The following seems to work ok:

```python
def addPropertyShapes(propertyName, defn):
    for pred, obj in defn.items():
        # We are only concerned with properties that have RDFS.domain or .range
        # predicate.  The temporary replacements for those predicates are
        # brick:expectedDomain and :expectedRange.  See properties.py for
        # explanation.

        if pred == RDFS.domain:
            addDomainShape(propertyName, obj)

        if pred == RDFS.range:
            addRangeShape(propertyName, obj)

        # subproperties are considered after "domain" and "range" are counted for
        if pred == "subproperties":
            for subprop, desc in obj.items():
                addPropertyShapes(subprop, desc)

for name, defn in property_definitions.items():
    addPropertyShapes(name, defn)
```

This won't handle everything that results from the `feedsAir` subproperty but should be sufficient for the site relationships

gtfierro (changes requested):
As far as the ontology part is concerned, I think this is great. We just need to iron out a couple of the details for the SHACL stuff

gtfierro:
@shreyasnagare how is this going? Anything I can help with in terms of getting the modeling to work?

shreyasnagare:
> It also looks like it is targeting subjects of both hasSite and isSiteOf, which is incorrect

I looked at the version of BrickShape.ttl in the master branch and it looks like this had always been the case even for the other property shapes. This PR should fix that.

shacl/generate_shacl.py
```diff
@@ -40,7 +40,7 @@ def addRangeShape(propertyName, expectedType):
     G.add((BSH[shapename], A, SH.NodeShape))
     G.add((sh_prop, SH["path"], BRICK[propertyName]))
     G.add((sh_prop, SH["class"], expectedType))
-    G.add((BSH[shapename], SH.targetSubjectsOf, BRICK[propertyName]))
+    G.add((BSH[shapename], SH.targetObjectsOf, BRICK[propertyName]))
```
gtfierro, line 43: I'm pretty sure this needs to be `targetSubjectsOf` because the `sh:property` property will target the objects of the target property (the property this SHACL shape is about) for us.

gtfierro (changes requested):
Much cleaner than what we had before! Nice work. Pointed out a small bug with the generated shapes but we should be good to merge soon.

shacl/generate_shacl.py
gtfierro:
```ttl
bsh:hasSiteRangeShape a sh:NodeShape ;
    sh:property [ sh:class brick:Site ;
            sh:message "Property hasSite has object with incorrect type" ;
            sh:path brick:hasSite ] ;
    sh:targetObjectsOf brick:hasSite .
```

This would be shape that was generated under the existing code. Because of `sh:targetObjectsof`, this shape applies to the `Site` instances! Which don't have `hasSite` properties.

epaulson:
I am not a fan of isSiteOf/hasSite. I would prefer that we just treat it as any other location class, and use hasPart for location composition and hasLocation to put assets into whatever nested location makes sense.

corymosiman12:
I would agree with Eric - it basically gets away from what was a core design concept in my opinion of Brick

gtfierro:
Cool cool -- this matches what we discussed in the dev meeting today.

@shreyasnagare can you remove the `isSiteOf`/`hasSite` implementation, but keep the rest of the changes to the SHACL generation (anything agnostic to the `isSiteOf`/`hasSite` implementation)? Let me know if you're too busy  and I can make those changes

gtfierro (approved):

PR #186, opened 2020-09-25
Fix decoding errors (for Windows 10)

shreyasnagare:
Building BRICK on **Windows 10** gives me the following error:
```
Traceback (most recent call last):
  File "generate_brick.py", line 472, in <module>
    add_definitions()
  File "generate_brick.py", line 313, in add_definitions
    for definition in dictionary:
  File "c:\users\shrey\anaconda3\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 6449: character maps to <undefined>
```
Specifying the encoding (for definitions.csv) seems to fix this.

### DIFF
--- a/generate_brick.py
+++ b/generate_brick.py
@@ -303,7 +303,7 @@ def add_definitions():
     adds it to the graph. If available, adds the source information of
     through RDFS.seeAlso.
     """
-    with open("./bricksrc/definitions.csv") as dictionary_file:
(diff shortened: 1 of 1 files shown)


gtfierro:
Thanks for fixing this!

PR #196, opened 2020-11-13
Update namespaces.py

GeorgFerdinandSchneider:
Dear BRICK team,

please find this proposal for fixing schema.org namespace in Brick generation script.

Best

Georg


gtfierro:
Thanks for the fix! Sorry I missed this PR earlier

PR #200, opened 2020-12-10
disable nightly build on forks?

epaulson:
I updated my fork, and the nightly builds are running (and failing, but that's separate)

I'm not sure most folks want their forks to run a nightly build? I'm not heart-set on turning this off by default, but it does seem like a lot of compute that most folks don't want. 

( instructions from here: https://github.community/t/do-not-run-cron-workflows-in-forks/17636 ) - not sure I got it right though

### DIFF
--- a/.github/workflows/nightly-builds.yml
+++ b/.github/workflows/nightly-builds.yml
@@ -7,6 +7,7 @@ jobs:
   build:
     name: Nightly Build
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


gtfierro:
Disabling nightly builds on forks sounds good to me! Just as long as we document the change somewhere in the `CONTRIBUTING.md` document

gtfierro:
@epaulson if you merge master into this branch and the tests start passing, I can merge this soon as well

epaulson:
Let's not worry about this one

PR #202, opened 2020-12-15
update deps to build with new pip

### LINKED ISSUE #201: nightly builds are timing out
The nightly build started timing out about 10 days ago.

Pip released a new version about 10 days ago with a new resolver:
http://pyfound.blogspot.com/2020/11/pip-20-3-new-resolver.html

Perhaps related?

epaulson:
Should fix #201. Note that I basically just took an existing version of Brick I had and checked to see which versions of black/flake8/etc it had installed and said "use these"

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,16 +1,15 @@
 isodate==0.6.0
-owlrl>=5.2.0
-git+https://github.com/RDFLib/OWL-RL.git@471d1dfe8f6 # because of issue https://github.com/RDFLib/OWL-RL/issues/29, use git directly until it's released.
+owlrl>=5.2.1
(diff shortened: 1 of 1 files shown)


gtfierro:
Thanks for the PR!

We should probably give this a try with `owlrl >= 6.0.0`, but I believe that dependency probably comes from `brickschema`, which hasn't updated its dependencies yet. I can make that change at a later date after I test the newer OWLRL package. Is this PR good to merge?

epaulson:
I think it's good to merge -  the tests passed and the Brick.ttl that came out of this is the same as the last good nightly version, except it's got the schema.org fix applied, as expected.

PR #203, opened 2020-12-17
Entity Properties proposal

gtfierro:
Current [README](https://github.com/BrickSchema/Brick/blob/entity-properties/entity_properties/README.md) reproduced here:

# Entity Properties

Entity Properties are attributes or characteristics of Brick entities. Entity Properties should change rarely, if ever. Entity Properties are identified by an RDF property name and related SHACL shapes.

Entity Properties are categorized broadly by the following dimensions:
- **Inherent** vs **Configurable** (mutually exclusive):
    - an Inherent Property is one that is a permanent and/or characteristic property of the entity. The dimensions of a room, max rated power of a motor, tonnage of an RTU are all examples of Inherent Properties.
    - Configurable Properties are temporary or actuatable properties of an entity
- **Physical** vs **Virtual** vs **Identity** (not mutually exclusive):
    - a Physical Property is a characteristic that exists in the physical world ("meatspace")
    - a Virtual Property is a characteristic that has a digital representation. May be a property of the virtual representation of an entity, or a computed property of the entity
    - an Identity Property is a characteristic that identifies the device

The properties themselves are *instances* of Entity Properties; a shallow hierarchy of properties is defined with `rdfs:subPropertyOf`.

**Example**: here are some Physical Properties defined in the ontology which define the following hierarchy:
- `prop:hasArea`:
    - `prop:hasGrossArea`
    - `prop:hasNetArea`
- `prop:hasVolume`
- `prop:hasHeight`
- `prop:hasWidth`
- `prop:hasLength`

```ttl
prop:hasArea    a  prop:PhysicalProperty ;
    skos:definition "Entity has 2-dimensional area" ;
    rdfs:range  prop:AreaShape .

prop:hasGrossArea    rdfs:subPropertyOf  prop:hasArea ;
    skos:definition "Entity has gross 2-dimensional area" ;
    rdfs:range  prop:AreaShape .

prop:hasNetArea    rdfs:subPropertyOf  prop:hasArea ;
    skos:definition "Entity has gross 2-dimensional area" ;
    rdfs:range  prop:AreaShape .

prop:hasVolume  a  prop:PhysicalProperty ;
    skos:definition "Entity has 3-dimensional volume" ;
    rdfs:range  prop:VolumeShape .
```

The subject of each property will be a Brick entity; the value is an anonymous class which is goverened by a SHACL Shape. Note that the use of SHACL is optional here. We don't need SHACL inference for anything; we can use the SHACL shapes for "type checking" on the properties. That kind of SHACL validation is fairly easy to find implementations of.

```ttl
prop:AreaShape a sh:NodeShape ;
    rdfs:range qudt:Quantity ;
    sh:property [
        sh:path brick:hasUnit ;
        sh:in   (unit:FT2 unit:M2) ;
        sh:minCount 1 ;
    ] .


prop:VolumeShape    a   sh:NodeShape ;
    rdfs:range qudt:Quantity ;
    sh:property [
        sh:path brick:hasUnit ;
        sh:in   (unit:FT3 unit:M3) ;
        sh:minCount 1 ;
    ] .
```

## Aggregations

Description coming soon...

### Example 1: Room and Floor Geometry

Here's an example of how the above properties would be used to describe the net area of a floor, and the height and area of a room. Benefits of this approach:
- easy to add other metadata to the value itself (precision, timestamp, etc)
- semantics are in the property name -- this will simplify queries (see below)
- compatible with (and based on) QUDT
- (*if we need it*) multiple Brick entities can refer to the same property value but with different relationships

```ttl
:floor1 a   brick:Floor ;
    prop:hasNetArea    [
        prop:value  "100"^^xsd:integer ;
        brick:hasUnit   unit:FT2 ;
    ] .


:room1  a   brick:Room ;
    prop:hasArea    [
        prop:value  "35"^^xsd:integer ;
        brick:hasUnit   unit:FT2 ;
    ] ;
    prop:hasHeight [
        prop:value  "7"^^xsd:integer ;
        brick:hasUnit   unit:F ;
    ] .

```

Here's a sample SPARQL query to get the room names and areas for all rooms on the 4th floor. I think this is a fair "trade" syntactically for the additional expressive power we get out of the feature. I'm having a hard time figuring out what else we can remove from the model to make it simpler, without giving up the ability to add things like units.

```sparql
SELECT ?room ?area ?unit WHERE {
    ?room a brick:Room .
    ?room brick:isPartOf bldg:Floor_4 .
    ?room brick:hasArea/prop:value  ?area .
    ?room brick:hasArea/prop:unit   ?unit
}
```

### Example 2: Electrical Metering Points

An example of an single-phase active power sensor that measures net power, and a daily energy usage meter

```ttl
:meter1 a   brick:Power_Sensor ;
    brick:hasUnit   unit:KiloW ;
    prop:hasComplexity   [
        prop:value  "active" ;
    ] ;
    prop:hasPowerFlow   [
        prop:value  "net" ;
    ] ;
    prop:hasPhaseCount   [
        prop:value  "1"^^xsd:integer ;
    ] ;
    prop:hasIP6Address  [
        prop:value  "fe80::1",
    ] .

:daily_energy_usage_meter   a   brick:Energy_Sensor ;
    brick:hasUnit   unit:KiloW-HR ;
    prop:aggregate [
        prop:aggregationFunction    "sum" ;
        prop:aggregationInterval    "RP1D" ;
    ] ;
    prop:hasComplexity [
        prop:value  "active" ;
    ] ;
    prop:hasPowerFlow [
        prop:value  "net" ;
    ]
.

:hourly_peak_real_power_meter   a   brick:Power_Sensor ;
    brick:hasUnit   unit:KiloW ;
    prop:aggregate [
        prop:aggregationFunction    "max" ;
        prop:aggregationInterval    "RPT1H" ;
    ] ;
    prop:hasComplexity [
        prop:value  "real" ;
    ]
.
```


gtfierro:
@epaulson @jbkoh @danhugo @shreyasnagare please take a look and let me know what you think!

entity_properties/README.md
```diff
@@ -0,0 +1,168 @@
...
+        sh:in   (unit:FT2 unit:M2) ;
+        sh:minCount 1 ;
+    ] .
+
+
+prop:VolumeShape    a   sh:NodeShape ;
+    rdfs:range qudt:Quantity ;
+    sh:property [
+        sh:path brick:hasUnit ;
+        sh:in   (unit:FT unit:M) ;
```
shreyasnagare, line 71:
```suggestion
        sh:in   (unit:FT3 unit:M3) ;
```

entity_properties/defns.ttl
```diff
@@ -0,0 +1,227 @@
...
+
+prop:hasVolume  a  prop:PhysicalProperty ;
+    skos:definition "Entity has 3-dimensional volume" ;
+    rdfs:range  prop:VolumeShape .
+
+prop:VolumeShape    a   sh:NodeShape ;
+    rdfs:range qudt:Quantity ;
+    sh:property [
+        sh:path brick:hasUnit ;
+        sh:in   (unit:FT unit:M) ;
```
shreyasnagare, line 86:
```suggestion
        sh:in   (unit:FT3 unit:M3) ;
```

shreyasnagare:
@gtfierro This looks great.

entity_properties/README.md
```diff
@@ -0,0 +1,168 @@
...
+prop:hasArea    a  prop:PhysicalProperty ;
+    skos:definition "Entity has 2-dimensional area" ;
+    rdfs:range  prop:AreaShape .
+
+prop:hasGrossArea    rdfs:subPropertyOf  prop:hasArea ;
+    skos:definition "Entity has gross 2-dimensional area" ;
+    rdfs:range  prop:AreaShape .
+
+prop:hasNetArea    rdfs:subPropertyOf  prop:hasArea ;
+    skos:definition "Entity has gross 2-dimensional area" ;
```
corymosiman12, line 35: "Entity has net ..." ?

entity_properties/defns.ttl
```diff
@@ -0,0 +1,227 @@
...
+prop:AggregationShape   a   sh:NodeShape ;
+    sh:property [
+        sh:path prop:aggregationFunction ;
+        sh:in   ("max", "min", "mean", "mode", "median", "count", "sum") ;
+        sh:minCount 1 ;
+    ] ;
+    sh:property [
+        # TODO: how to represent durations?
+        sh:path prop:aggregationInterval ;
+        sh:datatype xsd:string ; # e.g. "15m" ?
```
corymosiman12, line 225: Hi Gabe here is what we do in BSync: https://github.com/BuildingSync/schema/blob/4db7102a6c627bade6e5dbb1861ac96633adfb9b/BuildingSync.xsd#L17805

corymosiman12:
Hi Gabe - I think overall this looks good.  Was just doing some quick checking against other things I know about.  I'm assuming you're not planning on getting too deep into geometrical representations, but figured I'd just throw some thoughts out:
1. Defining a L/W/H or even just a L/W for areas pushes a rectangular / box view (which is fine, maybe just want to make note of why this was chosen).  However, just defining Area & Volume keeps you safe from imposing geometrical constraints and getting into all of the things geojson / similar have to deal with.
2. If the intention is to allow some geometrical representations, gbxml basically takes the approach:
    a. Surfaces exist between 3D things (Spaces)
    b. Surfaces can either define RectangularGeometry or PlanarGeometry, which is a 'list of points defining a loop'.
3. Can you provide more info on what is / isn't an IdentityProperty?  In BSync, we have [PremisesIdentifier](https://github.com/BuildingSync/schema/blob/4db7102a6c627bade6e5dbb1861ac96633adfb9b/BuildingSync.xsd#L15215), so just trying to think how those would line up.  I know PNNL has also done a bunch of work on the [UBID](https://ubid.pnnl.gov/) in case that is of interest.  You could talk to Mark Borkum
4. BEDES has [gross](https://bedes.lbl.gov/bedes-online/gross) and [net](https://bedes.lbl.gov/bedes-online/net) if you are interested in using those

entity_properties/README.md
gtfierro: Thanks! Forgot to fix in the copy paste

gtfierro:
@corymosiman12 thanks for the perspective on geometry --- I don't want to stray too far into into supplanting gbXML/IFC geometry, so it is probably best if we stick to Area and Volume for now (I've removed them from the implementation).

Regarding Identity Property: my intention is to provide a category for any property that could be used to name an entity: IP address, MAC address, serial number, UUID, ID from an asset management system, NMI (for meters in Australia), etc. Instances of `prop:IdentityProperty` could be added to support BSync PremisesIdentifier, PNNL UBID, or any other established or emerging standard. You can use the additional `Inherent` vs `Configurable` categorization to indicate if the identifier is expected to change at a later date or not. Do you think this is a sufficient categorization?

gtfierro:
@shreyasnagare definitely! Would you able to draft up (either as a list of concepts or as the TTL itself) some of those HVAC-related properties so we can see how well they fit?

JoelBender:
1) I agree that it would be better to stay away from gbXML stuff, [BOT](https://w3c-lbd-cg.github.io/bot/) has a [simple3Dmodel](https://w3c-lbd-cg.github.io/bot/#hasSimple3DModel) which punts to [Web3D](https://www.web3d.org/).

2) I would expect that any domain that has a notion of identity would also have their own IRI for what something means to have that identity, and it might not be a literal.  I would stay away from it.

3) ISO-8601 duration encoding is a good idea, is [OWL-Time](https://www.w3.org/TR/owl-time/) too much?

gtfierro:
The Entity Properties proposal above was discussed on the weekly Brick development call, and the consensus was to move forward with the current design for the following reasons:
- this design has no evident major flaws and covers known use cases
- the design aligns with design choices made in the development of the ASHRAE 223P standard

Current tasks are:
- [ ] develop a hierarchy of property names reflecting common use cases 
- [ ] develop inference "shortcuts" where existing/future Brick classes can be inferred from or defined by entity properties

corymosiman12:
Hey Gabe - I had suggested a `hasStage` Datatype property. Based on the above, I believe it would be a configurable property. It is intended to mimic the Haystack stage: https://project-haystack.dev/doc/lib-phIoT/stage. Thoughts?

gtfierro:
That's a good idea! I'll make sure to include. I want to check my understanding that this captures the *number* of operating stages that a particular equipment has, and not *which* of those stages it is in at a particular moment, right? To track which stage an equipment is in at a given time, that should be a timeseries of type `brick:Stage_Status`

corymosiman12:
I think you have multiple `Stage_Enable_Command` points (hasStage: 1, hasStage: 2, etc.) for a specific equipment (I know this mainly for motors).  And then you would write a boolean value to whichever 'stage' you want it to be in. I'm not sure whether a `Stage_Status` really exists, but I can ask Selam tomorrow.

gtfierro:
We got `Stage_Status` from somewhere, so it may just be rare? Good to know (and totally make sense) that `Stage_Enable_Command` would need the stage associated with it. Should be easily done with the Entity Properties

ashatdal:
Late to the discussion here but just wanted to add our experience. Part of these may already be covered from other sources, but here are some of the basic asset properties to carry with Equipment class.
"assetId": {
    "description": "asset id",
    "type": "string"
},  
"description": {
    "description": "A free-form description of the entity to be used at the users discretion",
    "type": "string"
},  
"serialNumber": {
    "description": "device serial no",
    "type": "string"
},  
"firmwareVersion": {
    "description": "firmwareVersion",
    "type": "string"
},  
"geoLocation": {
    "$ref": "#/definitions/geoJSON",
    "description": "geoLocation"
},  
"installationDate": {
    "description": "installation date",
    "format": "date-time",
    "type": "string"
},  
"ipAddress": {
    "description": "ipAddress",
    "type": "string"
},
"macAddress": {
    "description": "macAddress",
    "type": "string"
},
"manufacturer": {
    "description": "manufacturer or make",
    "type": "string"
},
"modelName": {
    "description": "model name of the device/equipment",
    "type": "string"
},
"streamURL": {
    "description": "URL for the streaming device, if applicable",
    "type": "string"
},
"warrantyDate": {
    "description": "warranty End Date",
    "format": "date-time",
    "type": "string"
}

GeoLocation information is useful for both equipment and location (and sensor point classes). Most geolocation formats are interchangeable (geoJSON is pretty well defined and so is WKT format).

PR #211, opened 2021-01-20
Minor Ontology Fixes + New Classes

gtfierro:
Addresses some different tag issues, including removing the `owl:AllDifferent` classification. Adding classes requested in #205, #199 and from prior (offline) conversations

PR #212, opened 2021-01-20
Use SHACL-AF to do Brick Tag to Class inference

gtfierro:
As I am noticing in #211, there are some issues with the current implementation when two classes which are *not* related through a `rdfs:subClassOf` relationship have sets of tags where one is fully contained within the other. So far, this hasn't been an issue for Brick. However, in the process of updating the tooling in #211, I noticed the following situation.

- `brick:Motor_Control_Center` has tags Motor, Control, Center and Equipment
- `brick:Motor` has tags Motor and Equipment
- `brick:Motor` is **not** a subclass or superclass of `brick:Motor_Control_Center`

As a result, any entities with the `Motor` and `Equipment` tags will be inferred as a `brick:Motor` --- including instances of `brick:Motor_Control_Center`! This is obviously problematic. Thankfully, SHACL-AF offers a path forward. It is recommended to read up on [SHACL-AF rules](https://www.w3.org/TR/shacl-af/#TripleRule) and be familiar with regular [SHACL](https://www.w3.org/TR/shacl/).

Turns out, we can define the semantics we want using SHACL shapes, which can express "negation-like" properties that OWL cannot.

```ttl
brick:Motor_Control_Center_Tag a owl:Class, sh:NodeShape ;
    sh:targetSubjectsOf brick:hasTag ;
    sh:rule [
		a sh:TripleRule ;
		sh:subject sh:this ;
		sh:predicate rdf:type ;
		sh:object brick:Motor_Control_Center ;
        sh:condition [
            sh:property [
                sh:path brick:hasTag ;
                sh:qualifiedValueShape [ sh:hasValue tag:Motor ] ;
                sh:qualifiedMinCount 1 ;
            ] ;
        ] ;
        sh:condition [
            sh:property [
                sh:path brick:hasTag ;
                sh:qualifiedValueShape [ sh:hasValue tag:Center ] ;
                sh:qualifiedMinCount 1 ;
            ] ;
        ] ;
        sh:condition [
            sh:property [
                sh:path brick:hasTag ;
                sh:qualifiedValueShape [ sh:hasValue tag:Control ] ;
                sh:qualifiedMinCount 1 ;
            ] ;
        ] ;
        sh:condition [
            sh:property [
                sh:path brick:hasTag ;
                sh:qualifiedValueShape [ sh:hasValue tag:Equipment ] ;
                sh:qualifiedMinCount 1 ;
            ] ;
        ] ;
        sh:condition [
            sh:property [
                sh:path brick:hasTag ;
                sh:maxCount 4;
                sh:minCount 4;
            ] ;
        ] ;
	] ;

.

brick:Motor_Tag a owl:Class, sh:NodeShape ;
    sh:targetSubjectsOf brick:hasTag ;
    sh:rule [
		a sh:TripleRule ;
		sh:subject sh:this ;
		sh:predicate rdf:type ;
		sh:object brick:Motor ;
        sh:condition [
            sh:property [
                sh:path brick:hasTag ;
                sh:qualifiedValueShape [ sh:hasValue tag:Motor ] ;
                sh:qualifiedMinCount 1 ;
            ] ;
        ] ;
        sh:condition [
            sh:property [
                sh:path brick:hasTag ;
                sh:qualifiedValueShape [ sh:hasValue tag:Equipment ] ;
                sh:qualifiedMinCount 1 ;
            ] ;
        ] ;
        sh:condition [
            sh:property [
                sh:path brick:hasTag ;
                sh:minCount 2;
                sh:maxCount 2;
            ] ;
        ] ;
	] ;
.

:mcc brick:hasTag tag:Motor, tag:Center, tag:Control, tag:Equipment ;
.
:neither brick:hasTag tag:Motor, tag:Control, tag:Equipment ;
.
:motor brick:hasTag tag:Motor, tag:Equipment ;
.
```

Applying SHACL-AF turns `:mcc` into a `brick:Motor_Control_Center`, `:motor` into a `brick:Motor`, and `:neither` into neither of those (it gets no class). This works by ensuring for each Brick entity that is the subject of the `brick:hasTag` relationship (read: all Brick entities with tags) that *if the following is true*:

- the entity has 1 instance of each of the required tags associated with the entity via `brick:hasTag`
- the entity has *at most* as many brick:hasTag` properties as unique tags associated with the class

*then* a triple is added to the graph which instantiates the entity as a member of the given Brick class.

~~This implementation has no performance penalty~~It is too early to tell if this has a performance penalty or not. The large number of SHACL shapes may slow down inference a bit, but it is unclear how much this will be offset by the simplified OWL-RL axioms in the ontology.

I recommend merging this PR for Brick v1.2 in order to fix the [hack](https://github.com/BrickSchema/Brick/pull/211/commits/d5da32a8eec165dd84f46cc2c588e140ec06bfbf) that is currently in #211


gtfierro:
After some investigation and a few tests, it looks like this implementation is *correct* in the sense that it properly captures the intended semantics. However, pySHACL's performance on these shapes seems to suffer --- the automated tests are timing out. Obviously, this is not ideal. I'll be exploring some alternative implementations with equivalent semantics that may perform better.

gtfierro:
**TODO**:

- [ ] move this into an "extension" so it can be optionally loaded

PR #214, opened 2021-01-25
v1.2 release

gtfierro:
This should be the target branch for v1.2 PRs

Remaining:
- [X] #203 Entity Properties
- [x] Entity Property documentation  (documentation [here](https://docs.brickschema.org/metadata/entity-properties.html))
- [x] #212 SHACL-AF shapes for tags
- [x] #216 VBIS fixes
- [x] Timeseries blank node model (#221)
- [X] Timeseries model documentation ([here](https://docs.brickschema.org/metadata/timeseries-storage.html))
- [x] #215 Cooling Towers and Condenser Plants
- [x] remove PROP namespace and replace with Brick
- [x] remove RDFS.domain of Brick.hasUnit
- [x] Location model
- [x] remove the `has` prefix for entity properties
- [x] enforce line endings with .gitattributes


corymosiman12:
Hi Gabe - I will look to add those new concepts and make a PR into here either today or tomorrow.

corymosiman12:
Hey Gabe - can you describe how, with the new IRI formation, one is supposed to convey the version of Brick used in constructing a model?

gtfierro:
As long as the major version number of Brick remains 1, which it will for some time, then Brick versions will be backwards compatible and you should use the latest version available. One can query the version number of the Brick graph loaded in through the `dcterms:version` property. One will also be able to load in a specific version of Brick using a versioned-URI (e.g. https://brickschema.org/schema/1.2/Brick), but the downloaded version of the ontology will have a versionless namespace. The hope is that this will facilitate the adoption of future Brick versions, and largely remove the need to worry about which specific version of Brick was used to define a model

gtfierro:
@epaulson @jbkoh @shreyasnagare can you give this a look over in the next few days so that I can merge soon? Thanks!

alignments/rec/Brick-REC-nodevice.rdf
```diff
@@ -6,7 +6,7 @@
      xmlns:xml="http://www.w3.org/XML/1998/namespace"
```
jbkoh, line 6: Is there a specific reason that REC alignment should be in rdf-xml instead of ttl?

jbkoh: I'd prefer using a single format across all the ontologies within the repo when it's possible.

gtfierro: That's the form of the file that we received from the REC team. I just left it in that form so that it would be easier to incorporate future changes/alignments if/when those occur

bricksrc/alarm.py
```diff
@@ -252,6 +252,21 @@
                     "Water_Loss_Alarm": {
                         "tags": [TAG.Point, TAG.Loss, TAG.Water, TAG.Alarm],
                     },
+                    "Water_Level_Alarm": {
+                        "tags": [TAG.Water, TAG.Level, TAG.Alarm, TAG.Point],
+                        "subclasses": {
+                            "Collection_Basin_Water_Level_Alarm": {
```
jbkoh, line 258: Maybe `High` instead of `Collection` because 1) `High` is more straightforward 2) `High` is more commonly used for the same meaning in other classes. If `Collection` contains a special context here, I'm fine with the current this.

gtfierro: How is `High` a substitute for `Collection`? `Collection Basin` is the entity containing the water. @corymosiman12 could you help elaborate?

bricksrc/entity_properties.py
```diff
@@ -0,0 +1,244 @@
+from .namespaces import BRICK, TAG, OWL, RDFS, SKOS, UNIT, XSD
+from rdflib import Namespace, Literal
+
+# these are the "relationship"/predicates/OWL properties that
+# relate a Brick entity to a structured value.
+# These are all instances of Brick.EntityProperty, which is
+# a subclass of OWL.ObjectProperty
+entity_properties = {
+    BRICK.hasArea: {
```
jbkoh, line 9: I know we briefly discussed this before without a deeper discussion, but now it concerns me a bit, so I think better late than never: 

I feel `has` is redundant to all the entity property names here. When it comes to a noun as a property name, it's quite obvious that all the objects would be a type of property for the subject. Not having `has` is a common practice across many data models. E.g., [Microsoft graph API](https://docs.microsoft.com/en-us/graph/api/resources/contact?view=graph-rest-1.0#properties) and [Haystack](https://project-haystack.org/tag/area).

(@corymosiman12 raised this concern too, let me know if you'd like to add anything here)

bricksrc/equipment.py
```diff
@@ -4,8 +4,8 @@
 Set up subclasses of the equipment superclass
 """
 equipment_subclasses = {
-    "HVAC": {"tags": [TAG.HVAC]},
-    "Weather_Station": {"tags": [TAG.Weather, TAG.Station]},
+    "HVAC": {"tags": [TAG.HVAC, TAG.Equipment]},
```
jbkoh, line 7: We now have System category. Do we want to add `HVAC_System` there in the near future? In that case, I would like to make it more explicit as `HVAC_Equipment`. (It's a breaking change, so maybe we need to collect it for v2.)

bricksrc/alarm.py
jbkoh: Closing this one as I realized taht `Collection_Basin` is a type of Basin.

bricksrc/system.py
```diff
@@ -59,6 +59,9 @@
                             },
                         },
                     },
+                    "Condenser_Water_System": {
+                        "tags": [TAG.Condenser, TAG.Water, TAG.System]
+                    },
                 },
             },
```
jbkoh, line 66: I feel we have to add `HVAC_System` and change the existing `HVAC` to `HVAC_Equipment` as mentioned in another comment.

bricksrc/timeseries.py
```diff
@@ -0,0 +1,57 @@
+"""Defines the timeseries reference model"""
+
+from .namespaces import BRICK, A, OWL, SKOS, SH, XSD
+from rdflib import Literal, BNode
+
+
+def define_timeseries_model(G):
```
jbkoh, line 7: Why do we need a script to generate this instead of posting a shacl file directly? This script is just a python version of a shacl file, which I don't think is useful to maintain but complicates maintenance. Still, I'm fine if the purpose is to make it consistent with other generation processes.

jbkoh (changes requested):
I finished a first pass with several changes. Overall, it looks great: such nice progress!

alignments/rec/Brick-REC-nodevice.rdf
jbkoh: But it's weird that we have multiple alignment files for the same ontology in different formats (https://github.com/BrickSchema/Brick/blob/efbbff7353cddae6f9c180d907d2bd02e8f084e2/alignments/rec/Brick-REC-alignment.ttl)

And they don't keep rdf-xml in alignments anyway: https://github.com/RealEstateCore/rec/tree/master/ontology/alignments

I'm fine either way, and if you still feel the rdf-xml is better, please resolve this conversation.

bricksrc/equipment.py
gtfierro: We don't need to remove "HVAC", but we should definitely add "HVAC Equipment" and clarify it as different from "HVAC System"

bricksrc/timeseries.py
gtfierro: The timeseries model may get more complex in the future, so I made sure to leave us some code that we can extend later

gtfierro:
Hi @jbulow! That file was supplied by external collaborators, so it is just an accident that it has CRLF. Thanks for the pointer for `.gitattributes`

bricksrc/alarm.py
jbkoh: I closed this one as it was my confusion.

bricksrc/equipment.py
jbkoh: It's quite confusing to have all of HVAC, HVAC_Equipment, and HVAC_System.

gtfierro: Should we just move directly to HVAC Equip + HVAC System and remove HVAC in this release?

corymosiman12: I think you can get away without HVAC and without HVAC_Equipment and just have everything under Equipment. HVAC_System still makes sense to me

epaulson:
Looks good to me

epaulson (approved):

shreyasnagare (approved):
This looks good. 

I was trying to build this on a fresh instance of Kali Linux earlier today and it looks like you can't generate `Brick.ttl` without Docker. I never realized this before because my environment already had Docker installed. If that is the case, we might want to add that information in the [Development Environment](https://github.com/BrickSchema/Brick/blob/v1.2-release/CONTRIBUTING.md#development-environment) section.

gtfierro:
@shreyasnagare Docker *shouldn't* be necessary (even though it is listed in the requirements: https://github.com/BrickSchema/Brick/blob/master/requirements.txt#L10) --- it just enables the use of Allegrograph which is faster than the default reasoner. Can you try uninstalling the Python docker package to see if Brick.ttl generation is possible without it?

Thanks for going through the PR by the way!

gtfierro:
@jbkoh can you approve the PR if there are no other changes you would like to be made?

shreyasnagare:
> @shreyasnagare Docker _shouldn't_ be necessary (even though it is listed in the requirements: https://github.com/BrickSchema/Brick/blob/master/requirements.txt#L10) --- it just enables the use of Allegrograph which is faster than the default reasoner. Can you try uninstalling the Python docker package to see if Brick.ttl generation is possible without it?
> 
> Thanks for going through the PR by the way!

Yes, it seems to work after I uninstalled the docker python package.

gtfierro:
Good to know! I put a note in the CONTRIBUTING.md document that docker is optional, but I kept it in there as a default install because it should work on most systems. If this ends up not being the case, we can revisit in the future

shreyasnagare:
> Good to know! I put a note in the CONTRIBUTING.md document that docker is optional, but I kept it in there as a default install because it should work on most systems. If this ends up not being the case, we can revisit in the future

I tried this on Windows and Ubuntu as well and ran into the same issue. Everything works fine when I have docker installed. I think this is something we can fix on the py-brickschema side of things: doing everything without docker if `docker.from_env(version="auto")` throws an error.

gtfierro:
@shreyasnagare can you try installing `brickschema==0.3.0a9` to see if it gracefully degrades without being able to connect to the docker client?

jbkoh (approved):
LGTM!

shreyasnagare:
> brickschema==0.3.0a9

```
2021-02-19:22:08:23,290 ERROR   [inference.py:94] Could not connect to docker (Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))); defaulting to naive evaluation
Traceback (most recent call last):
  File "generate_brick.py", line 41, in <module>
    from bricksrc.quantities import quantity_definitions, get_units
  File "/root/Brick/bricksrc/quantities.py", line 11, in <module>
    g.expand(profile="brick")
  File "/usr/local/lib/python3.6/dist-packages/brickschema/graph.py", line 198, in expand
    return self.expand("owlrl+shacl+owlrl", backend=backend)
  File "/usr/local/lib/python3.6/dist-packages/brickschema/graph.py", line 194, in expand
    self.expand(prf, backend=backend)
  File "/usr/local/lib/python3.6/dist-packages/brickschema/graph.py", line 216, in expand
    self._inferbackend = OWLRLAllegroInferenceSession()
TypeError: __init__() should return None, not 'OWLRLNaiveInferenceSession'
```

Issue #197, opened 2020-11-23
Chiller Plant Modeling


gtfierro:
This issue will track progress/discussion on augmenting the set of chiller plant-related concepts defined by Brick

Required Points:
- [ ] Chilled Water Entering Temperature Sensor (w.r.t. evaporator?)
- [ ] Chilled Water Leaving Temperature Sensor (w.r.t. evaporator?)
- [ ] Chilled Water Flow Sensor (w.r.t. evaporator?)
- [ ] Entering Water Temperature Sensor (w.r.t. condenser)
- [ ] Water Temperature Sensor (inside condenser?)
- [ ] Water Flow Sensor (inside condenser?)
- [ ] chilled water supply temperature (w.r.t. ahu; incoming to AHU)
- [ ] chilled water return temperature (w.r.t. ahu; outgoing from AHU)

david-waterworth:
I'll add to these as I identify but I'd categorize into systems/subsystems and then equipment and then identify the points. The main thing to work out is how to represents the flow in/out of each equipment (a chiller or cooling tower) and the aggregate flow (i.e. there may be three cooling towers each of which have inlet/outlet water temperature, the water from each is combined and return to the condenser side of the chillers and then split again).

CH - a chiller (located in a CHWS)

Chilled Water Entering Temperature Sensor   - the temperature of the water entering a chiller (evaporator)
Chilled Water Leaving Temperature Sensor   - the temperature of the water leaving a chiller (evaporator)
Chilled Water Flow Sensor   - the flow rate of the water through the chiller (evaporator)
Condenser Water Entering Temperature Sensor   - the temperature of the water entering a chiller (condenser)
Condenser Water Leaving Temperature Sensor   - the temperature of the water leaving a chiller (condenser)
Condenser Water Flow Sensor   - the flowrate of the water through the chiller (condenser)

CT - a cooling tower (located in a CHWS)

Condenser Water Entering Temperature Sensor   - the temperature of the water entering a cooling tower
Condenser Water Leaving Temperature Sensor   - the temperature of the water leaving a cooling tower
Condenser Water Flow Sensor   - the flow rate of the water through the cooling tower

CHWS - a chilled water system, consists of chillers, cooling towers, primary and secondary pumps, bypass valves, neutral flow lines.

Chilled Water *Leaving Temperature Sensor   - the temperature of the water being sent to the field from a chws
Chilled Water *Entering Temperature Sensor   - the temperature of the water being returned from the field from a chws
[...]

PR #215, opened 2021-01-27
Add concepts for Cooling Towers, Condenser Plants

corymosiman12:


corymosiman12:
@gtfierro I'm just getting one failed test regarding obsolete definitions.  This is what the `obsolete_definitions.json` file looks like:
```
[
  "https://brickschema.org/schema/1.2/Brick#Outside_Air_Wet_Bulb_Temperature"
]
```

??

gtfierro:
Obsolete definitions are when you have a definition for a class that doesn't exist. My guess is that the definition should be for `Outside_Air_Wet_Bulb_Temperature_Sensor` rather than `Outside_Air_Wet_Bulb_Temperature`?

gtfierro:
@corymosiman12 can you describe the intended use of the `Loop` class hierarchy? I think this might be a duplicate with our intended `System` class

corymosiman12:
hey @gtfierro the diagram [here](https://github.com/BrickSchema/Brick/issues/197#issuecomment-767821273) and just added a definition for a loop. How does that look?

gtfierro:
@corymosiman12 My question was more how the choice of modeling a Loop differs from the intended use of System. In Brick, we've already defined HVAC System, Chilled Water System, etc. The intended use here is models like the following:

```ttl
:bldg   a   brick:Building ;
    brick:hasPart   :hvac ;
.

:hvac  a brick:HVAC_System ;
    brick:hasPart  :ahu1, :vav1, :vav2, :vav3 ;
.
# and so on
```

Equipment can be grouped together into Chilled Water Systems, Hot Water Systems, etc ---- it is an optional model, but it is potentially something that offers organization to existing equipment (outside of the `feeds` relationship). Is your decision to model a Loop object complementary or redundant with the above design? Maybe the "chilled Water system" is *all* of the chilled water equipment/etc in the building (including multiple loops), but a particular loop is a subset of that system in which all the components are connected?

corymosiman12:
Hey Gabe - good questions. I'm gonna try to run this by a few people to get their thoughts. I know that people generally discuss loops when they are doing plant / system design because they are designing pumps to deal with pressure drops. I mostly hear talk of 'air distribution systems' in real world buildings as opposed to 'air loops' which comes from the energy modeling perspective and balancing mass flows / thermodynamics / etc.

> Is your decision to model a Loop object complementary or redundant with the above design?

Loops give you opportunity to talk about the 'in-between'. For example, the set of cooling towers in the linked graphic would often be referred to as the condenser water plant. The condenser water system would include the entire condenser water plant, the condenser loop, as well as the chilled water plant (the set of chillers) (i.e. there is overlap between the condenser water system and chilled water system). The condenser water loop describes the things connecting the condenser plant and chilled water plant, the 'in-between'. In that sense, it is just providing a name to something.

> a particular loop is a subset of that system in which all the components are connected?

This might be a good way to put it. Gets to the 'pressure drop' aspect of things. We should probably just get some people who are actually plant engineers for a living to review :). I'll see who I can drum up.

It might be easiest to remove the loop elements / tree for now and try to get the other stuff into 1.2, then come back with a better thought through proposal regarding the loops stuff. Let me know what you think of that - I can open a new ticket and remove the stuff tomorrow if makes sense.

gtfierro:
Thanks for the explanation! It is worth getting this right --- I agree with your assessment that we should defer the Loop model to 1.2.x or 1.3.

corymosiman12:
ok @gtfierro, loops removed, PR marked ready for review

gtfierro (changes requested):
These look great, @corymosiman12! The only request I have before merging these is if you could please add their definitions to `bricksrc/definitions.csv`? Then I should be able to merge this in

corymosiman12:
whoops - those got removed when I merged in the new Brick namespace. Should be good now.

gtfierro (approved):
Great work @corymosiman12! Thanks for contributing this

gtfierro:
One question before I merge: a Boiler a kind of Water Heater? We might want the `parents` annotation on the `Boiler` class so that it is organized under Water Heater as well

gtfierro:
Hit the "merge" button by accident!! Let me know what your take on the Boiler / Water Heater issue is and I can make the change in the 1.2 branch

corymosiman12:
all good. now that you mention that, I think you're right. My inclination though is that a `Water_Heater` isn't just HVAC, and should actually just be a subclass of `Equipment`. It can obviously be used in an hvac context, but water heaters / boilers can also be used for domestic hot water / snowmelt / etc. etc. Thoughts?

gtfierro:
Good point! I'll implement the following:

```
Water_Heater subclassof Equip
Boiler subclassof HVAC, Water_Heater
```

PR #217, opened 2021-02-04
Switch BRICk.hasUnit to PROP.unit for entity properties

gtfierro:
`BRICK.hasUnit` has a domain of `BRICK.Point`. `PROP.unit` is instead a subproperty of `QUDT.unit` which, matches the semantics more clearly. An alternative here is to change the definition of `BRICK.hasUnit` so that the domain is unspecified.


gtfierro:
Unneeded given the changes in #225

PR #219, opened 2021-02-05
add several entity properties from haystack

jbkoh:
Adding
- hasYearBuilt
- hasCoolingCapacity
- hasBuildingPrimaryFunction
adopted from Project Haystack


gtfierro (approved):
Looks great! Thanks for adding

Issue #37, opened 2017-11-13
Utilising Building Topology Ontology


blip2:
Building Topology Ontology / bot https://github.com/w3c-lbd-cg/bot provides a central lightweight ontology for space and physical relations within buildings with the aim to align with or provide a base level of classes for other domains.

Given that Brick is a domain specific ontology (controls), I think Brick should either aim to align with BOT (there is an existing alignment here: https://github.com/w3c-lbd-cg/bot/pull/10/commits/bfee1e4ee97d0c5cb35fe062a30428007663b8c2#diff-73c0fa08510685b98f9b9cfe1ed24cfc) or, preferably re-jig classes like Location, Zone, etc to directly use bot, minimising duplication.

Thoughts?

gtfierro:
Thanks for the suggestion, @blip2! Locations is definitely an area that we need to expand out a bit. Looks like BOT has `space`, `floor` and `building` as the base classes, with a couple space subclasses such as `garage`, which is a good start. We should have a little discussion about what we need to handle space-wise in Brick.

Some issues we should think about:
- is building -> floor -> room and zone -> room enough? Do we need other spatial composition idioms?
- need a basic listing of "types" of rooms. I think Brick now has laboratory, office, basement.
- do we want to capture connections between rooms? Just `adjacentTo`, or `hasWallTo`/`hasNoWallTo`/etc?
- do we want to capture orientation of the room? `myroom   hasOrientation  WestFacing`?

jbkoh:
Thanks for the pointer, Ben!

I looked a bit into BOT's literature. This must be its core design:
![image](https://user-images.githubusercontent.com/1572627/33061451-08e82ba8-ce51-11e7-961a-6402a584b6bd.png)

- Basically the only difference so far seems to be "adjacentZone" and "adjacentElement".  
  - While Element can be anything other than space, there is an example using it to describe Wall<->Room relationship. It is an interesting concept and probably scale well by putting more properties to Wall if needed. Other than that, I can't think of adjacentElement's usage.
  - Whether adjacentZone is needed or not should be discussed. I do not know when to stop putting more detailed location information but I am fine with adjacency. Unfortunately, I could not find their logistics on why adjacency is needed.
- Relying on an existing schema should be carefully considered. We do not know how much BOT will last as people worry about Brick with the same reason. As a first step, we can just provide an alignment as an extension. I saw the alignment they did, but we may be able to put it more comprehensively.
[...]

PR #220, opened 2021-02-05
Addition of BOTAlignment.ttl

GeorgFerdinandSchneider:
Addition of BOTAlignment as plain .ttl file

rdfs:seeAlso https://github.com/BrickSchema/Brick/issues/37#issuecomment-773805892


gtfierro (approved):
Looks great! Thanks for adding

PR #221, opened 2021-02-05
Defining timeseries storage model

gtfierro:
Adds the following:

- `brick:timeseries` property between a Brick Point and a new `TimeseriesReference`
- a `TimeseriesReference` (may be implicitly realized as a blank node) has two properties:
    - `brick:hasTimeseriesId`: a String; must have exactly 1
    - `brick:storedAt`: defined as a property; can have at most 1

Biggest thing right now is:
- [x] does this cover the intended use cases? (I believe it does, as per the discussions that took place)
- [x] how do we want to define the `storedAt` property? @jbkoh @epaulson 

Example:

```ttl
:xyz    a      brick:Temperature_Sensor ;
    brick:hasUnit   qudt:DEG_F ;
    brick:timeseries [
         brick:hasTimeseriesId "88698ade-9b77-40fa-bbb4-dd9d0adb1f16" ;
         brick:storedAt "postgresql://1.2.3.4:5432/mydata" ; # open for feedback here; how do we want to standardize this?
    ] ;
.
```


gtfierro:
@jbkoh @epaulson I've added some documentation on this feature [here](https://docs.brickschema.org/metadata/timeseries-storage.html). Let me know what you think!

PR #222, opened 2021-02-05
update units and refs for haystack-origin properties

jbkoh:


jbkoh:
something messed up. will create a new PR in a different branch.

PR #230, opened 2021-02-12
add floor number

corymosiman12:


corymosiman12:
hey @gtfierro do you think you will be able to get this into the v1.2 release?

gtfierro:
Possibly! Can you add a description of why these need to be modeled and an example of how they would be used in a model? Even if they are not included in the immediate v1.2 release, they can appear in subsequent v1.2.x releases

corymosiman12:
FloorNumber is a common entity property for floors (knowing where things are located). Having a standardized nomenclature for how to number floors (i.e. ground floor is 0) is also important.

```turtle
:floor-1 a brick:Floor ;
    brick:hasFloorNumber 1 .
```

corymosiman12:
@gtfierro descriptions and examples added. lmk if anything else

corymosiman12:
@gtfierro where should I add a new PR?

Issue #190, opened 2020-10-20
Location Modeling


gtfierro:
This has a few main parts:

1) ensure that the mapping with BOT is sensible and up to date with the Brick v1.1 locations
2) produce documentation outlining idiomatic use of Brick v1.1 location classes 
3) decide how to attach "use cases" to locations in the Brick model, e.g. classroom, laboratory, office, etc. There are some existing location use cases in some ASHRAE standards that we can use to bootstrap this

gtfierro:
This topic was discussed during the Brick dev meeting today, and currently we are moving forward with a design that uses OWL classes to denote different types of spaces (categorized by use). Multiuse rooms are modelled by having a Space instance have multiple Space types. We should hopefully have a prototype soon

epaulson:
We put together some possible class extensions for 1.2 that map to the top-level classes of Brick 1.1. It is fairly extensive but only augments what's already there in 1.1. (Well, we think we should bring back 'city' from 1.0.3.)

We can convert this into Python and open a pull request if that would be easier to review, or we can work on this Markdown version and then translate it to Python for an actual PR when we're satisfied with the hierarchy.

For now it's all single inheritance but I think it's all compatible with the plan we discussed of giving a space instance multiple space types to work with multi-use.  

* [**Location** (1.1)](https://brickschema.org/ontology/1.1/classes/Location) 
  * [**Building** (1.1)](https://brickschema.org/ontology/1.1/classes/Building) 
    * Parking_Structure (new) - 
  * [City (1.0.3)](https://brickschema.org/ontology/1.0.3/classes/City) 
  * [**Floor** (1.1)](https://brickschema.org/ontology/1.1/classes/Floor) 
    * [**Basement** (1.1)](https://brickschema.org/ontology/1.1/classes/Basement) 
    * [**Roof** (1.1)](https://brickschema.org/ontology/1.1/classes/Rooftop)
    * Parking_Level (new) - 
  * Location_Portfolio (new) -  - describes a collection of locations, especially Sites that could describe the set of sites belonging to an organization for example
  * Outdoor_Area (new) - 
    * Bench_Space (new) -  - specific to play areas in stadium, an area for participants and referees by the side of the field of play
    * Concourse (new) -  - large outdoor public area usually around a public building
    * Field_Of_Play (new) -  - e.g. a soccer pitch
    * Information_Area (new) -  - e.g. a information kiosk or booth where visitors would look for information
[...]

PR #231, opened 2021-02-14
add more location classes

epaulson:
Includes location classes, mostly as defined in the comment from #190 

I added a few new intermediate classes for media rooms, for service rooms, for storage rooms, and for teleco rooms. I slipped in a few new classes for telcos, and renamed 'kitchen' to be 'office kitchen' which is what I think we meant. I'll try to go back through the original issue and update the comment


gtfierro (changes requested):
Thanks @epaulson! Do you think that `Location_Portfolio` counts a location? I could see this being an external term for now --- we haven't yet taken a position on what administrative/organizational structures should make their way into Brick yet.

A separate issue is how to rectify subclasses of a Building with the Energy Star "building primary function" classification (https://github.com/BrickSchema/Brick/blob/v1.2-release/bricksrc/entity_properties.py#L122). I can work on a way of making them equivalent, which should solve the issue

epaulson:
For Location Portfolio, I did think that it usually has some kind of geographic connotation but might not follow a strict geopolitical boundary (and might have weird overlaps).  In that sense, it is a "location"

I think we need some kind of location "container" and 'portfolio' lines up with what some of the energy management lit uses, so that's where we got it from. 

If we want to think this one through a bit more we can take it out and revisit post 1.2

corymosiman12:
I like the concept of portfolio outside of just a location. I agree with @epaulson that a Portfolio may have a geographic connotation, but I think that a portfolio is generally just a grouping of buildings or properties based on some lens of interest. I don't know exactly what that makes it, though.

gtfierro:
I created #233 to capture the discussion about collection/groups of entities. I suppose that if an HVAC Zone is a location, then a Portfolio may also be a location as well, but I'd like to have a dedicated discussion about it so that we can be sure we are modeling similar concepts in similar ways. @epaulson would you be alright with removing `Location_Portfolio` for now so I can merge this and we can discuss how to include that in #233 ?

epaulson:
Removed Location Portfolio.

gtfierro (approved):
Thanks @epaulson!

PR #240, opened 2021-03-12
Class-based shapes

jbkoh:
This is the starting point of discussions for Class-based shapes. We have relied on OWL for defining property constraints, but due to its limited coverage, it was impossible to represent many things, which had users scratch their heads when it comes to using certain properties with certain Classes.

I propose to use SHACL NodeShapes to define constraints at entity levels. The current draft contains PointShape, LocationShape, and EquipmentShape. This is also promising in extensibility for modeling more specific shapes in the future. E.g., a VAV may have ReheatCoil as a part.

It'd be unnecessary to have *.py or *.json abstraction instead of directly using *.ttl for shapes because SHACL shapes are pretty readable in ttl alone. However, we can reimplement this later once we find more complex patterns which should be handled by scripts.

TODO
- [x] ~~examples~~
- [x] ~~documentation~~
decided to put those in the website later.

More controversial changes (maybe in a subsequent PR)
- [x] Add `not hasLocation` for `Points`.


tests/test_shapes.py
```diff
@@ -0,0 +1,75 @@
+import sys
+from pdb import set_trace as bp
+import rdflib
+import brickschema
+from bricksrc.namespaces import A, OWL, RDFS, SKOS, BRICK, SH, BSH, bind_prefixes
+import pyshacl
+
+schema_g = brickschema.Graph().parse('shacl/BrickShape.ttl', format='turtle')
```
gtfierro, line 8: You can use `.load_file('shacl/BrickShape.ttl')` here

tests/test_shapes.py
```diff
@@ -0,0 +1,75 @@
...
+:point a brick:Point.
+:loc a brick:Location.
+"""
+
+def test_no_relations():
+    data = base_data
+    data_g = rdflib.Graph().parse(data=data, format='turtle')
+    conforms, results_graph, results_text = pyshacl.validate(
+        data_g, shacl_graph=schema_g,
+    )
```
gtfierro, line 28: If you use `brickschema.Graph` then the `g.validate()` method maps to the pyshacl function with the same arguments

tests/test_shapes.py
```diff
@@ -0,0 +1,75 @@
...
+
+def test_equip():
+    valid_data = base_data + """
+:equip brick:hasLocation :loc.
+"""
+    valid_g = rdflib.Graph().parse(data=valid_data, format='turtle')
+    conforms, results_graph, results_text = pyshacl.validate(
+        valid_g, shacl_graph=schema_g
+    )
+    assert conforms
```
gtfierro, line 39: Same here; you can simplify syntax with the brickschema Graph. I think we also want to throw away (or assign to `_`) the returned values we aren't using

gtfierro (changes requested):
Some minor changes requested to bring the implementation to use the brickschema package, but otherwise this looks good. Having some documentation added to docs.brickschema.org would I think be a helpful component; currently the [validation page](https://docs.brickschema.org/lifecycle/validation.html) is pretty sparse. That wouldn't need to be part of this PR though

epaulson (approved):
This looks good to me - though I really don't have any experience with SHACL so that might not be saying much. But I was able to follow what was going on and I'm pretty sure I understand and think it makes sense.

I do think long-term want to move the constraints into the python dicts where we're declaring classes, but we have plenty of time to think about that.

tests/test_shapes.py
jbkoh: I was editing things, and then thought it's better not using brickschema package when it doesn't provide a large benefit. It's a unnecessary dependency while also causing circular dependencies.

jbkoh:
@gtfierro `sh:messages` have been added.

tests/test_shapes.py
gtfierro: The dependency already exists and it isn't circular. I feel like it is better to keep the implementation of certain actions, such as validation, consistent across the codebase

jbkoh: I meant the circular dependency in functionality of those two repositories in general. The python package helps to easily use the ontology, while we are using that package to validate the ontology. Anyway, we don't have to waste our energy here, and I can use brickschema package.

The package currently does not support loading from string data, which is used in this code. I'd appreciate if you can update the package to enable that. (another function called `load_data`?)

gtfierro: The `brickschema.Graph` object is a subclass of `rdflib.Graph` so you can use the same methods here (e.g. `g.parse`)

jbkoh:
@gtfierro please approve the change at you convenience.

gtfierro (approved):
Thanks!

PR #241, opened 2021-03-16
Fix Nightly Builds

gtfierro:
This PR fixes the following behavior:
- should have 1 existing nightly build that gets regularly replaced
- should not break when the main build doesn't break


gtfierro:
Tested this on https://github.com/gtfierro/reasonable/releases and it seems to work well

Issue #243, opened 2021-03-17
No dcterms:version in Dublin Core namespace


gatemezing:
Hello, 
Also, I found a small typo? in the ontology file. The property `dcterms:version` seems not present in the Dublin Core namespace.  I suggest the use of `owl:versionInfo` because the range of the`dcterms:hasVersion` seems not suitable as well. HTH.

gtfierro:
Thanks for catching that! Don't know how we ended up with something that doesn't even exist in the ontology. #245 should address these issues

Issue #242, opened 2021-03-17
Adding more metadata in Brick Ontology


gatemezing:
Hi guys, I just found this awesome ontology. I want to add it into [LOV](https://lov.linkeddata.es/) so that people can easily reuse it in the LOD space. However, I suggest adding more metadata in the ontology, such as the first issued date, modified date each time there is a new release, etc. Something along the following:

> brick: a owl:Ontology ;
 rdfs:label "Brick" ;
	dcterms:description "A uniform metadata schema for buildings";
    dcterms:creator ( [ a sdo:Person ;
                sdo:email "gtfierro@cs.berkeley.edu" ;
                sdo:name "Gabe Fierro" ] [ a sdo:Person ;
                sdo:email "jbkoh@eng.ucsd.edu" ;
                sdo:name "Jason Koh" ] ) ;
    dcterms:license <https://github.com/BrickSchema/brick/blob/master/LICENSE> ;
    dcterms:version "1.2.0" ;
	dcterms:issued "2020-12-20" ; #first issued date
	dcterms:modified "2021-02-21" ; # your last modified date
	dcterms:publisher <https://brickschema.org/consortium> ;
    rdfs:isDefinedBy <https://brickschema.org/schema/1.2/Brick> ;
    rdfs:seeAlso <https://brickschema.org> .

gtfierro:
Hello @gatemezing! Thanks for the suggestion to add more metadata to Brick . I've just filed #245 which adds the items that you mentioned. The PR will amend the ontology definition to the following:

```ttl
<https://brickschema.org/schema/1.2/Brick> a owl:Ontology ;
    rdfs:label "Brick" ;
    dcterms:creator ( [ a sdo:Person ;
                sdo:email "gtfierro@cs.berkeley.edu" ;
                sdo:name "Gabe Fierro" ] [ a sdo:Person ;
                sdo:email "jbkoh@eng.ucsd.edu" ;
                sdo:name "Jason Koh" ] ) ;
    dcterms:issued "2016-11" ;
    dcterms:license <https://github.com/BrickSchema/brick/blob/master/LICENSE> ;
    dcterms:modified "2021-03-23" ;
    dcterms:publisher [ a sdo:Consortium ;
            sdo:legalName "Brick Consortium, Inc" ;
            sdo:sameAs <https://brickschema.org/consortium/> ] ;
    rdfs:isDefinedBy <https://brickschema.org/schema/1.2/Brick> ;
[...]

PR #245, opened 2021-03-23
Update Ontology Metadata

gtfierro:
- dcterms:publisher
- dcterms:issued
- dcterms:modified (will be updated whenever Brick is built)
- owl:versionInfo instead of dcterms:version

Should address the feedback in #242 and #243

### DIFF
--- a/bricksrc/ontology.py
+++ b/bricksrc/ontology.py
@@ -20,7 +20,7 @@ ontology = {
         },
     ],
     DCTERMS.license: URIRef("https://github.com/BrickSchema/brick/blob/master/LICENSE"),
-    DCTERMS.issued: Literal("2016-11"),
+    DCTERMS.issued: Literal("2016-11-16"),
     DCTERMS.modified: Literal(datetime.now().strftime("%Y-%m-%d")),
     DCTERMS.publisher: {
         RDF.type: SDO.Consortium,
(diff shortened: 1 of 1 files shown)


bricksrc/ontology.py
```diff
@@ -8,13 +9,25 @@
...
-        {SDO.email: Literal("jbkoh@eng.ucsd.edu"), SDO.name: Literal("Jason Koh")},
+        {
+            RDF.type: SDO.Person,
+            SDO.email: Literal("jbkoh@eng.ucsd.edu"),
+            SDO.name: Literal("Jason Koh"),
+        },
     ],
     DCTERMS.license: URIRef("https://github.com/BrickSchema/brick/blob/master/LICENSE"),
-    DCTERMS.version: Literal(BRICK_FULL_VERSION),
+    DCTERMS.issued: Literal("2016-11"),
```
gatemezing, line 23: dcterms:issued should probably be in the form "YYYY-MM-DD".

gatemezing (changes requested):
Thanks for the update. Just found a small typo in the Literal for dcterms:issued.

bricksrc/ontology.py
gtfierro: The abbreviated dates are technically allowed [according to the purl docs](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http%3a%2f%2fpurl.org%2fdc%2felements%2f1.1%2fdate) but I can update it to a full date

Issue #228, opened 2021-02-12
Measurements for air quality analysis


DanieleAntonucci:
There are some parameters that we measure to evaluate the air quality.  Some of them have been already implemented few days ago. 

**Parameter:                        Unit**
Carbon monoxide :              [ppm]
Formaldehyde :                     [ppm]
VOC or tVOC :                       [ppb] or [ppm] or [ug/m3]
Particulate_matter_1 :          [ug/m3]
Particulate_matter_2.5 :      [ug/m3]
Particulate_matter_10 :        [ug/m3]
Ozone :                                  [ppm]
Nitrogen_dioxide :                [ppm]

Thanks

Daniele

epaulson:
We might want to add radon and methane?

Also the ASHRAE COVID air quality resources might be good to take a look over but that might be better in a separate issue
https://www.ashrae.org/file%20library/technical%20resources/covid-19/ashrae-building-readiness.pdf

PR #247, opened 2021-03-25
Adding quantities, sensor classes for air quality

gtfierro:
Will address https://github.com/BrickSchema/Brick/issues/228 when finished.

Changes include new quantities:
- methane, ozone, formaldehyde, NO2 concentration
- radioactivity, radon concentration

as well as new "level" sensors for each of those quantities. 

- [X] add quantities
- [X] add sensor classes
- [x] definitions for all quantities and sensors
- [x] example models showing how to use the new classes and quantities


gtfierro:
It is still in progress, but @DanieleAntonucci you should take a look at the example I put together [here](https://github.com/BrickSchema/Brick/tree/8d8bd231af2b338be01cd77d718160e18d6cd35d/examples/air_quality_sensors). Is the usage of the new classes clear in the TTL file? Does this cover your intended use case?

DanieleAntonucci:
Hi Gabe, The example is very clear. One question regarding the parameters. I didn't find the tVOC and VOC. will they be included? Thanks

gtfierro:
There is already a TVOC sensor in Brick (see [here](https://github.com/BrickSchema/Brick/blob/deabee257c84894acfe41a2c1919a651ff772bbd/bricksrc/sensor.py#L263-L272)). I am open to add a VOC sensor as well, but do you have a resource that would help me understand what is appreciably different between VOC and TVOC?

PR #248, opened 2021-03-26
generalizing solar panel and adding its subclasses

jbkoh:
The original discussion: https://groups.google.com/u/1/g/brickschema/c/hvverLxC16A/m/NlvGEvcJBgAJ

- Generalize the definition of `Solar_Panel` as proposed by @DanieleAntonucci
- Introduces subclasses as proposed by Ambuj

### DIFF
--- a/bricksrc/definitions.csv
+++ b/bricksrc/definitions.csv
@@ -764,7 +764,10 @@ https://brickschema.org/schema/Brick#Site,A geographic region containing 0 or mo
 https://brickschema.org/schema/Brick#Smoke_Alarm,An alarm that indicates the off-normal conditions associated with smoke.,
 https://brickschema.org/schema/Brick#Solar_Azimuth_Angle_Sensor,Measures the azimuth angle of the sun,
 https://brickschema.org/schema/Brick#Solar_Irradiance,The power per unit area of solar electromagnetic radiation incident on a surface,
-https://brickschema.org/schema/Brick#Solar_Panel,A colloquial term for a photovoltaic module. A Solar panel is an assembly of photovoltaic cells mounted in a framework for installation that use sunlight as a source of energy to generate direct current electricity.,https://en.wikipedia.org/wiki/Solar_panel
+https://brickschema.org/schema/Brick#Solar_Panel,A panel exposed to radiation from the sun used to heat water or when mounted with solar cells to produce electricity direct esp for powering instruments in satellites.,https://www.thefreedictionary.com/solar+panel
+https://brickschema.org/schema/Brick#Solar_PV_Panel,An integrated assembly of interconnected photovoltaic cells designed to deliver a selected level of working voltage and current at its output terminals packaged for protection against environment degradation and suited for incorporation in photovoltaic power systems.,https://encyclopedia2.thefreedictionary.com/pv+module
+https://brickschema.org/schema/Brick#Solar_Thermal_Panel,A type of solar panels that converts solar radiation into thermal energy.,
+https://brickschema.org/schema/Brick#Solar_PVT_Panel,A type of solar panels that convert solar radiation into usable thermal and electrical energy,
 https://brickschema.org/schema/Brick#Solar_Radiance,The amount of light that passes through or is emitted from the sun and falls within a given solid angle in a specified direction,
 https://brickschema.org/schema/Brick#Solar_Radiance_Sensor,The amount of light that passes through or is emitted from the sun and falls within a given solid angle in a specified direction,
 https://brickschema.org/schema/Brick#Solar_Zenith_Angle_Sensor,Measures the zenith angle of the sun,
--- a/bricksrc/definitions.csv
+++ b/bricksrc/definitions.csv
@@ -764,10 +764,9 @@ https://brickschema.org/schema/Brick#Site,A geographic region containing 0 or mo
 https://brickschema.org/schema/Brick#Smoke_Alarm,An alarm that indicates the off-normal conditions associated with smoke.,
 https://brickschema.org/schema/Brick#Solar_Azimuth_Angle_Sensor,Measures the azimuth angle of the sun,
 https://brickschema.org/schema/Brick#Solar_Irradiance,The power per unit area of solar electromagnetic radiation incident on a surface,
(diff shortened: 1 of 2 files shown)


david-waterworth:
Looks like I didn't 'reply all' when I responded to the original google group messages so I'll copy here.

'My book [Kalogirou, Soteris A. "Solar Energy Engineering: Processes and Systems"] actually splits these into totally separate concepts. Solar Energy Collectors are referred to as "special kinds of heat exchangers that transform solar radiation energy to internal energy of the transport medium". There's a large number of examples many of which aren't flat i.e. Flate-plate collector (FPC), Stationary compound parabolic collector (CPC), Evacuated tube collector (ETC) etc. There are 2 chapters on these systems before they introduce Photovoltaic Systems and Photovoltaic Panels.'

jbkoh:
@david-waterworth Could you elaborate what you are proposing based on the quote?

david-waterworth:
That a Solar Collector / System is a different concept to a Photovoltaic Panel / System. The former is a heat exchanger that "transforms solar radiation energy to internal energy of the transport medium" and the latter converts solar radiation to electrical energy. A PV System will consist of an array, inverter, electrical meters etc. They're quite different components to a Solar Collector System as far as I'm aware.

david-waterworth:
And a couple of additional thoughts. I think an array is a more useful concept than a panel at the brick level. An array is a collection of panels, usually all with the same axis and asimuth (direction and tilt). An array would have telemetry, only the angle metadata and perhaps array size and location. A site can have multiple arrays (different roofs for example). Then for the electrical sensors they'd be associated with an Inverter. Again there can be one or more and each can invert multiple arrays.

In my experience the inverters usually aren't integrated with the BMS, we have to directly interface to access the sensor data.

HervePruvost:
Hi Everyone,

thank you for the conceptualization of solar thermal panels or collectors. While working on some pilot building that includes a solar thermal HVAC system in a research project, I also identified this gap in Brick. But by including solar collectors some other gaps appeared to me as a snow ball effect.

Usually the solar collector is connected to a heat storage system like a water tank that accumulates and stores the heat produced. Moreover, in that building the solar system is coupled with a heat pump which also assists heat generation and distribution to a floor heating system. I could not find these concepts as well in Brick, so maybe we could add corresponding concepts. For example like this:

| missing concept | proposed superclass | proposed class |
| ------------------ | ---------------------- | ------------------|
| tank | brick:Energy_Storage | brick:Tank |
| floor heating | brick:Terminal_Unit | brick:Floor_Heating |
| heat pump | brick:HVAC_Equipment | brick:Heat_Pump |
| Heat pump (water/water) | brick:Heat_Pump | brick:Water_Water_Heat_Pump |
| Heat pump (water/air) | brick:Heat_Pump | brick:Water_Air_Heat_Pump |
| Heat pump (air/air) | brick:Heat_Pump | brick:Air_Air_Heat_Pump |
| Heat pump (air/water) | brick:Heat_Pump | brick:Air_Water_Heat_Pump |

The current class "Energy_Storage" is a subclass of "Electrical_Equipment" but maybe it should be generalized 
as there are several types of storage (for heat, electricity...). In that sense "Energy_Storage" would rather fit under "Equipment".

I will put this post in the issues tracker cause it is getting out of scope of the primary pull request.

BR
Hervé

jbkoh:
Sorry that I was dormant in this week. I really appreciate all the inputs here. I am summarizing my thoughts here based on the above comments.

@david-waterworth 
- Remove Solar_Panel, and introduce PV_Panel and Solar_Collector. Even though the lexical meaning of solar panels includes both PV panels and solar collectors, it's widely accepted as PV panels. Thus, I replace Solar_Panel with PV_Panel. PV panels and solar collectors both exploit solar radiation, but their usage model is largely different so I don't see a strong benefit of having a super class of them in terms of usability (i.e., people would be confused.)
- It makes sense to introduce PV_Array. The keyword of `System` reserves a specific meaning in Brick like `HVAC_System` and `Lighting_System`, and `Solar_System` would mean a full system including energy storages, etc. PV_Array seems to be the right term. I expect PV_Array may have associated telemetry related to solar tracking, condition of the arrays, etc., but we can add them later with concrete evidences.

@HervePruvost 
- It makes sense to generalize Energy_Storage. I will update that in the next commit.
- Brick has `Heat_Exchanger`. Are `Heat_Pumps` different from `Heat_Exchangers`?
- Tanks are a useful concept, though `Tank` can be used in many ways, I think, like water tanks for domestic usage. I would add that as a top class for now.
- It's correct that floor heating is one of the missing concepts in Brick yet, but I'd rather not use a gerund for the class name. Would `Radiant_Floor` work for you?

jbkoh:
@HervePruvost `Heat_Pump` is being added in this PR separately: https://github.com/BrickSchema/Brick/pull/252.

HervePruvost:
@jbkoh  Hello, sorry for this later reply. I was quite under water last week preparing a project demo.

I could catch up the discussion #252 about heat pumps in the last working group meeting.

To your questions:

- Generally speaking, `Heat_Exchanger` and `Heat_Pump` are different equipment I think since a heat exchanger is an element that transfers heat between non-mixing media over plates, shells or tubes. Even if the goal is the same by heat pumps, a heat pump is a rather more complex piece of equipment whose name refers to the thermodynamic process used inside it. In fact heat pumps can be inside composed of heat exchangers, like tube heat exchangers generally.
- You are right `Radiant_Floor` is a better name that points to the underfloor as a heat distribution equipment. Floor heating is too general and relates maybe more to the technolgy or system used for heat distribution.

gtfierro:
Hi all

I think it makes sense to keep this PR specifically about the solar related equipment. I've been working on the collection types (including PV array and Battery Energy Storage System) in https://github.com/BrickSchema/Brick/pull/263 which would complement these nicely. @HervePruvost I can follow up on your other issue to discuss the heat pumps and solar collectors.

@jbkoh are there any other changes you'd like to make to this PR or can we look at merging it soon?

jbkoh:
Ready for review. The PR is limited to classes for solar equipment. I removed `Solar_Panel` as described in a previous comment in this thread.

gtfierro (approved):

Issue #244, opened 2021-03-22
Description of hasPoint


david-waterworth:
The description element of "hasPoint" is a bit misleading ("The subject has a digital/analog input/output point given by the object") since points can be AV/BV's or other. The description from https://docs.brickschema.org/brick/relationships.html seems more appropriate "the subject has a source of telemetry identified by object".

gtfierro:
Thanks for the feedback! I agree that the other definition is a bit more clear, but it is still made in the context of the full document. I'll make a note to revisit the description of that relationship soon

gtfierro:
@david-waterworth can you take a look at #251 when you get a chance and let me know if the updated definitions work for you?

PR #251, opened 2021-03-30
Amend definition for isPointOf/hasPoint

gtfierro:
Addresses the comment in #244 and brings the definition in the ontology up to speed with what is in the [relationship documentation](https://docs.brickschema.org/brick/relationships.html)


david-waterworth (approved):
Looks fine to me.

PR #237, opened 2021-02-22
Added new classes

carlosduarteroa:
Proposed the addition of new classes that I created based on going through the process of developing Brick models for three buildings. Two of the buildings have a [radiant heating and cooling system](https://en.wikipedia.org/wiki/Radiant_heating_and_cooling) for their primary HVAC system.

PR #252, opened 2021-04-01
Heat pump classes

carlosduarteroa:
Proposed new classes to model various heat pump types.

Note: Cherried-picked commits c84411b and 4c97fbb from PR #237 to make this new PR.


bricksrc/equipment.py
```diff
@@ -1,3 +1,4 @@
+from typing import Reversible
```
gtfierro, line 1: Hey @carlosduarteroa this looks like something that was pulled in automatically by an IDE. Is this required by the implementation or can we remove it?

gtfierro:
Thanks @carlosduarteroa! Hope it was fun learning how to cherry-pick :). I'll add notes from the working group meeting soon

bricksrc/equipment.py
carlosduarteroa: I don't remember adding any additional packages and I think it is not necessary. We can remove it since I do use it. I wonder how it got there. I didn't know IDE could automatically pull in packages.

gtfierro: Yeah, I noticed things like that happening when students were getting build errors in a class project because their IDE decided to be "helpful"

gtfierro (changes requested):
@carlosduarteroa this looks really good. Few small changes:

- remove the `from typing import ...` from the top of the equipment file
- instead of `TAG.Water_Air`, just use the simpler tags like `TAG.Air` and `TAG.Water`. We aren't using the tags for definitions, they're just annotations

bricksrc/equipment.py
carlosduarteroa: I removed this line.

carlosduarteroa:
@gtfierro I made the requested changes. Also, we talked about refrigerant to air heat pumps (VRF) but are not included in the current structure. Should I go ahead and integrate these types of systems into this pull request?

gtfierro:
I think that would be a good idea. If you're able to finish that by this afternoon then we can discuss the design in the working group meeting today and next week.

gtfierro (approved):
I'm approving this but I'm going to merge it into a feature branch for heat pump models. That way multiple PRs can be brought together before it is finally merged in

PR #257, opened 2021-04-13
Heat Pump / Chiller Plant Model

gtfierro:
Target branch for PRs related to heat pumps and chiller plants:

- [X] heat pump classes (https://github.com/BrickSchema/Brick/pull/252)
- [x] example configurations in the `examples/` folder
- [X] discuss and decide on the `brick:connectsTo` property
- ~~add documentation to explain the role of `enumerations` for dynamic graph structures~~

Development is being done through the Ontology Working Group. Recent discussions can be found on the [Brick wiki](https://github.com/BrickSchema/Brick/wiki/); in particular, [this meeting](https://github.com/BrickSchema/Brick/wiki/Ontology-Working-Group-Notes#2021-04-06--2021-04-15) informed many of the design choices explored here.

---

* Added heat pump classes

* added non and reversible subclasses for hp

* broke up class tags into simpler tags

* added non- and -reversible air to air hp classes

* added air to refrigerant hp classes

* fixed names to title case

* broke up single heat pump tag into two tags

Co-authored-by: Gabe Fierro <gtfierro@cs.berkeley.edu>


MatthewSteen:
Here are my comments based on experience as a design consultant doing energy modeling of new commercial buildings. I haven't been in the working group meetings, so I'm sorry if these issues have already been discussed. 

I'm a proponent of using established standards whenever possible, so I think the following should be considered.

- ASHRAE Terminology for formal definitions.
- [OmniClass Table 21 (ISO 12006) and Table 22](https://www.csiresources.org/standards/omniclass/standards-omniclass-about)  for class hierarchy (I haven't found a better standard yet). 

## General

1. Heat pump equipment by definition have a reversing valve and can both cool and heat, so I don't think it makes sense to have NonreversibleHeatPump and ReversibleHeatPump classes, which seems overly complex and potentially confusing. A non-reversible heat pump is just an air conditioner right (chiller or cooler more generally)? 

   https://xp20.ashrae.org/terminology/index.php?term=heat+pump&submit=Search
   > heat pump 
thermodynamic heating/refrigerating system to transfer heat. The condenser and evaporator may change roles to transfer heat in either direction. By receiving the flow of air or other fluid, a heat pump is used to cool or heat. Heat pumps may be the air source with heat transfer between the indoor air stream to outdoor air or water source with heat transfer between the indoor air stream and a hydronic source (ground loop, evaporative cooler, cooling tower, or domestic water).

   For example, do `Nonreversible Ground to Air Heat Pumps` or `Nonreversible Water to Air Heat Pumps` even exist? I don't know of GSHPs or WSHPs without a reversing valve that only cool. Similarly, a `Nonreversible Air to Air Heat Pump` is just an AC. 

2. With (1) in mind, I don't think Chillers are a child of HeatPump. There are "heat pump chillers" or "reversing chiller" (e.g. Denver Water's central plant), but those are heat pumps since they also heat. 

3. Consider using [ASHRAE's Chiller definitions](https://xp20.ashrae.org/terminology/index.php?term=chiller&submit=Search)

MatthewSteen:
## Heat Pumps
Heat pumps cool or heat air, water, or refrigerant by moving heat from a source to a sink via the VCR cycle.
https://xp20.ashrae.org/terminology/index.php?term=heat+pump&submit=Search
> heat pump 
thermodynamic heating/refrigerating system to transfer heat. The condenser and evaporator may change roles to transfer heat in either direction. By receiving the flow of air or other fluid, a heat pump is used to cool or heat. Heat pumps may be the air source with heat transfer between the indoor air stream to outdoor air or water source with heat transfer between the indoor air stream and a hydronic source (ground loop, evaporative cooler, cooling tower, or domestic water).

- Air-Source Heat Pump = air is the heat source/sink 
- Ground-Source Heat Pump = ground is the heat source/sink
- Water-Source Heat Pump = water is the heat source/sink (typically a condenser loop served by cooling tower and boiler)

### OmniClass Table 22
Note, it's missing GSHP, but I think that might be a subclass/child of WSHP based on the ASHRAE definition. Not to be confused with _geothermal_ electric generation equipment.
image

MatthewSteen:
## Chillers
Consider using [ASHRAE's Chiller definitions](https://xp20.ashrae.org/terminology/index.php?term=chiller&submit=Search).

### OmniClass Table 22
image

gtfierro:
For this PR:
- pull in substances
- `connectsTo` property

MatthewSteen:
Link to gist that I don't think is included here. https://gist.github.com/gtfierro/c5ed6e25a99c60053c43befbebe1f740

gtfierro:
I like the idea of going with the Table 23 chiller hierarchy. Would we want to put 'split system' chillers vs 'packaged' chillers in different trees? Then we could have rules about those different kinds of subsystems while still preserving the difference between screw/scroll/etc in the subclasses

epaulson:
Just as a random comment, I found a fun site from PNNL (it's got quite a pedigree - it's domain is still PNL and not PNNL) with some useful documentation, and also a good example of what a real application is looking for. Here's the help: 
https://energycode.pnl.gov/COMcheckWeb/robohelp/Software_Compliance_Guide.htm

the docs are all frames so you might have to dig around, but a good page on just how they use chillers: https://energycode.pnl.gov/COMcheckWeb/robohelp/Plant_Cooling_Equipment_Details.htm

epaulson:
Screenshot 2023-06-04 at 7 54 45 PM

MatthewSteen:
@epaulson following up on this... 

## OmniClass
This could be a good starting point. However, I'm not sure I agree with their organization levels of compressor type => condenser *location* (packaged vs. split). Condenser *type* (air-cooled, evap-cooled, water-cooled, etc.) might be a better primary organization level. 

AHRI could also serve as a good starting point. https://www.ahrinet.org/sites/default/files/2023-02/AHRI%20Standard%20550%20590-2023%20I-P_0.pdf

## ComCheck
I've used this tool a few times for energy code compliance work and it *should* just align with ASHRAE 90.1, which then refers to other equipment-specific standards such as AHRI.

MatthewSteen:
Not sure if this belongs here or in #197 but Brick needs the concept of `Condenser_Water_Loop`, which is a major gap. Also important for Guideline 36 work in BuildingMOTIF.

epaulson:
I'm doing a Thanksgiving tab sweep, and one thing I've had open for a bit, kinda related: JCI puts out BIM models of its chillers - here's one where we put in a suggested Omniclass classification (among other classification systems) and even include some of the nameplate properties:  

https://www.bimobject.com/en-us/johnsoncontrols/product/jci-yvaa

gtfierro:
Closing so we can continue development in a single location on a more recent version of Brick (see #589)

connorjcantrell:
@MatthewSteen @gtfierro Here is the new PR: https://github.com/BrickSchema/Brick/pull/590

PR #261, opened 2021-04-19
Make sure subPropertyOf is added to Entity Properties

gtfierro:
Simple fix --- just forgot the annotation on the recursive call when generating the entity properties.


gtfierro:
duplicate of https://github.com/BrickSchema/Brick/pull/261

Issue #233, opened 2021-02-16
Container Classes for Grouping Entities


gtfierro:
There has been some call from the community to provide classes that define collections of other Brick entities:
- *Location Portfolio* or *Project* for a collection of Sites
- *System* for a group of Equipment + Points assembled for a particular set of use cases
- *Loop* for a discrete group of Equipment + Points through which a substance flows (a Loop may be a part of a System?)

List others! We should be able to include some of these in a future 1.2.x release

epaulson:
Just to come back to 'Location Portfolio', Portfolio seems to be the term of the art as one of the top level containers for buildings or site. CRE firms talk about their 'portfolios' of buildings and there's the Energy Star Portfolio Manager which you can "Use [PM] to benchmark the performance of one building or a whole portfolio of buildings, all in a secure online environment." 

(Unfortunately, PortfolioManager has a portfolio of properties in their UI rather than a portfolio of buildings, which isn't going to make life less confusing for graph database users)

So you do wind up seeing 'portfolio' as the top-level data structure in the UIs of these sorts packages - if you collapse all navigation trees you'll see only the portfolios on left sidebars. For analytics, we'll also want to be able to exchange "portfolios" of buildings to do benchmarking against other portfolios.

I don't think "project" is the right term for this.

david-waterworth:
A couple of other collections I've come across recently are AHU's and FCU's. Each floor of a building has a number of AHU's and FCU's, each of which has it's own BACnet controller. There's also a controller responsible for the overall coordination of all the floor AHU's and another for all the FCU's. Each has its own set of points which are shared with the AHU's / FCU's which are being controlled - points such as average floor temperature etc.

I'm not 100% sure but I suspect it's easier to program a BACnet system where shared points are written/read to/from a single central device (hub / spoke model) rather than each controller having to read from all their peers.

I'm not 100% sure that this is necessary for a Brick model, in our proprietary schema we did create a "Common AHU" and "Common FCU" class which is referred to by all the AHU's / FCU's which share points between the floor controllers. You could probably just model the points as sensors etc located on Floor X and not associated them with a specific controller but I thought I'd raise it anyway.

gtfierro:
Please see https://github.com/BrickSchema/Brick/pull/269 for the collection types implementation

I think in the future we may explore collections which involve controllers, etc but I would like to postpone that until we have a better treatment of BMS concepts

david-waterworth:
That makes sense. BTW here are some examples:

[...]

PR #263, opened 2021-04-20
Solar Arrays and other Collections

gtfierro:
This resolves the modeling needs described in https://github.com/BrickSchema/Brick/issues/233 .

Introduces a new top-level Brick class called `Collection`, which itself is not a Location, Equipment etc but rather an unordered set of those components. A Portfolio is a collection of Sites; a System is a collection of Equipment and/or Loops; a Loop is a collection of connected Equipment.

There are also a few subclasses included that are mentioned in [this discussion](https://groups.google.com/g/brickschema/c/oIS9JZqFuHY) about PV systems.

TODOs:
* [x] add examples for how these classes are meant to be used 
* [x] add "air loop"?
* [ ] add documentation to docs.brickschema.org
* [ ] add SHACL shapes to constrain to correct usage
* [x] add entity properties for PV panels/arrays? (angle, orientation)

How to handle static vs dynamic orientation + tilt? Also "MPPT" for solar panels


gtfierro:
Closing this PR to split into multiple

PR #265, opened 2021-04-28
Add building electrical meter example

gtfierro:
Simple meter with a few sensors; timeseries identifiers are also included

Might also want another example for a more complete submetering infrastructure

### DIFF
--- a/examples/building_meter/generate.py
+++ b/examples/building_meter/generate.py
@@ -1,0 +1,56 @@
+import brickschema
+from brickschema.namespaces import A, OWL, BRICK, UNIT
+from rdflib import Namespace, Literal
+
+# our entities will live in this namespace
+BLDG = Namespace("urn:example#")
+
+# create a graph for our model
+g = brickschema.Graph()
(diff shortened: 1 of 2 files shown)


examples/building_meter/generate.py
```diff
@@ -0,0 +1,56 @@
...
+g.bind("bldg", BLDG)
+
+# define a building in a site
+g.add((BLDG["mysite"], A, BRICK.Site))
+g.add((BLDG["mybldg"], A, BRICK.Building))
+g.add((BLDG["mysite"], BRICK.hasPart, BLDG["mybldg"]))
+
+# add a full building meter to the building
+g.add((BLDG["meter"], A, BRICK.Building_Electrical_Meter))
+g.add((BLDG["mybldg"], BRICK.hasPart, BLDG["meter"]))
```
jbkoh, line 19: @gtfierro meters cannot be a part of buildings based on `hasPart`'s definition.

gtfierro:
You're right! I had that as hasLocation before; I'll fix that soon

PR #271, opened 2021-05-07
Radiant system temperature sensors

carlosduarteroa:
Added temperature sensors and setpoints that are useful for radiant systems.


gtfierro:
Thanks for this and #270! Are these the last of the classes to be added from https://github.com/BrickSchema/Brick/pull/237? If so, do you mind if I close #237 and focus on #271 and #270?

gtfierro:
Thanks for the heads up! We can discuss the PRs you've finished during the ontology WG meeting next Thursday

gtfierro (approved):

bricksrc/sensor.py
```diff
@@ -1666,6 +1666,63 @@
...
+                                ],
+                                "subclasses": {
+                                    "Core_Temperature_Sensor": {
+                                        "tags": [
+                                            TAG.Point,
+                                            TAG.Core,
+                                            TAG.Temperature,
+                                            TAG.Sensor,
+                                        ],
+                                    },
```
gtfierro, line 1721: A few comments:
- Given that "embedded" is an overloaded term (it may refer to a temperature sensor on an "embedded" platform such as a sensor mote), what might be a clearer word we can use for Embedded (and Core) temperature sensors? I understand these may be "terms of art", so maybe something like `Embedded_Panel_Temperature_Sensor` and `Core_Panel_Temperature_Sensor` (or `Panel_Embedded...`, etc)?

bricksrc/sensor.py
```diff
@@ -1666,6 +1666,63 @@
...
+                            "Embedded_Temperature_Sensor": {
+                                "tags": [
+                                    TAG.Point,
+                                    TAG.Embedded,
+                                    TAG.Temperature,
+                                    TAG.Sensor,
+                                ],
+                                "substances": [
+                                    [BRICK.measures, BRICK.Temperature],
+                                    [BRICK.measures, BRICK.Internal_Material],
```
gtfierro, line 1711: Let's leave off the `[BRICK.measures, BRICK.Internal_Material]` for now. I think the better model would be to have `Concrete_Temperature_Sensor`, etc variations for the different kinds of materials that might exist in a radiant panel? Then we have the question if we need to capture whether it is Embedded Temperature or Core Temperature: is this a crucial distinction, and if so, why is Core a subclass of Embedded rather than being a sibling class?

carlosduarteroa: I agree that we could leave off the generic "internal material" to replace it for the specific materials.

The idea with the class `Embedded_Temperature_Sensor` is to be the generic version of a temperature measurement inside a concrete slab. We often don't know how deep the sensor is in the slab. The idea with `Core_Temperature_Sensor` was to specify that the measurement is at the same level as the hydronic pipe of the radiant system. The better way might be just to have one of the classes and then add the sensor depth by adding a property to the class.

bricksrc/sensor.py
carlosduarteroa: Good point! I need to think about this. I think the term needs to capture that it is a temperature measurement inside the material but not necessarily belonging to a radiant system.

bricksrc/sensor.py
gtfierro: Guess it broke when that property got removed. Let me give some thought for how to model this more appropriately

PR #274, opened 2021-05-10
Additional air handling unit types

carlosduarteroa:


bricksrc/equipment.py
```diff
@@ -439,6 +439,28 @@
     "AHU": {
         "tags": [TAG.Equipment, TAG.AHU],
         "subclasses": {
+            "Dedicated_Outdoor_Air_System": {
```
gtfierro, line 442: What do you think about using `Dedicated_Outdoor_Air_System_Unit` so that it is clear from looking at the term that this is an Equipment and not a System?

bricksrc/equipment.py
```diff
@@ -439,6 +439,28 @@
...
+                "tags": [
+                    TAG.Equipment,
+                    TAG.Dedicated,
+                    TAG.Outdoor,
+                    TAG.Air,
+                    TAG.System,
+                ],
+                OWL.equivalentClass: BRICK["DOAS"],
+            },
+            "DOAS": {
```
gtfierro, line 452: This could still be `DOAS` if that's a common abbreviation

bricksrc/equipment.py
carlosduarteroa: Yeah, I agree with your suggestion to make it clearer.

bricksrc/equipment.py
carlosduarteroa: Yeah, that is the common abbreviation.

carlosduarteroa:
Thanks for reviewing. I made the changes.

gtfierro (approved):

gtfierro:
Thanks @carlosduarteroa!

PR #276, opened 2021-05-11
Added Differential Pressure Points

carlosduarteroa:


gtfierro (approved):

gtfierro (commented):
One issue I need to address is the definition of differential quantities. QUDT [has support](https://github.com/qudt/qudt-public-repo/wiki/Support-for-measures-of-absolute-values-and-for-intervals-(differences)) for those types --- I just need to figure out how to best encode the `isDeltaQuantity` property. Nothing for you to do; just wait for https://github.com/BrickSchema/Brick/issues/284

PR #277, opened 2021-05-11
Added Differential Temperature Points

carlosduarteroa:
Added differential temperature points and water temperature setpoints


gtfierro (approved):

gtfierro (commented):
This also depends on https://github.com/BrickSchema/Brick/issues/284

gtfierro:
@carlosduarteroa the differential quantity definitions are merged now, so we are almost ready to merge in this PR and https://github.com/BrickSchema/Brick/pull/277 . My question for you is what kind of differential pressure best characterizes these points? Differential static pressure, differential dynamic pressure, or do none of these make sense?

carlosduarteroa:
A note on the differential pressure terminology, it can technically be used for either both (static, dynamic) differential pressures. When we say in the HVAC industry "static pressure" what we are actually referring to is the differential static pressure since we are usually measuring the difference between the outside (atmosphere) absolute static pressure and inside a duct absolute static pressure. My question is then if we want to be more precise in the `Differential_Pressure_Sensor` and `Static_Pressure_Sensor` classes and subclasses?

gtfierro:
Thanks for taking a look!

PR #278, opened 2021-05-11
Added Air System Classes

carlosduarteroa:
Classes used to describe an air loop.


gtfierro (approved):
I believe this is redundant with an earlier PR? I'll keep this around just in case I'm mis-remembering. Plus, it will be neat to see how GitHub handles duplicate diffs if that is the case

gtfierro (approved):

PR #281, opened 2021-05-12
Added Open Close Status Classes

carlosduarteroa:


gtfierro (commented):
How do you feel about having the `Open_Close_Status` just be the only point, and `hasPoint` relates it to the damper and valve? We can put some documentation in the `definitions.csv` to list the kinds of Brick entities that this may apply to

carlosduarteroa:
One of the things that I struggled with while building Brick models is trying to figure out best practices for defining these types of points. If you only have the one point, then you would need to explicitly define additional parts of the larger system. I think I came across this while using `Valve`, `Command`, and `Valve_Command` entities for a radiant terminal. It was much easier to build this part of the model by associating `Valve_Command` directly to the radiant terminal instead of associating `Command` to a `Valve` which in turn needs to be associated with the terminal. I didn't think adding a `Valve` part would add value to the model.

In short, I think it gives flexibility to the modeler to choose how detailed the model should be but it could also add ambiguity to it by skipping intermediate parts. So it's a tradeoff.

gtfierro:
For consistency we want to encourage the `Valve` to be modeled explicitly, rather than just suggested

PR #289, opened 2021-06-20
Updating Nightly Builds

gtfierro:
Getting some build errors over the last 3 days or so. 
- explicitly pulling v7.1.0 allegrograph container now
- build shacl shapes before generating Brick


gtfierro:
The build breaking seems to coincide with the [re-pushing of the allegrograph docker contaienr tags](https://hub.docker.com/r/franzinc/agraph/tags) 4 days ago

PR #291, opened 2021-07-01
Initial attempt of a System shape

jbulow:
While developing a UI tool for creating Brick datasets we found that the System class was lacking a shape specifying valid relations. I.e without hardcoding logic in the tool, we can not know what relations to use when defining a System of any kind.

Is this a reasonable beginning of a shacl shape for System?

Disclaimer: I'm a shacl newbie. 

Relevant mailing thread: [Systems and Relationships](https://groups.google.com/g/brickschema/c/oIzFf7dtWYs)

### DIFF
--- a/shacl/BrickEntityShapeBase.ttl
+++ b/shacl/BrickEntityShapeBase.ttl
@@ -120,7 +120,7 @@ bsh:PointShape a sh:NodeShape;
     .
 
 bsh:SystemShape a sh:NodeShape;
-    sh:targetClass brick:System;
+    sh:targetClass brick:Collection;
     sh:and (
         [sh:not [ sh:class brick:Equipment ] ; sh:message "System is an exclusive top class." ]
         [sh:not [ sh:class brick:Location ] ; sh:message "System is an exclusive top class." ]
--- a/shacl/BrickEntityShapeBase.ttl
+++ b/shacl/BrickEntityShapeBase.ttl
@@ -15,6 +15,7 @@ bsh:LocationShape a sh:NodeShape ;
         [sh:not [ sh:class brick:Equipment ] ; sh:message "Location is an exclusive top class." ]
         [sh:not [ sh:class brick:Substance ] ; sh:message "Location is an exclusive top class." ]
         [sh:not [ sh:class brick:Quantity ] ; sh:message "Location is an exclusive top class." ]
+        [sh:not [ sh:class brick:Collection ] ; sh:message "Location is an exclusive top class." ]
     );
     sh:property [
         sh:path brick:hasPart;
@@ -46,6 +47,7 @@ bsh:EquipmentShape a sh:NodeShape ;
         [sh:not [ sh:class brick:Location ] ; sh:message "Equipment is an exclusive top class." ]
         [sh:not [ sh:class brick:Substance ] ; sh:message "Equipment is an exclusive top class." ]
         [sh:not [ sh:class brick:Quantity ] ; sh:message "Equipment is an exclusive top class." ]
+        [sh:not [ sh:class brick:Collection ] ; sh:message "Equipment is an exclusive top class." ]
     );
     sh:property [
         sh:path brick:hasPart;
@@ -56,12 +58,17 @@ bsh:EquipmentShape a sh:NodeShape ;
         sh:property [
             sh:path brick:isPartOf;
             sh:class brick:System;
-            sh:message "Equipment can be part of a System or other Equipment"
+            sh:message "Equipment can be part of a Collection, System or other Equipment"
         ]
         sh:property [
             sh:path brick:isPartOf;
             sh:class brick:Equipment;
-            sh:message "Equipment can be part of a System or other Equipment"
+            sh:message "Equipment can be part of a Collection, System or other Equipment"
+        ]
+        sh:property [
+            sh:path brick:isPartOf;
+            sh:class brick:Collection;
+            sh:message "Equipment can be part of a Collection, System or other Equipment"
         ]
     ) ;
     sh:property [
@@ -95,6 +102,7 @@ bsh:PointShape a sh:NodeShape;
         [sh:not [ sh:class brick:Location ] ; sh:message "Point is an exclusive top class." ]
         [sh:not [ sh:class brick:Substance ] ; sh:message "Point is an exclusive top class." ]
         [sh:not [ sh:class brick:Quantity ] ; sh:message "Point is an exclusive top class." ]
+        [sh:not [ sh:class brick:Collection ] ; sh:message "Point is an exclusive top class." ]
     );
     sh:message "Point is an exclusive top class." ;
     sh:or (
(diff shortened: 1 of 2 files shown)


shacl/BrickEntityShapeBase.ttl
```diff
@@ -54,8 +56,11 @@ bsh:EquipmentShape a sh:NodeShape ;
     ];
     sh:property [
         sh:path brick:isPartOf;
-        sh:class brick:Equipment;
-        sh:message "A piece of Equipment's parts should be always other Equipment."
+        sh:or (
+            [ sh:class brick:Equipment ]
+            [ sh:class brick:System ]
+        ) ;
+        sh:message "An piece of Equipment can only be part of another piece of Equipment or a System."
```
gtfierro, line 63: I am pretty sure this needs to be phrased as two separate `sh:property` shapes:

```turtle
sh:or (
  sh:property [ ... ] 
  sh:property [ ... ]
)
```

due to SHACL's requirement that the arguments to `sh:or` are, themselves, valid shapes

gtfierro (changes requested):
Thanks for putting this together, and sorry that it took me so long to find cycles to comment on this! I'm in a new role now where i will have much more time to work on Brick :)

As part of https://github.com/BrickSchema/Brick/pull/269, we will be adopting a `Collection` type that is a superclass of `System`, `Loop` and other groups of entities. I think that the shape that you have below will still work for `Systems` (and can continue to refer to that concept), however, some of the shapes will need to be altered to account for `Collection` instead. I'll make the comments on the PR where this would need to be done.

If you're comfortable doing so, you could consider merging in the `add-collection-types` branch into this one to verify that the code works; else, we can just wait until #269 is merged and then this PR will automatically test against it

shacl/BrickEntityShapeBase.ttl
```diff
@@ -15,6 +15,7 @@ bsh:LocationShape a sh:NodeShape ;
         [sh:not [ sh:class brick:Equipment ] ; sh:message "Location is an exclusive top class." ]
         [sh:not [ sh:class brick:Substance ] ; sh:message "Location is an exclusive top class." ]
         [sh:not [ sh:class brick:Quantity ] ; sh:message "Location is an exclusive top class." ]
+        [sh:not [ sh:class brick:System ] ; sh:message "Location is an exclusive top class." ]
```
gtfierro, line 18: This should change to `Collection`

shacl/BrickEntityShapeBase.ttl
```diff
@@ -46,6 +47,7 @@ bsh:EquipmentShape a sh:NodeShape ;
         [sh:not [ sh:class brick:Location ] ; sh:message "Equipment is an exclusive top class." ]
         [sh:not [ sh:class brick:Substance ] ; sh:message "Equipment is an exclusive top class." ]
         [sh:not [ sh:class brick:Quantity ] ; sh:message "Equipment is an exclusive top class." ]
+        [sh:not [ sh:class brick:System ] ; sh:message "Equipment is an exclusive top class." ]
```
gtfierro, line 50: This should also change to `Collection`

shacl/BrickEntityShapeBase.ttl
```diff
@@ -54,8 +56,11 @@ bsh:EquipmentShape a sh:NodeShape ;
     ];
     sh:property [
         sh:path brick:isPartOf;
-        sh:class brick:Equipment;
-        sh:message "A piece of Equipment's parts should be always other Equipment."
+        sh:or (
+            [ sh:class brick:Equipment ]
+            [ sh:class brick:System ]
```
gtfierro, line 61: Equipment can be part of `Collection`s, not just `System`s with the new PR

shacl/BrickEntityShapeBase.ttl
```diff
@@ -84,6 +89,7 @@ bsh:PointShape a sh:NodeShape;
         [sh:not [ sh:class brick:Location ] ; sh:message "Point is an exclusive top class." ]
         [sh:not [ sh:class brick:Substance ] ; sh:message "Point is an exclusive top class." ]
         [sh:not [ sh:class brick:Quantity ] ; sh:message "Point is an exclusive top class." ]
+        [sh:not [ sh:class brick:System ] ; sh:message "Point is an exclusive top class." ]
```
gtfierro, line 92: `System` -> `Collection`

shacl/BrickEntityShapeBase.ttl
```diff
@@ -96,6 +102,23 @@ bsh:PointShape a sh:NodeShape;
...
+        [sh:not [ sh:class brick:Substance ] ; sh:message "System is an exclusive top class." ]
+        [sh:not [ sh:class brick:Quantity ] ; sh:message "System is an exclusive top class." ]
+        [sh:not [ sh:class brick:Point ] ; sh:message "System is an exclusive top class." ]
+    );
+    sh:message "System is an exclusive top class." ;
+    sh:property [
+        sh:path brick:hasPart;
+        sh:class brick:Equipment;
+        sh:message "A System can be associated with Equipments."
+    ];
```
gtfierro, line 119: Systems can have other Systems, Equipment, Locations and Points as parts

gtfierro: At least in the proposed #269 ; let me know if you have a different opinion! I've written up some basic docs on the idea here: https://github.com/BrickSchema/docs/pull/3/files

gtfierro:
[Link to proposed docs on the proposed #269 additions](https://github.com/BrickSchema/docs/pull/3/files)

shacl/BrickEntityShapeBase.ttl
```diff
@@ -111,6 +119,40 @@ bsh:PointShape a sh:NodeShape;
     ) ;
     .
 
+bsh:SystemShape a sh:NodeShape;
+    sh:targetClass brick:System;
```
gtfierro, line 123: I'm so sorry I missed this, but this should also be changed to `Collection` instead of `System`:  I think its just the `CollectionShape` and `brick:System`

gtfierro: Then the rest of the use of "System" should be changed to "Collection"

gtfierro (changes requested):
So sorry that I missed a couple last things! That should be the last of the changes

shacl/BrickEntityShapeBase.ttl
jbulow: I'm a bit confused. If I change SystemShape to CollectionShape then there will be no SystemShape. Should I add an CollectionShape besides the SystemShape? Currently, there is no CollectionShape.

gtfierro: We will want a CollectionShape in this file, yes. Brick actually generates a NodeShape for the `brick:System` class that is done automatically (see the last line):

```turtle
brick:System a owl:Class,
        sh:NodeShape ;
    rdfs:label "System" ;
    rdfs:subClassOf   brick:Collection ;
    skos:definition "A System is a combination of equipment and auxiliary devices (e.g., controls, accessories, interconnecting means, and termi­nal elements) by which energy is transformed so it performs a specific function such as HVAC, service water heating, or lighting. (ASHRAE Dictionary)."@en ;
    sh:property [ sh:or ( [ sh:class brick:Equipment ] [ sh:class brick:Point ] [ sh:class brick:Loop ] [ sh:class brick:System ] [ sh:class brick:Location ] [ sh:class brick:PV_Array ] ) ;
            sh:path brick:hasPart ] . 
```

We can definitely put a `SystemShape` in this file --- maybe even move the System shape definition to this file --- but we will need a `CollectionShape` nonetheless

gtfierro (approved):

gtfierro:
Nice work @jbulow! Thanks for sticking with me for the back-and-forth. I'll run the tests locally (I know they timeout on GH actions) and merge when they pass

jbulow:
> Nice work @jbulow! Thanks for sticking with me for the back-and-forth. I'll run the tests locally (I know they timeout on GH actions) and merge when they pass

Is there anything more I should do on this PR before commit?

gtfierro:
Nope! You are good to go. I am just working on fixing build times before I merge a few of these; hopefully should be ready to go by Monday of next week.

PR #297, opened 2021-07-14
Add UFAD and diffuser equipment and points

carlosduarteroa:


gtfierro:
This looks good to me! One question is: do we want to model an Air Plenum as an *equipment*? Is this how it is normally thought of, or is there a more natural organization?

carlosduarteroa:
Yes, I think an air plenum would be best under _equipment_. An air plenum is sometimes used in lieu of ducts to distribute air to or collect air from the zone.

gtfierro:
Ok! Thanks for the clarification, and for the contribution!

PR #298, opened 2021-07-18
Add Outside Fan to ontology

wcrd:
In line with Outside Air points and Outside Damper - I have added an Outside Fan. I have given it an equivalent class of 'Fresh_Air_Fan'. Would normally call this an _Outside Air Fan_ but in keeping with the convention for other fans I have dropped the _Air_ term.

It would also be my preference to remove the _Fresh_Air_Fan_ class; all I can hear is my old boss telling me "just because the air is coming from outside it does not mean it is _fresh_".

Thanks!


gtfierro:
Thanks for the edits!

One idea on how to approach this --- I'd rather have something that is more correct than something that is an unnecessary compromise. Should we make **Fresh Air Fan** a subclass of **Outside Air Fan**? That way we can distinguish between fans that are for fresh air vs fans that are just for the outside?

wcrd:
Hi Gabe,

That sounds ok - I will make those suggested changes to the pull request. I still have some reservations around 'fresh air' which I've noted below. 

From my perspective, it is not clear what a 'fresh air fan' is (as a subclass of outside fan). It probably serves some stream of outside air that meets some air quality standard - however it is not possible to say in the field without some measurement device in stream, and a given standard to follow. This would also mean that the stream could change from 'fresh air' to 'outside air' depending on the sensor readings, which would give the equipment a dynamic name.

I suppose fresh air could exist as the output of some process; where if it were then moved by some fan in that stream then a 'fresh air fan' could exist. Perhaps also in some sort of manufacturing/medical process, however this would be more akin to 'sterilized air' and not necessarily a subclass of 'outdoor air' I think - would be interesting to hear from the broader community on this as I have not dealt with systems that reference 'fresh air' before!

Based on a very limited literature review:

- Ashrae Standard 62.1: Ventilation for Acceptable Indoor Air Quality
- Trane: Air Conditioning Manual
- Ashrae Handbook: Fundamentals (2009)

'Fresh Air' is not really defined anywhere; rather just IAQ requirements for Outside Air. Ashrae Handbook: Fundamentals (2009) Section 16.2 Forced Air Distribution Systems states that "Air brought in intentionally from the environment is outdoor air (OA). Because outdoor air may need treatment to be acceptable for use in a building, it should not be called 'fresh air'" (It then doesn't state if fresh air can be created from outdoor by some process).

In my experience I have only seen 'fresh air' used (incorrectly ;)) as a synonym for 'outside air' hence my original class being equivalent.

Let me know you thoughts - I hope I haven't overlooked something obvious! Hopefully someone can provide some additional clarity on this...

JoelBender:
I concur with @wcrd.  In my experience "fresh" is only used in the context of describing the purpose of an "outside air fan" is to "bring in fresh air" as it is used to replace the "stale air" in the building, but the fan itself remains being classified as an outside air fan.

gtfierro:
Thanks for the context, and for taking the time to investigate the term! "Fresh Air Fan" was, i believe, a term we pulled out of one of the original 6 buildings that informed the original Brick release. This would track with the mistaken labeling that you've observed in practice. It sounds like we should prioritize "Outside Fan" instead. We don't yet have good practices for deprecating terms in future releases, but we should determine a practice or that.

How about this... we make "fresh air fan" a subclass of "outside fan" as proposed above, but we add an `owl:deprecated true` annotation to "fresh air fan"

In the code that should work out something like:

```python
{
  "Fresh_Air_Fan": {
   "tags": [ "Fresh", "etc" ],
  OWL.deprecated: Literal(true)
  }
}
```

You may need to import `OWL` from `brickschema.namespaces` at the top of the file

wcrd:
Thanks for the input @JoelBender.

@gtfierro I have made the changes requested. Agree that a practice for deprecating terms should be set (and how/how long they should be displayed in the ontology explorer on the brick website) - could be a good convo for the working groups

gtfierro:
We have added support for `brick:deprecation` annotations in v1.3, so I am merging this into the `v1.3.0` branch with some minor changes. `Fresh_Air_Fan` is being marked as deprecated, and to be replaced with `Outside_Fan` in the future. The deprecation model will raise warnings then `Fresh_Air_Fan` is used and can also automatically add the correct `Outside_Fan` type to the affected entities.

PR #302, opened 2021-07-28
Add Chilled Beams (and variants), Induction Units

wcrd:
Adding Chilled Beams including Passive and Active types, and Induction Units to the Terminal Units class.

I had a quick check of the issues and other pull requests and did not see any relevant items. Let me know if I have missed something.

@gtfierro The tags.py file (following CONTRIBUTING.md) is no longer in the repo so let me know if I need to also update some other file.


gtfierro:
@carlosduarteroa can you take a look at these definitions as well and let me know if these line up with the radiant system stuff you were doing?

gtfierro:
Thanks for submitting @wcrd! I'll take a look soon after @carlosduarteroa has had a chance to give his input as well (he was working on some radiant systems that might be relevant).

Thanks also for the pointer to the CONTRIBUTING.md being slightly out of date. You edited exactly the right files

carlosduarteroa:
Chilled beams are not radiant systems but I took a look at the chilled beam definitions and they look good to me. This PR can move forward.

gtfierro (approved):
Thanks for taking a look @carlosduarteroa!

Issue #305, opened 2021-08-05
Some applicable units seem to be missing?


shreyasnagare:
I'm not sure if this was intentional but there are a few applicableUnits missing for some quantities. 

Examples from Brick v1.2.0:
- `brick:Temperature qudt:applicableUnit unit:DEG_F .`  (this one's fixed now)
- `brick:Temperature qudt:applicableUnit unit:DEG_C .`  (missing?)

Question: Shouldn't units applicable to narrow concepts (say brick:Dry_Bulb_Temperature) be applicable to a broader concepts (brick:Temperature) as well?

Also, should there be more members in this SHACL list or is it meant to target only ambient temperatures for power outputs?
```ttl
brick:Temperature a brick:Quantity ;
    rdfs:label "Temperature" ;
    qudt:applicableUnit unit:DEG_F,
        unit:DEG_R,
        unit:K,
        unit:MilliDEG_C,
        unit:PlanckTemperature ;

brick:TemperatureShape a owl:Class,
        sh:NodeShape ;
    sh:property [ a sh:PropertyShape ;
            sh:datatype xsd:double ;
            sh:minCount 1 ;
            sh:path brick:value ],
        [ a sh:PropertyShape ;
            sh:in ( unit:DEG_C unit:DEG_F ) ;
            sh:minCount 1 ;
            sh:path brick:hasUnit ] .
```

gtfierro:
There's a few issues wrapped up in here so I will try to separate them.

The applicableUnit structure in Brick is based on QUDT and the equivalencies between Brick Quantities and QUDT QuantityKinds. In QUDT, CelsiusTemperature is _narrower_ than ThermodynamicTemperature, but Brick Temperature is only connected to ThermodynamicTemperature, hence we don't pick up the Celsius units. I'm not going to debate the choice that QUDT made, but I think the easiest fix is to equate Brick Temperature to CelsiusTemperature, and then pick up the _broader_ quantity kinds and their associated units automatically.

Your first question is about applicableUnits applying _up_ or _down_ the SKOS hierarchy. I'm less convinced this is a good idea. We should see what we can get out of the QUDT hierarchy traversal first

We could definitely add more units to the TemperatureShape.

PR #306, opened 2021-08-05
Expand ApplicableUnits annotations

gtfierro:
Updates QUDT to a more recent version; uses `skos;broader` hierarchy to find applicable units within QUDT. Also switching to QUDT.CelsiusTemperature for our Temperature quantities because this gets us `DEG_C` in addition to the usual units.


generate_brick.py
```diff
@@ -691,7 +691,7 @@ def add_definitions():
 res = G.query(
     """SELECT ?quantity ?qudtquant WHERE {
                 ?quantity rdf:type brick:Quantity .
-                ?quantity brick:hasQUDTReference ?qudtquant
+                ?quantity skos:broader*/brick:hasQUDTReference ?qudtquant
```
jbkoh, line 694: Assuming narrower concepts are analogous to subclasses, broader concepts should encompass narrower concepts, but not the other way around.

I propose to change this to `skos:narrower` and change `qk:CelsiusTemperature` to `qk:ThermodynamicTemperature` if we want to keep `qk:ThermodynamicTemperature` instead of `qk:Temperature`.

jbkoh: For example if I read the query, I don't expect to pull all the concepts other than `CelsiusTemperature` when I ask for `CeclsiusTemperature`. But if I query for `Temperature` (to avoid the `ThermodynamicsTempreature` issue), I expect `Celcius` to be included.

jbkoh (commented):
My comments will potentially demand changes, but let's resolve each item one by one.

bricksrc/quantities.py
```diff
@@ -636,10 +636,12 @@ def get_units(qudt_quantity):
         },
     },
     "Temperature": {
-        BRICK.hasQUDTReference: QUDTQK["ThermodynamicTemperature"],
+        # we use CelsiusTemperature for now because it includes DEG_C as well as
+        # the ThermodynamicTemperature units
+        BRICK.hasQUDTReference: QUDTQK["CelsiusTemperature"],
```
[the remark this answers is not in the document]
gtfierro: I have no idea why I didn't just use this before :). Great catch

generate_brick.py
gtfierro: After some investigation, I've found that QUDT is pretty consistent about having compatible units up and down the broader/narrower links. The real fix here is to ensure that `qk:Temperature` was used as the reference for `brick:Temperature`. Outside of that, I've noticed no change in the compiled model when we use the `skos:narrower` traversal. I think that we can avoid those hierarchy traversals for now, which should address the issues you are seeing in #305

gtfierro:
@shreyasnagare to address the other issue you noted in #305 I am deriving the units for some shapes from the Brick quantities themselves, so at least they will be internally consistent.

gtfierro:
@shreyasnagare @jbkoh would you be able to get feedback to me on this PR soon? I'd like to merge so that I can finish up the v1.2.1 release

shreyasnagare (approved):
Looks good to me!

jbkoh (approved):
LGTM!

PR #309, opened 2021-08-06
V1.2.1 release

gtfierro:
Preparing for v1.2.1 release of Brick!

- [x] remove `controls` relationship for now (@jbkoh ?)
- [X] fixed 'transmittence' typo


jbkoh:
Removing controls is proposed in #310

jbkoh:
A possible bug report:
`brick:Collection_Basin_Water_Heater` is a type of `brick:Equipment` but it has `tag:Collection` as a tag in different contexts afaik. `tag:Collection` means that the associated class is a `brick:Collection`. And in my understanding `brick:Collection` and `brick:Equipment` are disjoint in the current design though it's not explicitly defined. 

So, please add the disjoint relationship between `brick:Equipment` and `brick:Collection` which, I believe, will detect errors.

gtfierro:
@jbkoh those abbreviations are all relatively common for radiant systems. A quick search for any of those terms with a "radiant" keyword returns relevant results; as such, I think we are OK to keep them.

As to your other comment, this should have been caught by existing tests. I'll dig into why that is not failing, but I should note that this issue will be addressed by the SHACL implementation of tag/class inference, so even though it is a bug, it will be rectified in v1.3 and we can postpone our fix until then.

PR #310, opened 2021-08-08
remove controls

jbkoh:
`brick:controls` was originally defined as "A determines or affects the internal state of B" as in AppliedEnergy 2018 for describing relationships between points for abstracting control loops. Down the road, it has been used as different meanings by different users, while we haven't seen strong use cases of the original definition.

I propose to remove the relationship for now instead of deprecating, because people haven't used the relationship with the original definition anyway.

Another option is to just change the name to `brick:isReferencedBy` as proposed by Ambuj, but I'd still remove until we have better use cases (such as combining it with actual control functions.)


gtfierro (approved):
LGTM!

PR #313, opened 2021-08-24
Fix RDFS Labels

gtfierro:
I got the truth value incorrect (`G.objects` always returns a generator, which is True) for some of the duplicate RDFS label logic, resulting in missing labels for many concepts. This rectifies the error and ensures that labels are included on Brick concepts.


gtfierro:
Tested offline and all tests pass

PR #318, opened 2021-09-08
Add hasRepresentation

gtfierro:
Generalizes `brick:timeseries` to allow other kinds of digital representations of Points. So far I have done `brick:timeseries` and `brick:BACnetRepresentation`. The former is the existing data model; the latter is borrowed from the bacnet-shacl effort that @JoelBender is developing.

```ttl
@prefix bldg: <urn:example#> .
@prefix bacnet: <http://data.ashrae.org/bacnet/2020#> .
@prefix unit: <http://qudt.org/vocab/unit/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

bldg:sample-device a bacnet:BACnetDevice ;
    # TODO: need rest of address from bacnet-shacl
    bacnet:hasPort [ a bacnet:Port ] .

bldg:ts1 a brick:Zone_Air_Temperature_Sensor ;
    brick:hasUnit unit:DEG_C ;
    brick:BACnetRepresentation [
        bacnet:object-identifier "analog-value,5" ;
        bacnet:object-name "BLDG-Z410-ZATS" ;
        bacnet:objectOf bldg:sample-device ;
    ] ;
    brick:timeseries [
        brick:hasTimeseriesId "756e1623-914f-4415-9000-b1b10ce8f5c9" ;
        brick:storedAt "postgres://1.2.3.4:5432/mydata" ;
    ] ;
.
```

v1.3.0 will contain support for BACnet; other protocols will be added in future Brick releases

TODO:
- [ ] ~~build in initial support for the BACnet metadata into `py-brickschema`, integrate with BACpypes? (optional install)~~ (do in future point release)
- [x] include the BACnet SHACL defintiion as an extension that can be programmatically loaded via `py-brickschema`


JoelBender:
Cool!  :+1: I have to remember to put in `bacnet:objectOf` in my other samples, I've been only using `bacnet:hasObject`.  As a complete aside, I've been storing timeseries data in [Redis](https://oss.redis.com/redistimeseries/) so the URL changes to `redis://host:port/db`.

epaulson:
I wonder if we need to think about how to differentiate between multiple "hasRepresentation"s for a point. There's good reason to have this - in the example, you might have present-value data in the BACnet representation, while the historic data is in the timeseriesDB. Should we stick a property into the blank nodes to say something about what range that representation covers (or make the blank nodes named nodes so we can put a property in the point entity to say what range/liveliness is handled by which representation?)

I think this will matter for BACnet, because the BACnet object could be trended, and the trendlog is a separate BACnet object, so maybe we have two hasRepresentation properties, one for the object and one for the trendlog of the object? I suppose that depends on how @JoelBender is handling trendlogs in the BACnet shcacl

JoelBender:
You are correct, in BACnet-shacl the thing and an associated trend log object are distinct beasts, so they are separate objects.  The **Log_DeviceObjectProperty** property is a **BACnetDeviceObjectPropertyReference** so the two objects can be in separate devices.  With a snapshot of a building it's not too terrible to query the graph to see if a point is being trended someplace.

In **12.25.14 Log_Buffer** it is written:

> The log buffer is not network accessible except through the use of the ReadRange service, in order to avoid problems with
> record sequencing when segmentation is required.

BACnet-shacl has a description of what a BACnetLogRecord would look like if it was serialized as RDF, so it can be included:
```turtle
[] bacnet:timestamp [
            bacnet:date "2021-09-09"^^xsd:date ;
            bacnet:time "23:04:00.96"^^xsd:time
        ] ;
        bacnet:log-datum [ bacnet:real-value "75.3"^^xsd:float ]
    .
```
...but keeping very long lists of content like this is really not space/time friendly, so I have not included them in my "snapshots" of an object.  Lists and arrays in BACnet-shacl are rdf:first/rdf:rest/rdf:nil lists which are only slightly tolerable for device-address-binding lists, tag lists, and event schedules.

I remember hearing about a proprietary extension to SPARQL to query another SQL-like database/table that wasn't in the triple/quad store but I can't find the reference.  GraphQL is being presented as a way to collect both graph and relational database content together, essentially throwing the problem "over the wall."

I would continue with what you have, and leave the problem of getting data from BACnet into the timeseries archive (via polling, COV notification, or copying data from a Trend Log Object) an application layer problem.

epaulson:
How is a BACnetDeviceObjectPropertyReference represented in RDF? 

If the Log_DeviceObjectProperty is on the Trend Log object, that means in order to find a trend log, you have to find the Brick sensor you want, find the BACnet representation object (which is a blank node in the example above but is of type BRICK.BACnetReference), and then find TrendLogs that have a Log_DeviceObjectProperty that points back to that BACnet object. 

I'm worried join between the Log_DeviceObjectProperty in the TrendLog back to the BRICK.BACnetReference object. 

Or, alternatively, we don't try to join back through the Log_DeviceObjectProperty and just have another Brick:hasRepresentation property in the bldg:ts1 entity that points at the bacnet:BACNetTrendLog entity (which we may need, because in Michael Newman's book it's mentioned that for a TrendLog, the Log_DeviceObjectProperty is optional, if the device is tracking something that doesn't correspond to another BACnet object?) 

(And then there's TrendLogMultiple, too)

For reading data, I think we're all on the same page - the Brick model has the info to find a point's value for BACnet, but we expect the application to have a BACnet library linked into it (or whatever the equivalent of linking is in your favorite language) and then use that BACnet library to go and fetch data. (But we should talk about what it would look like if we're updating 'present values' in the Brick graph, even though that's usually probably a bad idea)

JoelBender:
> How is a BACnetDeviceObjectPropertyReference represented in RDF?
```turtle
[] bacnet:device-identifier "device,123" ;
    bacnet:object-identifier "analog-value,1" ;
    bacnet:property-array-index "3"^^xsd:nonNegativeInteger ;
    bacnet:property-identifier "present-value" .
```
The property array index is only used when the property is an array, and if it is omitted then the entire array is being referenced.  The device identifier is also optional and used when the object being referenced is in a different device, for example, if an Event Enrollment Object references a Analog Value Object in the same device then the `device-identifier` would be omitted.

Note that the device identifier is an object identifier as a string literal, not an RDF link to the Device Object instance in a device or to the Device itself.  But that's fine, every device has one and only one Device Object so `?device bacnet:hasObject [ bacnet:object-identifier "device,123" ]` has only one answer in a BACnet internetwork.

> I'm worried join between the Log_DeviceObjectProperty in the TrendLog back to the BRICK.BACnetReference object.

I think there are enough data pieces in these components to make that "join" but I don't think that you would want to do it so often that it would justify "materializing" the relationship like giving it a special predicate.  There are two questions I think you are asking (and I've been calling these DOPRs 'cause it takes to long to type):

1. Is this thing that has a DOPR, being trended by some BACnet device?  If it is, how do I get to the trend data?
2. Is this DOPR that is in my BACnet network referenced by something in my Brick graph?  If it is, what kind of thing is it?

> ...if the device is tracking something that doesn't correspond to another BACnet object?

Correct, it's footnote 8, _This property is required if, and shall be present only if, the monitored property is a BACnet property_.  You can have trend logs of all kinds of other "properties" in your device like performance metrics that do not have a corresponding BACnet object/property to read them.

I hadn't considered the idea that you would want to map any trend data in a BACnet device to something in Brick.  A DOPR is good for ReadProperty[Multiple], but something like:
```turtle
bldg:ts1 a brick:Zone_Air_Temperature_Sensor ;
    brick:BACnetTrend [
        bacnet:device-identifier "device,123" ;
        bacnet:object-identifier "trend-log,6" ;
    ]
.
```
I haven't seen any Trend Log Multiple objects, so this is a wild stab in the dark...
```turtle
bldg:ts1 a brick:Zone_Air_Temperature_Sensor ;
    brick:BACnetTrend [
        bacnet:device-identifier "device,123" ;
        bacnet:object-identifier "trend-log-multiple,2" ;
        bacnet:property-identifier "log-device-object-property" ; # optional?
        bacnet:property-array-index "3"^^xsd:nonNegativeInteger ;
    ]
.
```
Maybe that can act as a "source reference" for the data that is otherwise in the contents referenced by the `brick:timeseries` predicate, or perhaps that _is_ the contents of the trend reference!  **_Muahaha..._**
```turtle
bldg:ts1 a brick:Zone_Air_Temperature_Sensor ;
    brick:timeseries [
        a brick:BACnetTrendReference ;
        bacnet:device-identifier "device,123" ;
        bacnet:object-identifier "trend-log-multiple,2" ;
        bacnet:property-identifier "log-device-object-property" ; # optional?
        bacnet:property-array-index "3"^^xsd:nonNegativeInteger ;
    ]
.
```

JoelBender:
> But we should talk about what it would look like if we're updating 'present values' in the Brick graph, even though that's usually probably a bad idea.

This is dancing so close together with 223 that they'd have to be separated in middle school:
```turtle
bldg:ts1 a brick:Zone_Air_Temperature ;
    a s223:ObservableQuantifiableProperty ;
    s223:hasSimpleValue "72.0"^^xsd:float ;
.
```
And maybe overkill for Brick, designed for the odd case where you want to have the same value with multiple units or coupled with a timestamp, going back to the log record concept:
```turtle
bldg:ts1 a brick:Zone_Air_Temperature ;
    s223:hasValue [
        a s223:Value ;
        s223:hasSimpleValue "72.0"^^xsd:float ;
        s223:hasUnits qudt:DEG_F .
        s223:hasTimestamp "2021-09-10T23:05:02.957825"^^xsd:dateTime .
    ]
```

gtfierro:
From Office Hours discussion:

- should include time ranges on representations that have historical data:
    - have start/stop time; stop time is optional to indicate that data is still being inserted
    - start time could also be optional?
- need `current-value: true` property? Or is this assumed you can get current value if there are no annotations about historical data?
- need `preferred representation` property?
- better name than `hasRepresentation`?:
      - `hasDataSource`: suggests read-only
- `BACnetObjectRepresentation` and `BACnetTrendLogRepresentation` as two separate objects, using the structures proposed above:
    - trend log requires additional properties above `BACnetObjectRepr`: needs to distinguish between trend log and trend log multiple (look at `object-identifier` field for the object type?)
- the BACnet scan that creates the RDF graph would take care of stitching together ObjectReprs and TrendLogReprs for the same entity:
   - could be part of the scan or more likely a post-processing pass that is specific to Brick
- there are trend logs that are not related to BACnet objects. Trend logs can also represent the data from a Modbus device which itself is not in BAcnet

Issue #315, opened 2021-08-31
New load/end-use-device EntityProperties requests


michaelpoplawskipnnl:
Current EntityProperties contains ratedPowerOutput and measuredPowerOutput for devices.
While these are useful and appropriate properties for power supplies/converters, they are not useful or appropriate for loads.
Loads are typically characterized by their input properties.

Examples:
Rated (nominal) Input Power, important for supply, wire, switching infrastructure sizing
Rated Input Voltage(s) e.g., 120 VAC, 120-277VAC, important for supply connection
Rated Maximum Input Current, important for wire, switching infrastructure sizing, component lifetime analysis, fault detection, etc.

Suggest adding new Input properties that could be defined for loads of various types e.g., lighting devices, plug-loads

gtfierro:
@michaelpoplawskipnnl take a look at these and let me know what you think of the definitions: https://github.com/BrickSchema/Brick/pull/319

PR #319, opened 2021-09-10
Add end-use device entity properties 

gtfierro:
Addresses the asks in https://github.com/BrickSchema/Brick/issues/315 , resulting in the following new entity property definitions:

- `brick:ratedPowerInput`
- `brick:measuredPowerInput`
- `brick:ratedVoltageInput`
- `brick:ratedMaximumCurrentInput`


gtfierro:
Changed to support the cross product of rated/measured input/output min/max current/voltage entity properties

PR #320, opened 2021-09-13
Reduce automated test time

gtfierro:
Remove hierarchy inference test from usual suite of tests to reduce feedback time on automated tests


gtfierro:
Down to 30 minutes from > 90 minutes :)

PR #322, opened 2021-09-13
Labels on properties and entity properties

jbulow:
When creating tools with a UI for creating brick data it is quite conventient to have rdfs:labels on most things. Is this interesting to include directly in the main source for the brick ontology?


jbulow:
I will add more labels if there is interest in this PR.

gtfierro:
Yes! This is definitely needed; thanks so much for doing what you've done so far. If there are other terms missing `rdfs:label` it would be fantastic to add those as part of this PR. Let me know how I can help

jbulow:
I've added labels to all entity properties and properties now.

gtfierro:
@jbulow can you merge the latest from master and push? I've made some changes to the unit tests that makes them run faster and give the correct result more reliably. There is something going on with the SHACL shapes that causes them to (incorrectly) fail a test from time to time

martinwebrant:
@gtfierro interested in maintaining translations in this repo as well?

jbulow:
All properties and entity properties now covered in this PR. Anything else to consider in this PR?

gtfierro:
@martinwebrant It would be great to have translations into different languages in Brick. I don't have time to produce that myself, but please let me know how I can support translation efforts. I was talking a little with @jbulow today in the Ontology Working Group and I mentioned having lookup tables in the `bricksrc/` folder which could store the mapping from Brick classes to different labels for various languages. I'd be happy to set up that infrastructure in the repo, but let me know if you have other ideas. I'll make a note to begin adding `@en` language tags to existing literals in Brick

gtfierro:
Thanks for your great work on the PR, @jbulow!

PR #326, opened 2021-09-20
Update files generated from source

jbulow:


gtfierro (approved):
We can definitely reinvestigate which files should be checked in to the Brick repo at a future date, but this is a great fix to have for now! Thanks for submitting this

PR #327, opened 2021-09-20
Ignore generated files

jbulow:


.gitignore
```diff
@@ -184,3 +184,6 @@ com_crashlytics_export_strings.xml
 crashlytics.properties
 crashlytics-build.properties
 fabric.properties
+/Brick+extensions.ttl
+/Brick.ttl
+/test.ttl
```
gtfierro, line 189: Can we remove `test.ttl` from this `.gitignore` file and instead remove the corresponding line from `tests/test_hierarchy_inference.py`? I think we should just stop generating the `test.ttl` file.

None of the files should require the leading `/`; listing, for example, just `Brick.ttl` in the `.gitignore` file is fine.

jbulow:
All changes should be addressed now.

gtfierro (approved):

gtfierro:
Thanks!

PR #328, opened 2021-09-26
add shading concepts

jbkoh:
### New shading concepts
- Shade
- Shade_Array

- Automatic_Tint_Window
- Tint_Command
- Tint_Status
- Illuminance_Setpoint

### A things to discuss further
- A `Shade` is also a specific type of `Equipment` while `Shading_Equipment` is a general concept for any type of shading equipment. While the names are distinguished by `Shading` vs `Shade`, their tags are the same. Currently I added `tag:TODO` to `Shade` for now. ***This needs to be resolved.***
    - A possible solution could be introducing a tag `Shading` though it feels redundant. If there is no other way, I will introduce `Shading` tag.
- A usability issue withf `Collection` in general: Currently, I currently modeled `Shade_Array` as a type of `Collection` like `PVArray`. A collection won't be useful until its members are defined. However, modeling a Shade is typically very difficult, especially through control systems given that only a controller for the entire shade array will be visible. Calling `Shade_Array` a type of `Equipment` will be much more usable while `Shades` can be its parts like other Equipment's `hasPart` relationship.


jbkoh:
I updated this one but the test fails for the other things in the 1.3.0. I guess it's ready for merge assuming the 1.3.0 will fix the issues.

@gtfierro  check this and please merge it into the 1.3.0 branch.

gtfierro (approved):

gtfierro:
tests look good to me!

PR #329, opened 2021-09-26
add missing points

jbkoh:
adding
- `Supply_Air_Temperature_Setpoint_Limit`, `Max_Supply_Air_Temperature_Setpoint_Limit`, and `Min_Supply_Air_Temperature_Setpoint_Limit` for their equivalence of `Discharge`.
- `Unoccupied_Humidity_Setpoint` and `Occupied_Humidity_Setpoint`
- `Valve_Status`
- Fix existing typos from `Cooling` to `Cool`.

removing 
- `Air_Humidity_Setpoint`: By definition, Humidity is Air humidity: "(General Physics) a measure of the amount of moisture in the air". We also don't have "Air_Humidity_Sensor" so no need for having the class.


gtfierro (changes requested):
This PR needs definitions for the new concepts, but otherwise the tests pass locally and I think these are good changes

jbkoh:
added defs.

gtfierro:
Tests are failing due to two definitions w/o corresponding classes:
- `Occupied_Cooling_Temperature_Setpoint`
- `Occupied_Heating_Temperature_Setpoint`

PR #331, opened 2021-09-26
add miscellaneous alarms

jbkoh:
Adding misc alarms.

@gtfierro I'm adding `Sensor_Failure_Alarm` indicating whether a specific sensing module is failing or not. For doing so, I should be able to add tags `Sensor` and `Alarm`, which will classify this as a `Sensor` as well. We had similar issues in different places, but I can't recall how we resolved it. Do you have any idea?

### DIFF
--- a/bricksrc/alarm.py
+++ b/bricksrc/alarm.py
@@ -18,6 +18,11 @@ alarm_definitions = {
                             },
                             "Low_Air_Flow_Alarm": {
                                 "tags": [TAG.Point, TAG.Air, TAG.Alarm, TAG.Flow, TAG.Low],
+                                "subclasses": {
+                                    "Low_Discharge_Air_Flow_Alarm": {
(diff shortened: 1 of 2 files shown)


gtfierro:
This issue is solved with the new SHACL-based tag/class resolution mechanisms that will land in v1.3. Don't worry about it at this point in time! Just indicate the tags in the `.py` files as you normally would and the `generate_brick.py` script for v1.3 will take care of the rest.

gtfierro (changes requested):
These all need corresponding definitions in the `definitions.csv` file, but otherwise this looks good

gtfierro (changes requested):
Tests are failing because there is a definition for `Low_Discharge_Air_Flow_Alarm` that does not have a corresponding class. I think you can either add the `Low_Discharge_Air_Flow_Alarm` class, or just remove the extra definition --- your call!

gtfierro (approved):

PR #333, opened 2021-09-26
restructure valve hierarchy and add TMV

jbkoh:
Valves are not specific to HVAC, so moving `Valve` out of `HVAC_Equipment` subtree. Introducing `HVAC_Valve` for replacing the existing hierarchy in case it's useful, but I'm happy to remove that as well.

Then, I add `Thermostatic Mixing Valve`.


gtfierro (approved):
I like the change! I'm running the tests locally to see if they are actually failing or if it is just GH actions hitting issues again

generate_brick.py
```diff
@@ -687,6 +688,7 @@ def add_definitions():
 define_classes(equipment_subclasses, BRICK.Equipment)
 define_classes(collection_classes, BRICK.Collection)
 define_classes(hvac_subclasses, BRICK.HVAC_Equipment)
+define_classes(hvac_valve_subclasses, BRICK.HVAC_Valve)
 define_classes(valve_subclasses, BRICK.Valve)
```
gtfierro, line 692: Now that `Valve` is the top level class in the `valve_subclasses` dictionary, this line should be:

```python
define_classes(valve_subclasses, BRICK.HVAC_Equipment)
```

otherwise there is a loop where `Valve` is a subclass of `Valve`

gtfierro (changes requested):

gtfierro:
Can you add a minor quality of life change to this PR? At the bottom of `tests/test_subclass_hierarchy.py`, can you replace the last assertion as follows?

```diff
-    assert len(loops) == 0, "Loops found in the class hierarchy!"
+    assert len(loops) == 0, f"Loops found in the class hierarchy! {loops}"
```

This will print out the loops that were detected and facilitate future debugging

gtfierro (approved):

PR #335, opened 2021-09-26
add misc concepts

jbkoh:
- Heat Sink temperature Sensor
- Atrium


jbkoh:
@gtfierro I observe this test failure line: https://github.com/BrickSchema/Brick/pull/335/checks?check_run_id=3712132104#step:8:234
but only in Python 3.7.9 test. I wasn't able to regenerate this issue locally in my Python 3.8.10 too. Is it a known issue?

gtfierro:
It occasionally fails on GH actions but I can't get the failure to reproduce locally. Does it fail locally for you?

jbkoh:
I didn't test it on the exact Python version. I will try that way later today.

gtfierro:
By and large I haven't found that the exact Python version really matters. GitHub actions just occasionally fails tests. There's probably a good reason for it, but I can't think of what it might be and I haven't had the time to really dig in to it. If the tests pass consistently on your machine, then I'm generally okay with just merging the PR even if the tests don't pass online. Even then, if you run it enough times it usually works eventually :)


From: Jason B. Koh ***@***.***>
Sent: Monday, September 27, 2021 12:31:47 PM
Cc: Gabriel Fierro ***@***.***>; Mention ***@***.***>
Subject: [External] Re: [BrickSchema/Brick] add misc concepts (#335)


I didn't test it on the exact Python version. I will try that way later today.

jbkoh:
sure, if it works, I'm good.

gtfierro (approved):

PR #336, opened 2021-09-26
more fire alarm concepts

jbkoh:
Various fire alarm devices are added.

1. However, the current PR needs to be updated for handling Alarm-related Equipment and Points. @gtfierro  what was your solution on this before?
2. `Fire_Control_Panel` seems not commonly adopted. Is there any reference to this concept? I do see `Fire_Alarm_Control_Panel` widely defined (e.g., https://en.wikipedia.org/wiki/Fire_alarm_control_panel). I currently maintained both in the current PR, but I'd rather replace `Fire_Control_Panel` with `Fire_Alarm_Control_Panel` @epaulson  what do you think?

### DIFF
--- a/bricksrc/definitions.csv
+++ b/bricksrc/definitions.csv
@@ -398,7 +398,7 @@ https://brickschema.org/schema/Brick#Filter_Differential_Pressure_Sensor,Measure
 https://brickschema.org/schema/Brick#Filter_Reset_Command,,
 https://brickschema.org/schema/Brick#Filter_Status,Indicates if a filter needs to be replaced,
 https://brickschema.org/schema/Brick#Final_Filter,"The last, high-efficiency filter installed in a sequence to remove the finest particulates from the substance being filtered",
-https://brickschema.org/schema/Brick#Fire_Alarm_Panel,Fire alarm panel is the controlling component of a fire alarm system.,https://en.wikipedia.org/wiki/Fire_alarm_control_panel
+https://brickschema.org/schema/Brick#Fire_Alarm_Control_Panel,Fire alarm panel is the controlling component of a fire alarm system.,https://en.wikipedia.org/wiki/Fire_alarm_control_panel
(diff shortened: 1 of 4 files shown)


gtfierro:
> However, the current PR needs to be updated for handling Alarm-related Equipment and Points. @gtfierro what was your solution on this before?

I'm essentially punting on this issue until we get the SHACL stuff merged in v1.3. I recall we had some conversation about what this looks like but let me know if you have further questions

> Fire_Control_Panel seems not commonly adopted. Is there any reference to this concept? I do see Fire_Alarm_Control_Panel widely defined (e.g., https://en.wikipedia.org/wiki/Fire_alarm_control_panel). I currently maintained both in the current PR, but I'd rather replace Fire_Control_Panel with Fire_Alarm_Control_Panel @epaulson what do you think?

This reminds me of the deprecation discussion in https://github.com/BrickSchema/Brick/pull/298. If `Fire_Control_Panel` is merely an alias for `Fire_Alarm_Control_Panel`, then we can just make them equivalent. If `Fire_Alarm_Control_Panel` is indeed a more correct label for `Fire_Control_Panel` (similar to discussion in #298), then we can mark the latter as deprecated in favor of the former

gtfierro:
@jbkoh I changed the base for this PR to be v1.3.0 so it can take advantage of the SHACL-based implementation.

I think we can replace the Fire Control Panel concept as you mentioned as well.

Are we almost good to merge?

jbkoh:
We are good. I removed `Control` tag from `Fire_Alarm_Panel`. I wanted to check the test passes but it seems the Github Action is disabled somehow. Any idea @gtfierro ?

I would like to deprecate `Fire_Control_Panel`. I can either wait for the deprecation PR to be merged first or merge this one and then mark it in the 1.3.0 branch.

epaulson:
I thought we were going with 'Fire_Alarm_Control_Panel' as the full name, but this has 'Fire_Alarm_Panel'? 

My understanding is that at the company, we use 'Fire_Control_Panel' as an (very very common but informal) name for 'Fire_Alarm_Control_Panel' - see https://www.johnsoncontrols.com/fire-detection/control-panels 

I really don't want us to drop 'control'.

jbkoh:
@epaulson  you are right. I just fixed it. Thanks for noticing it!

gtfierro (approved):

PR #339, opened 2021-10-01
Added heating, cooling enable commands

carlosduarteroa:
Enable commands for equipment that show if heating or cooling is enabled


gtfierro:
Thanks for the clarification! Could you include in the definitions of these points something to indicate to the user that for actual heating/cooling activity they should look at Heating/Cooling Command points? Then we can merge this

carlosduarteroa:
I have updated the definitions to include your suggestions.

PR #341, opened 2021-10-06
SHACL-based Brick Implementation

gtfierro:
**Aspects of this may change drastically!**

Replaces OWL 2 RL reasoning requirements with SHACL-based implementations


gtfierro:
Fixing these tests requires https://github.com/RDFLib/pySHACL/issues/96 to be merged

PR #349, opened 2021-11-03
Add quantity kinds, substances as explicit annotations on all Point classes

gtfierro:
Most Brick Point classes should now have `brick:hasSubstance` and `brick:hasQuantity` annotations to assist in finding related points and for describing those classes more consistently than the previous `brick:measures` model (which is now completely removed).

Example:

```turtle
  brick:Air_Temperature_Setpoint a owl:Class, sh:NodeShape ;
      rdfs:label "Air Temperature Setpoint" ;
      rdfs:subClassOf brick:Temperature_Setpoint ;
      brick:hasAssociatedTag tag:Air,
          tag:Point,
          tag:Setpoint,
          tag:Temperature ;
      brick:hasQuantity brick:Temperature ;
      brick:hasSubstance brick:Air .
```

These annotations are associated with the Brick class, not the instances.

One use case is one can query the substances and/or quantities to find alternative classes offering similar kinds of data.

```sparql
PREFIX brick: <https://brickschema.org/schema/Brick#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
SELECT * WHERE {
  brick:Damper_Position_Sensor rdfs:subClassOf*/brick:hasQuantity ?quant .
  ?class brick:hasQuantity ?quant
}
```
gives as results

class | quant
-- | --
brick:Damper_Position_Command | brick:Position
brick:Damper_Position_Setpoint | brick:Position
brick:Position_Command | brick:Position
brick:Position_Sensor | brick:Position


gtfierro:
```sparql
PREFIX brick: <https://brickschema.org/schema/Brick#>
PREFIX rdfs: <http://www.w3.org/2000/01/rdf-schema#>
SELECT * WHERE {
  brick:Damper_Position_Sensor rdfs:subClassOf*/brick:hasQuantity ?quant .
  ?class brick:hasQuantity ?quant .
  ?class rdfs:subClassOf* brick:Status .
}
```

Example for restricting returned classes to only Statuses

bricksrc/status.py
```diff
@@ -4,6 +4,9 @@
     "Status": {
         "tags": [TAG.Point, TAG.Status],
         "subclasses": {
+            "Damper_Position_Status": {
+                BRICK.hasQuantity: BRICK.Position,
```
jbkoh, line 8: Shouldn't this be `Damper_Position`?

jbkoh:
`Electrical_Power_Sensor` is missing `hasQuantityKind`. Also the name doesn't match to `Electric_Power`, and I'd suggest to stick to one of them.

bricksrc/status.py
gtfierro: Do we want equipment-flavored quantities? I don't believe we have any others, e.g. "fan speed"

gtfierro:
@jbkoh I will leave equipment-flavored quantities off for now; the recent changes to the PR should fix the unit association and I have standardized terminology on "Electric" rather than "Electrical"

PR #350, opened 2021-11-03
Deprecation support

gtfierro:
This PR adds properties, a shape and a rule for handling the deprecation of Brick classes.

To mark a Brick class as deprecated, the `brick:deprecation` property is used. The value of this property is an object with a `brick:deprecatednInVersion` property (string-valued with the Brick version which deprecated the concept) and a `brick:deprecationMitigation` property (string-valued with directions for how to handle the deprecation).

Example:

```turtle
brick:Fresh_Air_Fan brick:deprecation [
    brick:deprecatedInVersion "1.3.0" ;
    brick:deprecationMitigation "Replace with brick:Outside_Fan" ;
] .
```

After running a SHACL-AF reasoner (`graph.expand("shacl")` in the [brickschema](https://github.com/BrickSchema/py-brickschema) package), the `owl:deprecated` property is added to the class with a value of `true`

(The irony of naming a PR "deprecation support" is not lost on me)


gtfierro:
@wcrd @epaulson let me know what you think

jbkoh:
I support all the proposed properties. I'd like to discuss a few more things

- Do we intend to actually remove them at some point or do we keep them forever?
  - If we keep them forever, what's the typical usage flow for end users? Should they commonly add `?class owl:deprecated false` for any SPARQL not to accidentally use deprecated concepts?
  - A solution could be we publish deprecated concepts in a separate file, but it would also harm the existing users.
- Similar to the other discussion like `seeAlso` (ignore the exact term), it'd be great if we materialize the mitigation methods in the schema as well.

gtfierro:
We can probably remove them from the class hierarchy and just keep the deprecation annotation in the model. To avoid deprecated concepts, I think it might be enough for SPARQL queries to include `FILTER NOT EXISTS { ?class owl:deprecated true }`, but we could infer the `deprecated = false` annotation if we needed.

Can you be more specific about what you mean by materializing the mitigation methods? In the case where it is just a simple substitution, we could make a rule for that, but in the general case there may be more complex mitigation that is harder to express as a SPARQL CONSTRUCT query

epaulson:
I'm OK with eventually removing things, but I wouldn't want to start doing that until we have more people putting a version annotation into their instance TTL files as best practice.

jbkoh:
After re-reading the proposal, I take back some of my previous comments. I'd say we should not introduce indirection unless it's really useful (either the concept is really a separate concept like measurement types or the indirected concept can have standalone relationships with others). In this case, `brick:deprecation` adds certain properties to the class and the separate grouping doesn't add much value, imho. The indirection would degrade the usability and complicate the data model.

I propose to have the proposed properties as direct properties for classes as
```turtle
brick:Fresh_Air_Fan  brick:deprecatedInVersion "1.3.0" ;
    brick:deprecationMitigation "Replace with brick:Outside_Fan".
```

jbkoh:
> Can you be more specific about what you mean by materializing the mitigation methods? In the case where it is just a simple substitution, we could make a rule for that, but in the general case there may be more complex mitigation that is harder to express as a SPARQL CONSTRUCT query 

Yeah, I understand a simple lookup is not sufficient for all the cases. I was proposing to define the lookup where it's simple enough as a triple. If we would like to avoid the situation some rules are provided while others are not, I'm fine with just the proposed textual description.

gtfierro:
I like the indirection because it makes explicit what information is related to the deprecation, rather than relying on the "deprecation*" prefix on the properties. It is also consistent with the Entity Property construct, which I like.

I've extended the deprecation shape to include mitigation rules and added an example deprecation implementation [here](https://github.com/BrickSchema/Brick/blob/9695f75ab0d1aabac519878d00e30f7fd57cc87b/bricksrc/deprecations.ttl). The rules will be executed automatically and the mitigations will be added to the graph; however, I don't think there is a way we can have SHACL *remove* information from the graph. In any case, the SHACL report will indicate  a Warning as long as the deprecated class is still used.

Also worth pointing out that the above implementation has only been tested for classes; we can make it work for properties as well but it would require a little adjusting.

jbkoh:
Instance deprecation would be out of scope for Brick and I'm fine with testing only over classes.

gtfierro:
@jbkoh I've changed the implementation to remove the intermediate object:

```ttl
brick:Fresh_Air_Fan 
    brick:deprecatedInVersion "1.3.0" ;
    brick:deprecationMitigationMessage "Fresh Air Fan is deprecated in favor of Outside Fan because the latter is a more accurate representation" ;
    brick:deprecationMigitationRule [
        a sh:NodeShape ;
        sh:rule [
            a sh:SPARQLRule ;
            sh:construct """
            CONSTRUCT {
              $this rdf:type brick:Outside_Fan .
            } WHERE {
              $this rdf:type brick:Fresh_Air_Fan .
            }""" ;
            sh:prefixes rdf:, brick: ;
        ] ;
        sh:targetClass brick:Fresh_Air_Fan ;
    ] ;
.
```

I still don't like this design as much because it deviates from the EntityProperty style, but am willing to concede if that's a blocker for you. If you're ok with the implementation in this PR let me know so I can merge this in

jbkoh:
@gtfierro It's a different style, and I agree you have a point too. I just wanted to reduce a mental leap for non-graph-experts to understand deprecation as an independent node unless the indirection is necessary. But at this point, my concern is pretty hypothetical, so I'd like to respect the actual contributor's opinion more. You may roll it back to what you originally proposed.

gtfierro:
I appreciate your flexibility! I've reverted the changes and will merge this in to the 1.3 branch soon. One more thing checked off!

PR #351, opened 2021-11-09
Fix: deterministic serialization

### LINKED ISSUE #324: Generation of ontology is not 100% deterministic
The same brick source generate slightly different `Brick.ttl` files on each run. The order of shape constraints are randomized. This is a minor problem but causes some confusion when inspecting the diff on that has been updated between revisions.

jbulow:
Avoid randomness by using explicit enumeration of blank node identities.

Another attempt to fix #324

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,10 +3,10 @@
 Brick.ttl: bricksrc/*.py bricksrc/*.ttl bricksrc/definitions.csv generate_brick.py
 	mkdir -p extensions
 	python generate_brick.py
-	cd shacl && python generate_shacl.py
+	cd shacl && python -c 'import sys; sys.path.append("../bricksrc")' generate_shacl.py
 
 shacl/BrickShape.ttl: bricksrc/*.py generate_brick.py shacl/generate_shacl.py
-	cd shacl && python generate_shacl.py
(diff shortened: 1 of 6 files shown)


jbulow:
The tests fails because I had to put the import statement before the `sys.path.append` statement in `shacl/generate_shacl.py` to please the pre-commit hook. I.e there is a logical dead-lock here.

jbulow:
Fixed flake8 issue by setting the sys.path outside of the source file.

gtfierro:
Rather than putting the path extension in the invocation, can you use the `noqa` pragma to fix the linting error instead? This is what we use elsewhere in the repository: (see [here](https://github.com/BrickSchema/Brick/blob/master/tests/test_class_structure.py#L5-L6))

I think the approach is clever, but I wonder if we can avoid depending on the order of calling `BNode` by using deterministic IDs that are generated from context? See [here](https://github.com/BrickSchema/Brick/blob/master/generate_brick.py#L155) for an example

jbulow:
I tried to make context based IDs but the code got really complex and deeply entangled with the enclosing context. I failed in the end as I couldn't identify what the exact context was and some BNodes ended up with non-unique ID:s.

I will submit that effort as a separate PR. Maybe you can spot the obvious errors in it.

PR #353, opened 2021-11-11
A third attempt to get deterministic builds

jbulow:
Currently broken.

### DIFF
--- a/shacl/generate_shacl.py
+++ b/shacl/generate_shacl.py
@@ -37,7 +37,7 @@ def addDomainShape(propertyName, expectedType):
 # Make shape for expectedRange property
 def addRangeShape(propertyName, expectedType):
     rangeShapeDict[propertyName] = expectedType
-    sh_prop = BNode()
+    sh_prop = BNode(f"RangeShape_{propertyName}_{expectedType}")
     shapename = f"{propertyName}RangeShape"
     G.add((BSH[shapename], SH["property"], sh_prop))
     G.add((BSH[shapename], A, SH.NodeShape))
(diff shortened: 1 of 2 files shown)


gtfierro:
@jbulow I fixed the errors you were seeing above; there were a couple typos in the f-strings and the parameters needed to be cleaned so that they could be serialized into the blank node id (not that these are visible outside the graph). Some initial testing shows that running this multiple times generates the same Brick model!

I can't incorporate the changes directly to your branch unless I am doing them as part of the merge, so I put the patch here so you can take a look, apply it locally and test, and then add to this PR if you like the changes: https://gist.github.com/gtfierro/1e978d684a4f583c9d26437a556f42e2

jbulow:
Anything more to do before this commit is ready for commit?

shacl/generate_shacl.py
```diff
@@ -37,7 +37,9 @@ def addDomainShape(propertyName, expectedType):
 # Make shape for expectedRange property
 def addRangeShape(propertyName, expectedType):
     rangeShapeDict[propertyName] = expectedType
-    sh_prop = BNode()
+    sh_prop = BNode(
+        f"RangeShape_{propertyName.split('#')[-1]}_{expectedType.split('#')[-1]}"
```
gtfierro, line 41: On line 41, you should use the convenience method `bn` rather than the whole `.split` construction

gtfierro (changes requested):
Hi @jbulow sorry for the delay --- just getting back from a US holiday. There's one small change to make here and then I think it is good to merge

gtfierro (approved):

gtfierro:
Thanks!

PR #354, opened 2021-11-12
test the applicable units are correct throughout the hierarchy

jbkoh:
Please consider including this in the PR. `Speed` doesn't have enough units like `Wind_Speed` which should be fixed on one of 1) our quantity generation logic, 2) our version of units.ttl, or 3) QUDT.

Current test resut:
> FAILED tests/test_units.py::test_points_hierarchy_units - AssertionError: assert not {'Speed_Sensor': {'Wind_Speed_Sensor': {'FT-PER-HR', 'FT-PER-SEC', 'KiloM-PER-HR', 'KiloM-PER-SEC', 'M-PER-HR', 'MI-PER-HR'}}}
, meaning that the listed units are observed for `Wind_Speed` but not for `Speed`.

### DIFF
--- a/tests/test_units.py
+++ b/tests/test_units.py
@@ -128,3 +128,30 @@ def test_all_quantities_have_units():
         warnings.warn(
             f"The following quantities do not have associated units: {quantities_without_units}"
         )
+
+def test_points_hierarchy_units():
+    g = brickschema.graph.Graph()
+    g.load_file("Brick.ttl")
+    qstr = """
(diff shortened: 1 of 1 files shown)


gtfierro:
Thanks for the test! The fix for this is to make sure that the units are inherited up the hierarchy  so that parent quantities have the union of the units for any narrower/more specific quantities

gtfierro:
I will merge this in shortly

PR #357, opened 2021-11-29
Add IFC reference

gtfierro:
See the example file for what this looks like. All attribute/etc names are up for debate

Relationship is `brick:IFCRepresentation`, mirroring `brick:BACnetRepresentation`. The object itself has two mandatory properties, `ifc:globalID` (for the entity) and `ifc:projectID` (for the project defining the entity) ; and two optional properties `ifc:name` a human-readable label of the entity in the IFC project, and `ifc:fileLocation` which points to the location of the IFC file.


epaulson:
Do we want to try to pull in some of the buildingSMART classes here? I think I've got this right, looking at https://www.mdpi.com/2076-3417/10/24/8794/pdf (also these: https://www.researchgate.net/publication/355425860_SHACL_is_for_LBD_what_mvdXML_is_for_IFC  and https://standards.buildingsmart.org/IFC/DEV/IFC4_1/OWL/index.html ) 

```ttl
@prefix ifc: <https://standards.buildingsmart.org/IFC/something-version-not-sure-which>
ex:blg1 a ifc:IfcBuilding
ex:blg1 ifc:globalId_IfcRoot ex:guid123

ex:guid123 a ifc:IfcGloballyUniqueId
ex:guid123 express:hasString "guid123"
```

I don't know that I'd want to say that ex:blg1 is also a brick:building, so I think it's safer to keep the brick:hasIFCRepresentation and maybe point it at ex:guid123?

I'm also fine with having our own shorter version and I don't want to add unnecessary complexity. To go between the two systems one side is going to have to know a bit about the other :) - also I think we'll want something similar for SVG, so keeping the hasRepresentation family of properties seems fine.

gtfierro:
I like the idea of keeping the `representation` objects in Brick fairly uniform in their construction, but if there are small things we can do to increase interop then I am all for that. One complication with using the IFC namespaces defined under `standards.buildingsmart.org` is that there are different namespaces for different IFC versions. It feels better to have a single unified definition of the base concepts for the linkage (globalID, projectID, etc) and then dip into the version-specific namespaces for the actual alignment of concepts (so `ifc:Ifcbuilding` would be from a particular IFC namespace defined by BuildingSmart but the Brick linkage to its global ID in another file would be in our IFC alignment namespace?)

epaulson:
OK. I don't want to bikeshed this too much - what you've got now is certainly enough to get us going so let's go with that for 1.3.0 and let's add more when we see a clear need for it.

PR #360, opened 2021-12-03
Discharge supply match

shatdal:
The changes here remove inconsistencies between Point subclasses.

All min_ classes are matched with max_ classes
All supply_ classes are matched with discharge_ classes and declared equivalent
All occupied_ classes are matched with unoccupied_ classes

A new test is introduced for verifying the matches in future updates

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,8 @@
 
 Brick.ttl: bricksrc/*.py bricksrc/*.ttl bricksrc/definitions.csv generate_brick.py
 	mkdir -p extensions
-	python generate_brick.py
-	cd shacl && python generate_shacl.py
+	python3.10 generate_brick.py
+	cd shacl && python3.10 generate_shacl.py
 
 shacl/BrickShape.ttl: bricksrc/*.py generate_brick.py shacl/generate_shacl.py
 	cd shacl && python generate_shacl.py
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,8 @@
 
 Brick.ttl: bricksrc/*.py bricksrc/*.ttl bricksrc/definitions.csv generate_brick.py
 	mkdir -p extensions
-	python3.10 generate_brick.py
-	cd shacl && python3.10 generate_shacl.py
+	python generate_brick.py
+	cd shacl && python generate_shacl.py
 
 shacl/BrickShape.ttl: bricksrc/*.py generate_brick.py shacl/generate_shacl.py
 	cd shacl && python generate_shacl.py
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -2,8 +2,8 @@
 
 Brick.ttl: bricksrc/*.py bricksrc/*.ttl bricksrc/definitions.csv generate_brick.py
 	mkdir -p extensions
-	python generate_brick.py
-	cd shacl && python generate_shacl.py
+	python3.10 generate_brick.py
+	cd shacl && python3.10 generate_shacl.py
```
gtfierro, line 7: Can you change these back to `python` from `python3.10`? We need them generic so we can test across multiple versions of Python

tests/test_matching_classes.py
```diff
@@ -0,0 +1,216 @@
...
+
+    for row in res:
+        c1 = row.c.toPython().replace("https://brickschema.org/schema/Brick#", "")
+        if row.p:
+            c2 = row.p.toPython().replace("https://brickschema.org/schema/Brick#", "")
+            if c1 in dict:
+                dict[c1].append(c2)
+            else:
+                dict[c1] = [c2]
+
```
gtfierro, line 28: This logic can be replaced with a `defaultdict(list)` from the `collections` module

```python
from collections import defaultdict

d = defaultdict(list)

    res = g.query(q)

    for row in res:
        c1 = row.c.toPython().replace("https://brickschema.org/schema/Brick#", "")
        if row.p:
            c2 = row.p.toPython().replace("https://brickschema.org/schema/Brick#", "")
            d[c1].append(c2)

File: tests/test_matching_classes.py [hunk 3]
```diff
@@ -0,0 +1,216 @@
...
+import sys
+
+sys.path.append("..")
+
+from bricksrc.version import BRICK_VERSION  # noqa: E402
+from bricksrc.namespaces import BRICK  # noqa: E402
+
+
+def getDict(g, q):
+    dict = {}
```
gtfierro, line 16: We should avoid using builtin keywords as variable names; change this to `d` or something that's not a keyword

gtfierro (changes requested):
Thanks for the PR! Just a couple changes to make

tests/test_matching_classes.py
shatdal: Agreed. Changed accordingly.

tests/test_matching_classes.py
shatdal: Done

Makefile
shatdal: Done

PR #363, opened 2021-12-06
New Meter/Submeter Model

gtfierro:
To address documented issues and inconsistencies with Brick's meter/submeter model, members of the Ontology Working Group are proposing that new meter-specific relationships be added to Brick. These are:

- `hasMeter`/`isMeterOf`: associates a Meter equipment with a Brick entity. The meter observes the production/consumption of some substance by the entity (may be a location, equipment, system/collection). For example `chiller3 hasMeter meter1`
- `hasSubMeter`/`isSubMeterOf`: describes the submeter topology

One question: is this too much? Do we only want to have `hasMeter`/`isMeterOf` and use that to also capture the submeter relationship? The right interpretation is probably that the  submeter `hasMeter` the parent meter. This is consistent with `X hasMeter Y` indicating that `X` is observed by `Y` --- the parent meter observes the "consumption" of the submeter.

If we keep the submeter relationships, then I will need some SHACL shapes to disallow having a Meter instance on both ends of the `hasMeter` relationship, in order to avoid misuse.

@epaulson @wcrd @jbkoh what do you think?


gtfierro:
- maybe eliminate the "Building meter" --- this can be determined by `isMeterOf <building instance>`
- fix "bldg:meter" in example

jbkoh:
In terms of the choice of words, I'd prefer using a verb over a noun because "has" can be interpreted in different ways with its generic nature. E.g., it can be interpreted in the same way as `hasPart` for some users. My proposal is to use `meters`/`isMeteredBy`.

Is there any documentation regarding what "submetering" would mean? I thought it was more about measuring components of a parent other than generically meaning any subset of the parent's meter data. E.g., metering floor electricity usage vs building electricity usage. I suspect that it's sufficient to model `hasPart` within the objects of metering. E.g., meter_a meters floor_a, meter_b meters building_a, and floor_a isPartOf building_b.
 but maybe I miss something and want to check the notes. (the most recent google groups discussion seems not detailed for submetering.)

gtfierro:
We want the submetering relationships to be explicit because this allows you to figure out what is metered and avoid double counting. One may have submeters attached to systems, or parts of systems (e.g., for chilled water consumption), or arbitrary sets of equipment. `hasPart` isn't sufficient for all of those. Having the submeter there allows you to observe the consumption/production of some quantity in more granularity than a fully aggregated meter such as a building meter.

Can you explain your reasoning for the following? I'm not sure I follow
> I'd prefer using a verb over a noun because "has" can be interpreted in different ways with its generic nature. E.g., it can be interpreted in the same way as hasPart for some users. My proposal is to use meters/isMeteredBy.

jbkoh:
I now recall the submetering discussion a bit. I'm still unsure if the explicit `hasSubmeter` is going to be useful in general and don't see why `hasMeter` cannot solve the problem you mentioned. A user can easily pull the things or the locations that meters are metering and compare them from the different meters. E.g., for a building meter, we can know that the meter is measuring all the things located in the building by a query (either after inference or not). Moreover, if there are two meters measuring different sets with some intersection, the user has to go back to comparing them anyway as `hasSubmeter` cannot express that. Thus, I support a generic approach as a first step. `hasSubmeter` might turn out to be useful later, but we can add that at that time. I'd like to avoid adding a feature without strong use case support. We can always gradually add. Again, maybe I missed some context, and happy to revisit such discussions.

Regarding `hasMeter` vs `meters`, "has" in `hasMeter` might be confusing as some people take it as “possession” of the meter, similar to `hasPart`, whereas we want to model the behavior of meters, which is "metering". Assume a hypothetical AHU equipped with its own meter. Then the meter is a part of the AHU so `AHU--hasPart-->Meter` should exist for the compositional relation (example application: asset management). However, the purpose of the meter is metering and should be represented separately as `Meter--meters-->AHU` (example application: energy usage analysis). My point is those two relations are different, and if we use `hasMeter`, some people interpret it as the former one, and other people read it as the latter one. Of course, documentation can help, but we should be more instructive whenever we can.

gtfierro:
TODOs:
- add `Virtual_Meter`?
- remove substance/quantity-flavored meter classes?

gtfierro:
Meters can be associated with: Locations, Equipment and Collections. *Not Points*. Need a shape for this.

jbkoh:
What was the consensus on the last issue? I'm still on the side of standardizing usage of Collections for metered objects like Zones for feeds, so that users can expect the same thing across buildings.

epaulson:
I think we're mostly set on meters pointing to collections, but still a bit up in the air about if we really want to go all that way for the really simple case of just one piece of equipment on a meter - do you have to have a collection that only has one element to have a meter?

(I think we're coming around to 'meters' as the property name, though)

The current plan is to start writing out documentation for docs.brickschema.org and see how it feels when we have to explain the why and the how of it - what will it feel like from a perspective of a user? I don't think we're going to get to that until after the holidays, though.

gtfierro:
@jbkoh @epaulson I've finished what I can remember of the meter implementation we discussed, and wrote up some documentation on the model here: https://github.com/BrickSchema/docs/pull/10 . Let me know what you think!

jbkoh:
@gtfierro This PR seems to have a rebase problem and has many unrelated changes. Would you be able to clean it up? I will check the updated document separately though.

gtfierro:
@jbkoh all of the relevant changes are in `rules.ttl` and `properties.ttl` as well as the example files

examples/submeter_hierarchies/analyze.py
```diff
@@ -0,0 +1,78 @@
...
+    for row in g.query(all_meter_query):
+        print(row)
+    print("-" * 80)
+
+"""
+Gets the building meter and its peak (max) power sensor
+"""
+building_meter_query = """SELECT DISTINCT * WHERE {
+    ?meter rdf:type/rdfs:subClassOf* brick:Meter .
+    ?meter brick:isMeterOf/rdf:type/rdfs:subClassOf* brick:Building .
```
jbkoh, line 45: do we use `brick:isMeterOf` instead of `meters`?

examples/submeter_hierarchies/analyze.py
```diff
@@ -0,0 +1,78 @@
...
+    g.expand("shacl")
+    return g
+
+
+FILES = [
+    "main-and-submeter.ttl",
+    "multiple-submeters.ttl",
+    "single-meter.ttl",
+]
+
```
jbkoh, line 18: Can we add another example for accommodating Collections? This will help us to validate how the collections/non-collections would work for `meters`.

jbkoh (changes requested):
Looks good overall. I left two minor comments.

Please make it sure the PR doesn't contain unrelated commits next time. I'd like to make it sure there are no unreviewed changes.

jbkoh:
Another thing we haven't really discussed is expressing what meter-related Points measures. I don't think there is much controversial but it will make this PR more complete I think. Two questions come to my mind for now.

- Are there any cases where different points in a meter measure different entities in the graph? I don't think so, but just planting a seed for thought.
- Users (app queries) are eventually interested in Points (e.g., `Power Sensor`) which carry actual data. In that case, the current model forces users to query for actual meters too to understand what the points actually measure. For example,
```sparql
select ?sensor ?target {
  ?sensor a brick:Power_Sensor.
  ?meter brick:hasPoint ?sensor.
  ?meter a brick:Meter.
  ?meter brick:measures/hasPart* ?target.
}
```
  I think it's acceptable and in case users feel this is not intuitive, we could introduce some relationship for point→measured entities in the future (and bind it with some rules with equipment--measures-->target)

OskarNSE:
Looks good! A few comments:
- Forgive my ignorance but why not use “a” instead of “rdf:type/rdfs:subClassOf*" in the SPARQL examples?
- site-prefix sometimes used in the PV-example
- Does this make e.g. `brick:Building_Electrical_Meter` redundant? Two non-equivalent ways of expressing the same thing
- Are virtual meters treated the same ways as real ones? 
- I think `brick:meters` sounds more natural than `brick:isMeterOf` 

About the energy-usage-types topic. As discussed on the call, I agree that Meter subclasses are not the right way to do it, maybe equipment/locations could have a usage type property. But this can wait for another PR

examples/submeter_hierarchies/analyze.py
gtfierro: fixed!

gtfierro:
Thanks for the comments @jbkoh and @OskarNSE! 

> Please make it sure PRs don't contain unrelated commits next time. I'd like to make it sure not to include unreviewed changes.

When possible, yes -- sometimes merge commits cause issues though. Especially when the PR is old

> Are there any cases where different points in a meter measure different entities in the graph? I don't think so, but just planting a seed for thought.

No -- this would be a different meter

> I think it's acceptable and in case users feel this is not intuitive, we could introduce some relationship for point→measured entities in the future (and bind it with some rules with equipment--measures-->target)

This would complicate the model even further unless we decided to remove Meters. Maybe we can revisit in the future but not for now

> Forgive my ignorance but why not use “a” instead of “rdf:type/rdfs:subClassOf*" in the SPARQL examples?

Just habit! I've fixed this in the docs

> site-prefix sometimes used in the PV-example

Thanks! fixed

> Does this make e.g. brick:Building_Electrical_Meter redundant? Two non-equivalent ways of expressing the same thing

I'm adding an inference rule that will add the correct "Building" class when a meter "meters" a building. I'm reluctant to remove the `building_` flavors due to backwards compatibility reasons.

> I think brick:meters sounds more natural than brick:isMeterOf

Agreed! We are not using isMeterOf and I've removed the old relationship from the PR

examples/submeter_hierarchies/analyze.py
gtfierro: There's one in the docs -- is that enough?

gtfierro:
> Are virtual meters treated the same ways as real ones?

Yes, that is the idea. @OskarNSE @jbkoh @epaulson do you think I should add a `brick:Virtual_Meter` class, or have it be an entity property / attribute? If we go the route of making a `brick:Virtual_Meter` class, should we also create corresponding subclasses for different kinds of quantities (steam, electricity, etc), or just keep it as a single `Virtual Meter` class?

jbkoh:
`Virtual_Meter` may be useful, though I'd be cautious of expanding its type into different substance types. One can associate a corresponding meter type and `Virtual_Meter` at the same time if it's really needed because `Meter` doesn't imply physical meters only at this point. E.g.,
```turtle
ex:meter a brick:Virtual_Meter, brick:Electric_Meter.
```
A property like `brick:isVirtual` can be an alternative to having too many types of virtual meters. Maybe it's generally useful for modeling virtual devices such as virtual sensors and virtual controllers.

I don't mind either at this point. And even we can defer the decision on Virtual_Meter in a separate PR (aiming for the next version) if there can be too many directions to discuss.

OskarNSE:
Yes, the virtual concept is larger than just meters and a property seems like a better solution. Don't want to get too much off topic but are there cases where an entity can be more than one class like `ex:meter` above (that is not inferred by Brick)? If not, to get better consistency maybe it could be enforced (or documented) that the modeller only chooses one class to represent their thing?

gtfierro:
I'm ok with deferring the discussion of virtual annotations to a future PR, as it does seem like there is a larger question here. Are we ok with merging in this PR w/o the virtual meter stuff, then?

epaulson:
I'd still like some time to review

jbkoh:
> Yes, the virtual concept is larger than just meters and a property seems like a better solution. Don't want to get too much off topic but are there cases where an entity can be more than one class like `ex:meter` above (that is not inferred by Brick)? If not, to get better consistency maybe it could be enforced (or documented) that the modeller only chooses one class to represent their thing?

@OskarNSE Brick and its underlying meta-models RDF/RDFS/OWL/SHACL allow having multiple classes per instance as long as the classes are not conflicting each other based on the rules explicitly defined in the model (e.g., `owl:disjointWith`, conflicting SHACL shapes, etc.) I don't see much confusion in the hierarchy that makes users instantiate classes inconsistently, but if you think that problem is significant, I'd love to discuss it more in a separate thread.

epaulson:
I would like to include virtual meters in 1.3, even if we might later want to expand the 'virtual' concept beyond meters. My gut says do this as a virtual meter class instead of an `isVirtual` property but I'm not heart-set on that. 

I'll leave a comment on the doc pull request - I think the info is good, though I would probably reorder everything and put the submeters last and put the example that's currently at the end as the very first thing, because that's the first place where you see how to go from a sensor with a timeseries to a 'meters' relationship to a space. We can build up from there.

In looking through our aggregate specifications, it's not clear to me what a basic instantaneous power meter should be using or if I always need one. I think I see how to do cumulative and I see how to do averaging/peak meters.

This can wait until 1.3.1 or beyond, but we will want to have a way to talk about net metering and sending power back to the grid - that was the main thing I saw that we were missing from Haystack. 

We may want to review our energy/power/demand/voltage/current/etc sensor class hierarchy at some point. Sometimes we have electrical subclasses in there and sometimes we don't

We should add a non-electrical metering example - a water example and/or a gas example

I've got a few more things to check on submeters so I need just a bit more time. 

If you've got a chance, to see if we're missing things, take a look at 
* the Haystack docs: https://project-haystack.org/doc/docHaystack/Meters 
* the JCI docs: https://docs.johnsoncontrols.com/bas/r/Johnson-Controls/en-US/OpenBlue-Enterprise-Manager-Configuration-Guide/4.0/Meter-configuration/Assigning-data-points-to-meters/Meter-configuration-examples

gtfierro:
> I would like to include virtual meters in 1.3, even if we might later want to expand the 'virtual' concept beyond meters. My gut says do this as a virtual meter class instead of an isVirtual property but I'm not heart-set on that

I added `brick:Virtual_Meter` to the PR and included an example file

>  net metering and sending power back to the grid - that was the main thing I saw that we were missing from Haystack

We do have the `powerFlow` entity property which supports the values "import", "export", "net", and "absolute" for describing the meter.

jbkoh:
> I added `brick:Virtual_Meter` to the PR and included an example file

@gtfierro Sorry for making this comment after the last change. I didn't know it was already concluded during the call. I gave more thoughts after the discussion today, and I think the property is a better choice for all the reasons raised today. (cc @epaulson)

- Multi-instantiation as in my example above is not much aligned with OOP mentality. A class may have multiple super classes but an instance should be an instantiation of a single most specific class. Even though RDF is not OOP, actual usage of our model will be (and is) in the programming world with OOP in many cases. It's difficult to support object mapping with multiple instantiations. 
- Virtual_Meter represents a set of the meter types while the rest of the types are physical. We don't like to introduce Physical_Meter while the property `isVirtual` can represent both of them by true/false. (I brought physical meters during the call to argue the property was better. I didn't mean to support introducing Physical_Meters.)
- The property `isVirtual` is more scalable to other virtual entities if needed. The examples include virtual sensors and controllers.
- In a query model with Virtual_Meter, the measurement type should be still asked in the query, so I don't think the query pattern with `Virtual_Meter` is simpler than `isVirtual`.
    - Please return equipment which is an instance of `Virtual_Meter` and `Electric_Meter`, vs.
    - Please return equipment which is an instance of `Electric_Meter` and `isVirtual` is True.

While the rest can be mitigated by inference between type and properties later, the first point should be respected more imho.

A possible downside of `isVirtual` property is that users may put arbitrary semantics to "virtual" than we would define. For example google DBO has different definition of [virtual devices](https://github.com/google/digitalbuildings/blob/8df8b24d03eb38f5ba475ee02fe08741f74c1886/ontology/docs/building_config.md#virtual-devices)
> In many cases the logical device and the reporting device are the same thing (i.e. the device that sends data is also the entity type you want in your model). When the two are different (i.e. the logical device does not send its own data) a virtual device is required.
Here, even though a VAV is a physical device, if its controller is not directly exposed but through a supervisory controller or a gateway, it's a virtual device. Still I think the concept of virtual is quite commonly accepted as we'd define, it's noteworthy. 

I probably missed some of the arguments for `Virtual_Meter` that Erik was saying during the call. If Erik can write it down here, I will take more time to fully ingest it.

epaulson:
Virtual as a class vs virtual as a property isn't something I feel that strongly about. We have tooling that uses classes for documentation and browsing so I think it's reasonable to just make 'Virtual_Meter' a class because it's discoverable in our ontology browser. But this isn't a hill I'm going to die on. 

However, I am 100% not conceding any design philosophy that says an entity in a Brick model should have at most one rdf:type/taking any position that says multi-instantiation is bad.

jbkoh:
> I would actually argue that having an isVirtual property is more confusing! Many Brick classes, especially points, are implicitly virtual. Does that mean that the isVirtual property must be added to all of those other entities, either manually or automatically? What other kinds of entities might be "virtual" in our model? Points, controllers, and meters are what I can think of; of these, only meters and controllers have meaningful physical alternatives. Why add a very generic property (isVirtual) if it is only going to be used in very specific cases?

Our properties do not have to be general but instead they can be limited only to necessary concepts. I.e., only meters and controllers may have the property isVirtual, same as yearBuilt, coolingCapacity, etc.

That said, I agree the term isVirtual sounds generic and it also looks arbitrary if we don't associate that with Points. I'm fine with the current Virtual_Meter implementation.

epaulson:
Being to say a meter is virtual is important. If it's a class or a property, I don't care. 

If we're going to add a property I'd prefer that for now we keep it very narrowly scoped, like 'ex:fl1Meter a brick:meter;  brick:virtualMeter true .' instead of trying to solve a wider ontological question about the nature of reality and what is virtual and what is not virtual.

gtfierro:
I've added 'virtual meter' as a property instead of a class. Looks like this:

```
bldg:meter a brick:Electrical_Meter ;
   brick:isVirtualMeter [ brick:value true ] .
```

which keeps it consistent with our entity property design. How does this look to all of you?

I'm writing up the brickschema docs now

gtfierro:
Updated docs [here](https://github.com/BrickSchema/docs/blob/741380f23ad4ead8083d8a35b262fa9c514dec30/modeling/meters.md#virtual-meters)

Issue #330, opened 2021-09-26
Clear distinction of delta from their related classes


jbkoh:
In the current design, delta concepts are subsumed by its related classes. E.g., `Water_Differential_Temperature_Sensor` is a subclass of `Water_Temperature_Sensor`. However, even though they are related, they are different. E.g., if a user queries, I would like to know the temperature of water, the query will be asking for getting `Water_Temperature_Sensor`, which incorrectly includes `Water_Differential_Temperature_Sensor` now. They will get totally different values for the same place.

This applies to `Deadband` concepts as well.

Those two should be differentiated all over the place. If I see agreement here, I will create a PR for that.

A side node, CO2_Differential_Sensor should be changed to `Differential_CO2_Sensor` for consistency.

gtfierro:
Definitely agree; I like the idea of putting Differential sensors under their own Sensor subtree. We can do the same for putting Deadband setpoints under their own Setpoint subtree.

jbkoh:
ok will work on a PR.

gtfierro:
Sounds good! Make sure to link the PR to this issue when it is submitted

jbkoh:
closing this one by observing #375

PR #375, opened 2022-02-25
adjust the hierarchy for differential setpoints

jbkoh:
Addressing https://github.com/BrickSchema/Brick/issues/330

- Added `Differential_Setpoint` for grouping all the difference related setpoints. Not super necessary, I figured it's useful to group detailed setpoints this way. Open to removing it.
- Introduced Differential_Temperature_Setpoint for grouping Water_Differential_Temperature_Setpoint and Differential_Air_Temperature_Setpoint.
  - Those two are another example of inconsistent naming but I left them as they are for now as I'm unsure if we can introduce breaking changes in the upcoming release. I do love to modify this sooner than later.
- Removed `Chilled_Water_Pump_Differential_Pressure_Deadband_Setpoint` as it's against our base principle on not having detailed equipment flavored points


gtfierro (changes requested):
It looks like you deleted a term which still has a definition or added a definition which doesn't have a corresponding class. Can you run the tests locally and check the contents of the `obselete_definitions.json` file to check?

jbkoh:
@gtfierro Removed an obsolete definition and put more description to this PR. Please check. Thanks.

gtfierro (approved):

Issue #376, opened 2022-03-02
Adding dual duct AHU and hot and cold deck equipments


lazlop:
I need to model a dual duct AHU system. Currently, I'm thinking the best way to do this would be to have 3 pieces of equipment. A dual duct AHU, a hot deck, and a cold deck. The dual duct AHU will contain the hot and cold decks, and the hot and cold decks will  feed other pieces of equipment. The diagram below shows something like this. Can I add this to the brick schema?
![image](https://user-images.githubusercontent.com/69093818/156326661-bdffe236-2a68-4219-8c88-2d14cc154f3d.png)

gtfierro:
Hi @lazlop! Thanks for filing the issue. Can you put together a PR implementing what you have in mind? We can discuss it at one of the upcoming Brick ontology working group meetings (https://brickschema.org/blog/working-groups/). I could see hot/cold deck as being "equipment", perhaps; or we can create a dual-duct AHU but otherwise simplify the rest of the topological model; or perhaps another approach altogether. Looking forward to what you have in mind!

PR #378, opened 2022-03-03
Adding concepts for Dual Duct HVAC systems

lazlop:
Added concepts for dual duct HVAC systems as mentioned in #376


gtfierro:
@lazlop let me know when you get a chance to come back to this and make the edits we discussed on the ontology working group call last week!

lazlop:
@gtfierro Sorry, the end of March has been pretty busy and I just got around to this! This is my first attempt at writing a SHACL shape, let me know if it looks right. Also, let me know where the location of the SHACL shape should be. Just added it to BrickEntityShapeBase.ttl for now

bricksrc/equipment.py
```diff
@@ -640,6 +642,14 @@
                 "tags": [TAG.Equipment, TAG.RTU],
                 OWL.equivalentClass: BRICK["Rooftop_Unit"],
             },
+            "Dual_Duct_Air Handling Unit": {
```
gtfierro, line 645: You will need underscores instead of spaces here

shacl/BrickEntityShapeBase.ttl
```diff
@@ -160,3 +160,25 @@ bsh:hasLocationShape a sh:Nodeshape;
     ];
     sh:message "Points are a virtual concept and always belonging to a physical device, represented by Equipment. Thus, it cannot have a Location alone."
     .
+
+bch:hasHotColdDeck
```
gtfierro, line 164: This should be 'bsh' instead of 'bch'

gtfierro (changes requested):
No worries @lazlop! I made a couple small suggestions but I think this is very close

gtfierro (approved):

gtfierro:
Thanks @lazlop!

PR #381, opened 2022-03-25
docs: clarify usage of `feeds` and `isFedBy` relationships

hicksjacobp:
This is just a simple improvement to the `feeds` and `isFedBy` relationship descriptions. I was browsing the ontology and it wasn't clear to me through just "subject" and "object" about what types of things should or should not be using these relationships.


gtfierro:
Thanks for taking the time to do a PR! I'll be sure to kick off the nightly doc rebuild so that the updated definitions are available on https://brickschema.org w/n the next day or so

PR #384, opened 2022-03-30
Use ref-schema instead of old internal Brick relationships

gtfierro:
Instead of `brick:timeseries` and related family now we will use the `ref-schema` definitions from https://github.com/gtfierro/ref-schema. This will also require fixing the Brick docs repository


epaulson:
I may not be following it correctly - but you're only using 
`@prefix bacnetref: <https://brickschema.org/schema/Brick/ref/bacnet#> .`

for the OWL import - was there something else you were going to do with it?

gtfierro:
I can't find `bacnetref` in the code anywhere now --- can you point me to where you are looking?

epaulson:
https://github.com/gtfierro/ref-schema/blob/main/model/bacnet.ttl

gtfierro:
That file imports Joel's BACnet.ttl file so it can use some of the definitions. A few other import statements are used to "build" the ref-schema ttl file but I don't include them in the final build.

epaulson:
In looking through https://github.com/BrickSchema/Brick/blob/v1.3.0/support/bacnet.ttl - line 9238, it looks like it's specing analog-value:1

but in our examples, we're using
bacnet:object-identifier "analog-value,5" 

(maybe we just have an old version of bacnet.ttl checked in?)

epaulson:
We should also think about pulling in bacnet:units into the ref build?

gtfierro:
Oh, good catch. The `analog-value,1` syntax is the correct one, and this seems to be remedied in the latest version of `bacnet.ttl`, which I'm fixing in https://github.com/BrickSchema/Brick/pull/389

> We should also think about pulling in bacnet:units into the ref build?

I definitely can -- I was initially worried about pulling in a duplicate "units" field that may have identifiers/strings that cannot be automatically resolved against the QUDT units. However, I suppose we can't begin rectifying those two identifiers unless they are both represented in the model

gtfierro:
The units definitions in the `bacnet.ttl` file are all EnumerationValues:

```ttl
bacnet:EngineeringUnits-watts-per-square-foot a bacnet:EngineeringUnitsEnumerationValue ; 
    bacnet:name "watts-per-square-foot" ;
    bacnet:value "34"^^xsd:nonNegativeInteger .
```

I can see a few options:

1) import these into the ref-schema distribution
2) import *all* of `bacnet.ttl` into ref-schema distribution
3) "lie" and change the datatype of the `bacnet:units` to the actual numerical enumeration or a string value, and resolve it later

JoelBender:
String an unsigned integer values will only work for standard properties, BACnet enumerations that can be extended by vendors have overlapping ranges and no control over string descriptions.  BACnet/WS resolves this by leaving them as strings and putting the vendor identifier in front like "15-glarcons" for Cornell, vendor identifier 15.

I think that (1) surgically extracting these (engineering units, property identifiers, maybe object types) for ref-schema is perfectly fine, (2) means that 99.44% will never be used in this context and seems bloated, (3) is irritating (with the non-zero possibility the string version might be something that needs to be added for BACnet/RDF be accepted by the the DM-WG).

epaulson:
Oh, this got more complicated than I realized, and also I should have been more precise - I was only suggesting adding bacnet:units as one of the properties the SHACL checks for/permits in a ref:BACnetReference - here's what I was doing:

```ttl
bldg:reaPPum_y_bacnet a ref:BACnetReference ;
    bacnet:description "Electical Power consumption of pump" ;
    bacnet:object-type "analog-value" ;
    bacnet:object-identifier "analog-value,16" ;
    bacnet:object-name "reaPPum_y" ;
    bacnet:units "watts" ;
   bacnet:objectOf bldg:boptest-proxy .
```

but it looks like things have changed and bacnet:units isn't an xsd:string anymore. I'm not sure we want to keep chasing after it and pull in all of the bacnet:EUEVs

Maybe it's more trouble than it's worth to pull in units - since you've got the object identifier, if you want to find out what the units are, just go out through BACnet and ask the object what it is.

examples/ifc/ifc.ttl
```diff
@@ -1,24 +1,24 @@
...
+        ref:ifcGlobalID "123" ;
+        ref:ifcName "Example Space" ;
     ] .
 
 bldg:space2 a brick:Space ;
-    brick:hasIFCReference [
-        brickifc:hasProjectReference bldg:ifc_project ;
-        brickifc:globalID "124" ;
-        brickifc:name "Example Space" ;
+    ref:hasExternalReference [
```
jbkoh, line 18: I understand the type of this reference can be inferred, but still it'd be useful to have the type for users.

examples/ifc/ifc.ttl
```diff
@@ -1,24 +1,24 @@
 @prefix bldg: <urn:example#> .
 @prefix brick: <https://brickschema.org/schema/Brick#> .
-@prefix brickifc: <https://brickschema.org/extension/ifc#> .
+@prefix ref: <https://brickschema.org/schema/Brick/ref#> .
 
-bldg:ifc_project a brickifc:Project ;
-    brickifc:projectID "abc" ;
-    brickifc:fileLocation "file://./building.ifc" ;
+bldg:ifc_project a ref:ifcProject ;
+    ref:ifcProjectID "abc" ;
```
jbkoh, line 6: - Just curious, is it okay this to be different from ifcGlobalID?
- Where is the definition of ifcProjectID?

support/ref-schema.ttl
```diff
@@ -0,0 +1,196 @@
...
+            sh:datatype xsd:string ;
+            sh:path ref:ifcName ],
+        [ a sh:PropertyShape ;
+            skos:definition "The global ID of the entity in the IFC project" ;
+            sh:datatype xsd:string ;
+            sh:minCount 1 ;
+            sh:path ref:ifcGlobalID ],
+        [ a sh:PropertyShape ;
+            skos:definition "Reference to an IFC Project object, containing the project ID" ;
+            sh:class ref:ifcProject ;
```
jbkoh, line 171: Can we have a shape for ifcProject as well?

support/ref-schema.ttl
```diff
@@ -0,0 +1,196 @@
...
+    rdfs:label "ifcName" ;
+    rdfs:range xsd:string ;
+    skos:definition "The name of the IFC entity" .
+
+ref:storedAt a owl:DatatypeProperty ;
+    rdfs:label "storedAt" ;
+    rdfs:range xsd:anyURI ;
+    skos:definition "A reference to where the data for this TimeseriesReference is stored"@en .
+
+ref:BACnetReference a owl:Class,
```
jbkoh, line 126: Shouldn't this be a subclass of `bacnet:Object`?

support/ref-schema.ttl
```diff
@@ -0,0 +1,196 @@
...
+        sh:NodeShape ;
+    rdfs:subClassOf ref:ExternalReference ;
+    skos:definition "A reference to the BACnet object represented by this entity." ;
+    sh:or ( [ sh:property [ a sh:PropertyShape ;
+                        sh:datatype xsd:string ;
```
jbkoh, line 131: shouldn't this be URI in the bacnet ontology? e.g. bacnet:present-value. Thus may change it to
```turtle
sh:class bacnet:StandardProperty
```

support/ref-schema.ttl
```diff
@@ -0,0 +1,196 @@
...
+                        sh:minLength 1 ;
+                        sh:path bacnet:object-name ],
+                    [ a sh:PropertyShape ;
+                        sh:datatype xsd:string ;
+                        sh:minCount 1 ;
+                        sh:path bacnet:object-identifier ] ] [ sh:property [ a sh:PropertyShape ;
+                        skos:definition "Clause Q.8 BACnet URI scheme: bacnet:// <device> / <object> [ / <property> [ / <index> ]]" ;
+                        sh:datatype xsd:string ;
+                        sh:path ref:BACnetURI ] ] ) ;
+    sh:property [ a sh:PropertyShape ;
```
jbkoh, line 151: Doesn't this force having a bacnet device for being a bacnet object always? Is it necessary? I'm just imagining a user who is too lazy to model bacnet devices but only objects.

jbkoh (changes requested):
Looks great. I left several comments. Please consider them. Thanks!

examples/ifc/ifc.ttl
gtfierro: Users can add the type to the object of `ref:hasExternalReference` but it can also be inferred

support/ref-schema.ttl
gtfierro: If you don't have the Device then it is difficult to find the object; I think it is better to model both and to remind the user if they do not model  the device

examples/ifc/ifc.ttl
gtfierro: Forgot to push from my local repo and sync with the copy in the Brick project. ifcProjectID is defined now

support/ref-schema.ttl
gtfierro: Yes it's in there now; see above

support/ref-schema.ttl
gtfierro: It could be, but there are other mandatory fields on `bacnet:Object` that aren't totally necessary for this particular use case. I felt it was more ergonomic to just require a subset of those fields

jbkoh: I see a lot min counts there. I agree.

examples/ifc/ifc.ttl
jbkoh: Yeah, as I said, I understand the type can be inferred. I think it's useful to have the type explicitly in this specific example because it's easier for users to read. For those not familiar with the inference logic which is embedded deep in the ttl file, it's just easier to see the type of the node explicitly. Without it, one may think the properties can be arbitrarily added here.

support/ref-schema.ttl
epaulson: BACnetReference should NOT be a subclass of bacnet:Object. bacnet:Object are the protocol things, a BACnetReference isn't quite the same thing

support/ref-schema.ttl
gtfierro: Joel has since added `bacnet:Property` class w/ the `bacnet:Present_Value` instance which best captures what we're doing here.

PR #386, opened 2022-04-07
Add currentValue property

gtfierro:
This aligns with the 223P model via `rdfs:subPropertyOf`, but uses Brick-named properties instead. Example included

### DIFF
--- a/bricksrc/entity_properties.py
+++ b/bricksrc/entity_properties.py
@@ -383,7 +383,6 @@ building_primary_function_values = [
 shape_properties = {
     BSH.AreaShape: {"units": [UNIT.FT2, UNIT.M2], "datatype": BSH.NumericValue},
     BSH.LastKnownValueShape: {
-        "datatype": BSH.NumericValue,
         "properties": {
             BRICK.timestamp: {"datatype": XSD.dateTime},
         },
--- a/bricksrc/entity_properties.py
+++ b/bricksrc/entity_properties.py
@@ -383,6 +383,7 @@ building_primary_function_values = [
 shape_properties = {
     BSH.AreaShape: {"units": [UNIT.FT2, UNIT.M2], "datatype": BSH.NumericValue},
     BSH.LastKnownValueShape: {
+        "datatype": RDFS.Resource,
         "properties": {
             BRICK.timestamp: {"datatype": XSD.dateTime},
         },
--- a/bricksrc/entity_properties.py
+++ b/bricksrc/entity_properties.py
@@ -14,11 +14,11 @@ entity_properties = {
         RDFS.range: BRICK.DeprecationShape,
         RDFS.label: Literal("Deprecation Notice"),
     },
-    BRICK.currentValue: {
-        SKOS.definition: Literal("The current value of the Point entity"),
+    BRICK.lastKnownValue: {
+        SKOS.definition: Literal("The last known value of the Point entity"),
         RDFS.domain: BRICK.Point,
-        RDFS.range: BSH.CurrentValueShape,
-        RDFS.label: Literal("Current Value"),
+        RDFS.range: BSH.LastKnownValueShape,
+        RDFS.label: Literal("Last known value"),
     },
     BRICK.area: {
         SKOS.definition: Literal("Entity has 2-dimensional area"),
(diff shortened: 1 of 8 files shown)


gtfierro:
@hammar what do you think of the implementation? Does this fit your needs?

hammar:
With the caveat that my understanding of the Brick structure is limited, this looks like reasonable structure and naming, at least the provided example. 

Two questions: 
1. Is there any way of making the expectation that a CurrentValueShape should have a brick:value more explicit? At present I don't see this either in the shape itself or in it's parent shapes (looking at generated brick.ttl). 
2. Is there a risk that the term "Current" is ambiguous (since you already have CurrentShape/CurrentFlowTypeShape referring to electrical current)?

Best,

Karl

gtfierro:
> Is there any way of making the expectation that a CurrentValueShape should have a brick:value more explicit? At present I don't see this either in the shape itself or in it's parent shapes (looking at generated brick.ttl).

Yes! I just pushed a fix for this.

> Is there a risk that the term "Current" is ambiguous (since you already have CurrentShape/CurrentFlowTypeShape referring to electrical current)?

I was thinking this too. Some other suggestions might be "recent value", "present value" or "timestamped value". Per my reading of REC 3.3, it seems you use "last value"/"last value time"? I could be ok with that terminology in Brick

epaulson:
For a name, I'd prefer 'presentValue' or 'lastKnownValue' but I'm not heart-set on anything in particular, and would defer to something else, especially if that makes compatibility with other ontologies easier.

hammar:
> > Is there any way of making the expectation that a CurrentValueShape should have a brick:value more explicit? At present I don't see this either in the shape itself or in it's parent shapes (looking at generated brick.ttl).
> 
> Yes! I just pushed a fix for this.

Just compiled and saw it works in the generated TTL. Thanks!

> > Is there a risk that the term "Current" is ambiguous (since you already have CurrentShape/CurrentFlowTypeShape referring to electrical current)?
> 
> I was thinking this too. Some other suggestions might be "recent value", "present value" or "timestamped value". Per my reading of REC 3.3, it seems you use "last value"/"last value time"? I could be ok with that terminology in Brick

We went back and forth on this a whole lot in REC. I actually would have preferred something like "currentValue", "presentValue", or the like, indicating the _intent_ for this to be used for up-to-date data. The counter-argument was that in many systems this value may for technical reasons be delayed substantially from the real world state it mirrors and in that case "currentValue" might be misleading to some classes of users; so "lastValue" that we finally settled on essentially means "the last value that we know of", giving no guarantee that this is representative of the right-now-state. 

Arguably this name is also misleading, as "last" is synonymous with "final" -- "latestValue" or Erik's suggested "lastKnownValue" would in that sense probably be more correct. 

I don't have a super strong opinion on this, but if I had to vote, I'd go for "presentValue" or maybe even "presentObservation" (since "value" is the next level indirection).

gtfierro:
Of the suggestions, I like `lastKnownValue` the best --- it captures the semantics of being up-to-date perhaps better than the others. I also prefer `value` over `observation` because not all of the values will be observations as you would find in a "sensor": alarms, commands, setpoints, etc

jbkoh:
- is the last known value always numeric? https://github.com/BrickSchema/Brick/pull/386/files#diff-af59cfca8ae8a3876c10323dca28df8981ff0e8423878d6e8cdd70ef6b888ee1R386
- Shall we add a `sh:maxcount` for `lastKnownValue` as 1?

gtfierro:
Good feedback, @jbkoh . I've removed the numeric requirement (I don't think we have *that* much of a range, but we should include string values at least) and added a rule + test for restricting the maxCount to 1

hammar:
In the latest commit, it seems the explicit linkage LastKnownValueShape no longer has a PropertyShape for the brick:value property. I guess such a shape doesn't have to require a numeric value, but it should still be required to exist, right?

gtfierro:
Sorry, yes! I added the datatype to be `rdfs:Resource` instead, but I got distracted and didn't commit/push

hammar:
> Sorry, yes! I added the datatype to be `rdfs:Resource` instead, but I got distracted and didn't commit/push

Looks good to me -- thanks!

PR #401, opened 2022-06-03
Add and adjust classes

jbkoh:
add many missing concepts as in the commit history.


epaulson:
Could we leave Exhaust_Fan_{Enable, Disable}_Command in the schema and just mark them deprecated?

jbkoh:
> Could we leave Exhaust_Fan_{Enable, Disable}_Command in the schema and just mark them deprecated?

done.

epaulson:
I was thinking we should leave them in entirely, so they still show up as subclasses of Enable_Cmd and Disable_Cmd in the documentation/tooling and in the reasoner steps. We should add code to the ontology browser to warn when a class is deprecated, e.g. we style this page different so people know it's deprecated: https://brickschema.org/ontology/1.2/classes/Exhaust_Fan_Enable_Command

I think the shacl construct rule is important to say what to do with it if it's gone but I'm not sure it's quite enough? I guess I'd prefer to leave them in the class hierarchy for now, with the extra deprecated annotation. But I can easily be talked out of that!

bricksrc/command.py
```diff
@@ -5,6 +5,7 @@
     "Command": {
         "tags": [TAG.Point, TAG.Command],
         "subclasses": {
+            "Boiler_Command": {"tags": [TAG.Boiler, TAG.Command, TAG.Point]},
```
gtfierro, line 8: The `Boiler_Command` doesn't have a very descriptive definition --- what is the intended use? Like an `On_Off_Command` or `Enable_Command` on a Boiler?

jbkoh: Boiler_Command is meant to be generic for any commands for a boiler like Fan_Command (compared with Fan_Speed_Command.) We might expand it to Boiler_On_Off_Command, Boiler_Power_Command, etc.

gtfierro (approved):

PR #402, opened 2022-06-14
Add Tablet to equipment

anil72007:
Add Tablet to equipment


gtfierro:
@jbkoh would you consider adding a Tablet equipment to the #452 ? Servers and other ICT equipment would also be good candidates

gtfierro:
We will address this for 1.4.0; `Tablet` will be added under a new `ICT Hardware` (better name?) class under the new `ICT Equipment` class. We can take some other classes from https://dev.realestatecore.io/ontology/Asset/Equipment/ICT-/ICTEquipment too

PR #408, opened 2022-07-05
shade to blind and remove redundant dry cooler

jbkoh:


jbkoh:
[OskarNSE](https://github.com/OskarNSE) please check if this PR is okay for you, (I can't add you as a reviewer somehow. Maybe you disabled being searched?)

OskarNSE:
Thanks for the effort, looks good! There are a couple of things missing compared to the table I posted in the issue, like the Shading Zone and the points. They (or something similar) can be added later so no issue there. 

About "Louver", when discussing the nomenclature for shading equipment I never heard it. Can we remove it? 

Also, small typo in "An group of" -> "A group of"

bricksrc/equipment.py
```diff
@@ -291,7 +291,7 @@
 Define classes of HVAC equipment
 """
 hvac_subclasses = {
-    "Dry_Cooler": {"tags": [TAG.Equipment, TAG.HVAC, TAG.Dry, TAG.Cooler]},
+    "Dry_Cooler": {"tags": [TAG.Equipment, TAG.HVAC, TAG.Dry, TAG.Cool]},
```
gtfierro, line 294: I think `Cool` *does* want to be `Cooler` here

jbkoh:
@OskarNSE You are right that the rest will be discussed after 1.3.0 release. I just removed Louver. Thanks for the feedback and I'm merging this one.

Issue #358, opened 2021-12-01
Swap Supply/Return for Entering/Leaving for water system Points


gtfierro:
While the use of supply/return in water systems is not uncommon, the measurement location of a sensor is ambiguous from the use of the word supply/return because these systems involve loops. In the HVAC domain, supply/return are defined with respect to the zone: "supply to the zone" and "return from the zone".

It has been suggested from multiple sources that instead of "condenser water return" and "condenser water supply", we use terms like "leaving condenser water" or "entering condenser water" to clarify that the water being measured is leaving/entering the condenser (respectively).

@wcrd @epaulson do I have this correct?

gtfierro:
(on `v1.3.0` branch)

```python
import brickschema
g = brickschema.Graph().load_file("Brick.ttl")
res = g.query("""SELECT ?point WHERE {
  ?point rdfs:subClassOf+ brick:Sensor .
  ?point brick:hasSubstance/rdfs:subClassOf* brick:Water .
  { ?point brick:hasAssociatedTag tag:Return } UNION { ?point brick:hasAssociatedTag tag:Supply}
}""")
for row in res:
   print(row[0])
```

```
https://brickschema.org/schema/Brick#Chilled_Water_Supply_Flow_Sensor
https://brickschema.org/schema/Brick#Hot_Water_Return_Flow_Sensor
https://brickschema.org/schema/Brick#Hot_Water_Supply_Flow_Sensor
https://brickschema.org/schema/Brick#Return_Water_Flow_Sensor
https://brickschema.org/schema/Brick#Return_Condenser_Water_Flow_Sensor
https://brickschema.org/schema/Brick#Supply_Water_Flow_Sensor
https://brickschema.org/schema/Brick#Supply_Condenser_Water_Flow_Sensor
https://brickschema.org/schema/Brick#Return_Condenser_Water_Temperature_Sensor
https://brickschema.org/schema/Brick#Supply_Condenser_Water_Temperature_Sensor
https://brickschema.org/schema/Brick#Chilled_Water_Supply_Temperature_Sensor
https://brickschema.org/schema/Brick#Return_Water_Temperature_Sensor
https://brickschema.org/schema/Brick#Differential_Supply_Return_Water_Temperature_Sensor
https://brickschema.org/schema/Brick#Differential_Supply_Return_Water_Temperature_Sensor
https://brickschema.org/schema/Brick#Differential_Supply_Return_Water_Temperature_Sensor
[...]

PR #411, opened 2022-07-20
Entering/Leaving Water, not Supply/Discharge/Return Water

gtfierro:
Addresses #358 . Deprecation notices ~~in-progress~~ added


bricksrc/definitions.csv
```diff
@@ -362,11 +342,18 @@ https://brickschema.org/schema/Brick#Energy_Storage_System,A collection of devic
 https://brickschema.org/schema/Brick#Energy_System,"A collection of devices that generates, stores or transports electricity",
 https://brickschema.org/schema/Brick#Energy_Usage_Sensor,Measures the total amount of energy used over some period of time,
 https://brickschema.org/schema/Brick#Energy_Zone,A space or group of spaces that are managed or monitored as one unit for energy purposes,
-https://brickschema.org/schema/Brick#Entering_Water,Water that is entering a piece of equipment or system,
-https://brickschema.org/schema/Brick#Entering_Water_Flow_Sensor,Measures the rate of flow of water entering a piece of equipment or system,
-https://brickschema.org/schema/Brick#Entering_Water_Flow_Setpoint,Sets the target flow rate of entering water,
+https://brickschema.org/schema/Brick#Entering_Water,,
```
jbkoh, line 345: Why do we remove the definition of `Entering_Water`?

bricksrc/definitions.csv
```diff
@@ -362,11 +342,18 @@ https://brickschema.org/schema/Brick#Energy_Storage_System,A collection of devic
...
-https://brickschema.org/schema/Brick#Entering_Water_Flow_Setpoint,Sets the target flow rate of entering water,
+https://brickschema.org/schema/Brick#Entering_Water,,
+https://brickschema.org/schema/Brick#Entering_Water_Differential_Pressure_Deadband_Setpoint,Sets the size of a deadband of differential pressure of supply water,
+https://brickschema.org/schema/Brick#Entering_Water_Differential_Pressure_Integral_Time_Parameter,,
+https://brickschema.org/schema/Brick#Entering_Water_Differential_Pressure_Proportional_Band_Parameter,,
+https://brickschema.org/schema/Brick#Entering_Water_Flow_Sensor,Measures the rate of flow of hot supply water,
+https://brickschema.org/schema/Brick#Entering_Water_Flow_Setpoint,Sets the flow rate of hot supply water,
+https://brickschema.org/schema/Brick#Entering_Water_Temperature_Alarm,An alarm that indicates the off-normal conditions associated with temperature of the supply water.,
+https://brickschema.org/schema/Brick#Entering_Water_Temperature_Deadband_Setpoint,Sets the size of a deadband of temperature of supply water,
```
jbkoh, line 352: supply has not been changed to entering in those definitions.

jbkoh:
I'd like to improve the naming convention on some points like "Hot Water Leaving Temperature Sensor", which caused inconsistency in naming. They are not consistent with the rest like "Leaving Water Temperature Sensor". Can we make them all consistent like "Leaving Hot Water Temperature Sensor"? Basically, the class names should be more structured with Substance / Quantity names. That way, we can validate the class names easier with a proper substance hierarchy, and users can find things more easily and relate one to another in the hierarchy.

bricksrc/alarm.py
```diff
@@ -192,22 +192,19 @@
                         "tags": [TAG.Point, TAG.Water, TAG.Temperature, TAG.Alarm],
                         "parents": [BRICK.Water_Alarm],
                         "subclasses": {
-                            "Discharge_Water_Temperature_Alarm": {
```
jbkoh, line 195: I think the idea was to keep deprecated concepts in the hierarchy. Is there a different way to preserve that even if we remove them here?

bricksrc/deprecations.py
```diff
@@ -0,0 +1,159 @@
...
+    },
+    BRICK.Supply_Condenser_Water_Flow_Sensor: {
+        "version": "1.3.0",
+        "mitigation_message": "Swapped supply/return for entering/leaving with water-related poitns",
+        "replace_with": BRICK.Entering_Condenser_Water_Flow_Sensor,
+    },
+    BRICK.Heat_Exchanger_Supply_Water_Temperature_Sensor: {
+        "version": "1.3.0",
+        "mitigation_message": "Swapped supply/return for entering/leaving with water-related poitns",
+        # "replace_with": BRICK.Leaving_Condenser_Water,
```
jbkoh, line 97: Are those commented out intentionally? Just heads up.

bricksrc/rules.ttl
```diff
@@ -106,7 +106,7 @@ CONSTRUCT {
 bsh:DeprecationRule
     a   sh:NodeShape ;
     sh:property [
-        sh:path brick:deprecation ;
+        sh:path (rdf:type brick:deprecation) ;
```
jbkoh, line 109: Please add more description to this in the PR or as a comment in the code. It seems to be changed from the original deprecation design.

bricksrc/definitions.csv
```diff
@@ -989,26 +954,7 @@ https://brickschema.org/schema/Brick#Supply_Air_Temperature_Setpoint,Temperature
...
-https://brickschema.org/schema/Brick#Supply_Water,,
-https://brickschema.org/schema/Brick#Supply_Water_Differential_Pressure_Deadband_Setpoint,Sets the size of a deadband of differential pressure of supply water,
-https://brickschema.org/schema/Brick#Supply_Water_Differential_Pressure_Integral_Time_Parameter,,
-https://brickschema.org/schema/Brick#Supply_Water_Differential_Pressure_Proportional_Band_Parameter,,
-https://brickschema.org/schema/Brick#Supply_Water_Flow_Sensor,Measures the rate of flow of hot supply water,
-https://brickschema.org/schema/Brick#Supply_Water_Flow_Setpoint,Sets the flow rate of hot supply water,
-https://brickschema.org/schema/Brick#Supply_Water_Temperature_Alarm,An alarm that indicates the off-normal conditions associated with temperature of the supply water.,
-https://brickschema.org/schema/Brick#Supply_Water_Temperature_Deadband_Setpoint,Sets the size of a deadband of temperature of supply water,
-https://brickschema.org/schema/Brick#Supply_Water_Temperature_Integral_Time_Parameter,,
-https://brickschema.org/schema/Brick#Supply_Water_Temperature_Proportional_Band_Parameter,,
```
jbkoh, line 1010: Where did you add the definition for the new classes?

jbkoh (changes requested):
Skimmed through the changes and look good overall. In addition to the minor comments, two changes that I didn't anticipate are:
- Definitions are mostly removed other than replaced. Is it correct? 
- We previously concluded to keep the deprecated concepts within the hierarchy but not in this code. Is it expected? You said you have some script to generate deprecated concepts, but unsure how you'd implement that without keeping the hierarchy (and they are not in this PR, right?)

bricksrc/rules.ttl
gtfierro: It just fixes a bug where deprecations warnings would fire if you had the deprecation definitions in the graph --- not when you actually instantiated a deprecated concept

bricksrc/definitions.csv
gtfierro: forgot to commit, whoops!

bricksrc/alarm.py
gtfierro: They will be in the deprecation file

gtfierro:
Thanks for the review! All comments have been addressed

PR #417, opened 2022-08-16
Supply_Air != Discharge_Air

hammar:


gtfierro:
We have `Supply Air` == `Discharge Air` throughout the codebase. Does this break an assumption on your end? Specifically, are you treating those two quantities as different, and if so, how do you clarify that?

hammar:
This breaks some scripts that we use to translate Brick substances into DTDL Enums. Nothing that can't be fixed. I rather assumed that this was a bug, since all other occurences of supply/discharge in substances.py lack such equivalence, as do all other subclasses of Air; so it seemed a superfluous/strange declaration to use in a one-off-manner like this. But if this is intended and expected behaviour, feel free to close this PR and I'll work around it on our end instead.

hammar:
It seems I messed this PR up by adding the Absolute_Humidity_Sensor commit to it. I'll try and disentangle the two..

PR #418, opened 2022-08-17
Fix sh:or/sh:property

ektrah:
#299 fixed some properties in [BrickEntityShapeBase.ttl](../blob/master/shacl/BrickEntityShapeBase.ttl) to look like this:

https://github.com/BrickSchema/Brick/blob/c15ba7fa4a7690b8f1c35dead065c9d1c3d5f733/shacl/BrickEntityShapeBase.ttl#L55-L71

It seems the intention here was to create a choice between 3 properties. However, if you look closely at it, the list actually contains 6 elements (the first/third/fifth element being `sh:property` and the second/fourth/sixth element being the blank nodes), so this seems to be invalid SHACL.

I think the actual intention is to have 1 property with a choice between 3 classes. There is [an example](https://www.w3.org/TR/shacl/#OrConstraintComponent) in the SHACL spec for this case that looks like this:

```
ex:PersonAddressShape
	a sh:NodeShape ;
	sh:targetClass ex:Person ;
	sh:property [
		sh:path ex:address ;
		sh:or (
			[
				sh:datatype xsd:string ;
			]
			[
				sh:class ex:Address ;
			]
		)
	] .
```

So this PR fixes the "fix" in #299 by reverting it 😄

### DIFF
--- a/shacl/BrickEntityShapeBase.ttl
+++ b/shacl/BrickEntityShapeBase.ttl
@@ -52,40 +52,28 @@ brick:Equipment a sh:NodeShape ;
         sh:class brick:Equipment;
         sh:message "A piece of Equipment's parts should be always other Equipment."
     ];
-    sh:or (
-        sh:property [
-            sh:path brick:isPartOf;
-            sh:class brick:System;
-            sh:message "Equipment can be part of a Collection, System or other Equipment"
-        ]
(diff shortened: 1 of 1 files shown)


gtfierro:
Thanks for the fix!

PR #420, opened 2022-08-20
Remove duplicate shackles for brick:Location

ektrah:
[BrickEntityShapeBase.ttl](../blob/master/shacl/BrickEntityShapeBase.ttl) currently seems to have duplicate information about `brick:Location`:

https://github.com/BrickSchema/Brick/blob/c15ba7fa4a7690b8f1c35dead065c9d1c3d5f733/shacl/BrickEntityShapeBase.ttl#L11-L39

https://github.com/BrickSchema/Brick/blob/c15ba7fa4a7690b8f1c35dead065c9d1c3d5f733/shacl/BrickEntityShapeBase.ttl#L152-L170

This PR removes the latter.

(Should one of them have been about `brick:Measurable`?)

### DIFF
--- a/shacl/BrickEntityShapeBase.ttl
+++ b/shacl/BrickEntityShapeBase.ttl
@@ -9,6 +9,7 @@
 
 
 brick:Location a sh:NodeShape ;
+    sh:message "Location is an exclusive top class." ;
     sh:and (
         [sh:not [ sh:class brick:Point ] ; sh:message "Location is an exclusive top class." ]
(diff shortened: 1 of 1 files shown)


gtfierro (changes requested):
Thanks for the fix!

Can you make sure the `sh:message` is preserved in the nodeshape? 

Like so: https://github.com/BrickSchema/Brick/blob/c14d327f975d679f7b5c4009d1bea7f5be122415/shacl/BrickEntityShapeBase.ttl#L42

Issue #415, opened 2022-08-12
Add parent property of Brick's inter-node properties


danjhugo:
Proposal: brick's inter-node properties to be defined as "`rdfs:subPropertyOf brick:interNodeProperty`", or similar. 

Explanation... 

Recently, I was composing a query that, among other things, needed to find instance nodes adjacent to a target node and the relationship between them (`?target ?rel ?other`), applying a constraint to drop matches where the object is something other than another brick instance node. While I could easily add the constraint "`?other rdf:type/rdfs:subClassOf* brick:Class`", property path quantifiers can be expensive (total query times in Blazegraph was in the order of 250-300ms which is excessive). 

I got this down to about 100ms by instead defining a parent class for Brick's inter-node properties and constraining for "`?rel rdfs:subPropertyOf brick:interNodeProperty`" in my query. (I also get the same times with "`FILTER( ?linkOut = brick:hasPart || ?linkOut = brick:isPartOf` ... etc", but that just looks clumsy). 

I know this is a bit of a corner case and doesn't change semantics, but it might be useful to include for the sake of query simplicity & performance? 

```
brick:interNodeProperty a owl:ObjectProperty ; skos:definition "Abstract superproperty for properties that relate two instance nodes".
brick:hasPart       rdfs:subPropertyOf brick:interNodeProperty .
brick:isPartOf      rdfs:subPropertyOf brick:interNodeProperty .
brick:feeds         rdfs:subPropertyOf brick:interNodeProperty .
brick:feedsAir      rdfs:subPropertyOf brick:interNodeProperty .
brick:isFedBy       rdfs:subPropertyOf brick:interNodeProperty .
brick:hasPoint      rdfs:subPropertyOf brick:interNodeProperty .
brick:isPointOf     rdfs:subPropertyOf brick:interNodeProperty .
brick:hasLocation   rdfs:subPropertyOf brick:interNodeProperty .
brick:isLocationOf  rdfs:subPropertyOf brick:interNodeProperty .
```

gtfierro:
[...]

PR #423, opened 2022-08-29
Add brick:Relationship superproperty

gtfierro:
Rationale in discussion of #415 . I am adding this not for query performance but so that we have a parallel structure to `brick:Entity` (the root class of all Brick classes) for relationships. This *only* is for inter-entity relationships, so EntityProperties are not included under the `brick:Relationship` property hierarchy

This should be merged into a 1.3.1 should the change be accepted.


ektrah:
Two quick questions:

* What's the relationship between the new `brick:Relationship` and the existing `brick:EntityProperty`?
* Would it make sense to call it `brick:Property` instead of `brick:Relationship`?

gtfierro:
> What's the relationship between the new brick:Relationship and the existing brick:EntityProperty?

Entity Properties would be sibling to the proposed Relationship; Relationships are properties whose value is another "entity" (instance of Brick class). Entity Properties are static attributes of Brick entities.

> Would it make sense to call it brick:Property instead of brick:Relationship?

I'd definitely consider that. Only reason I went with Relationship is because that is the word we've historically used in Brick publications to talk about the relationship between Brick entities.

gtfierro:
@jbkoh thoughts?

epaulson:
We've been using 'relationship' for a long, long time. I wouldn't be excited about changing that term.

jbkoh:
I like Relationship and also its distinction from EntityProperty, which is common in property graphs.

tests/test_properties.py
```diff
@@ -0,0 +1,14 @@
+import brickschema
+from brickschema.namespaces import BRICK, RDFS
+import sys
+
+sys.path.append("..")
+from bricksrc.properties import properties  # noqa:E402
```
jbkoh, line 6: Maybe we can change the file name and the object name to relationships from now on to be explicit on the naming? It's not critical but would be nice to have.

jbkoh (changes requested):
The pedantic side of me wants to change the file name from properties.py to relationships.py to be aligned with the new names though it's not critical. I will let you decide it.

tests/test_properties.py
gtfierro: Nice idea! I'll see how easy it is to make that adjustment

Issue #422, opened 2022-08-21
BRICK.value property generation in define_shape_properties


ektrah:
[`define_shape_properties`](https://github.com/BrickSchema/Brick/blob/c15ba7fa4a7690b8f1c35dead065c9d1c3d5f733/generate_brick.py#L441-L546) currently generates multiple BRICK.value properties for certain cases, leading to output like this:

```
bsh:CoolingCapacityShape a owl:Class,
        sh:NodeShape ;
    rdfs:subClassOf bsh:ValueShape ;
    sh:property [ a sh:PropertyShape ;
            sh:in ( unit:TON_FG unit:BTU_IT-PER-HR unit:BTU_TH-PER-HR unit:W ) ;
            sh:minCount 1 ;
            sh:path brick:hasUnit ],
        [ a sh:PropertyShape ;
            sh:minCount 1 ;
            sh:or bsh:NumericValue ;
            sh:path brick:value ],            ←
        [ a sh:PropertyShape ;
            sh:maxCount 1 ;
            sh:minCount 1 ;
            sh:path brick:value ] .           ←
```

Is that intended?

gtfierro:
No, that is not intended! Thanks for noticing; I'll see what I can do to fix this in the compilation script.

PR #424, opened 2022-08-29
Unify EntityProperty property shapes that reference brick:value

gtfierro:
Fixes issue #422 by ensuring that adding property shapes to EntityProperty definitions does not create different property shapes for `brick:value`

### DIFF
--- a/examples/solar_array/solar_array.ttl
+++ b/examples/solar_array/solar_array.ttl
@@ -7,9 +7,9 @@
 site:panel_efficiency  a brick:EfficiencyShape ;
     brick:hasUnit unit:PERCENT ;
     brick:ambientTemperatureOfMeasurement [
-        brick:value "25"^^xsd::double ;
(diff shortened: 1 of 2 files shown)


ektrah:
There are now some lonely blank nodes in Brick.ttl looking like this:

```
[] sh:minCount 1 .
```

Could that be related to this PR?

gtfierro:
Probably is! I thought I found all of those and fixed them...

I'll take a look today


From: ektrah ***@***.***>
Sent: Wednesday, August 31, 2022 11:02:16 AM
Cc: Gabriel Fierro ***@***.***>; State change ***@***.***>
Subject: [External] Re: [BrickSchema/Brick] Unify EntityProperty property shapes that reference brick:value (PR #424)


There are now some lonely blank nodes in Brick.ttl looking like this:

[] sh:minCount 1 .


Could that be related to this PR?

Issue #419, opened 2022-08-19
Additions for electrical metering


danjhugo:
I would like to propose a couple of minor additions for electrical metering purposes...

1. Add a class brick:Electrical_Energy_Sensor, as a subclass of brick:Energy_Sensor.

2. Add "peak" to the list of literals in brick:AggregationShape (i.e. allow "peak" to be the object of a triple with predicate brick:aggregationFunction). 

3. Rename brick:PowerComplexityShape to be brick:ElectricalComplexityShape, and rename brick:PowerFlowShape to be brick:ElectricalFlowShape. 

Rationale for these changes...

1. An Electrical_Energy_Sensor is a mathematical integration of Electrical_Power_Sensor, and is a type that is more specific than brick:Energy_Sensor. It's a type that we have required in our project work, and which we have used internally in our own patches to Brick. 

2. Peak aggregation was a project requirement we encountered, and it seems like a common need others will have. This allows a modeller to set an aggregation over any metering Point, rather than being limited to brick:Peak_* classes.

3. This is just a minor detail about clarity. These shapes are not really specific to 'power' as distinct from energy, but concern electricity in general (whether energy or power). I know renaming classes is discouraged, but these shapes are 'internal' in the sense that they're used by automated validation tools rather than being referenced in models, so I think this refactoring should be safe.

Please let me know what you think!

gtfierro:
1. Sounds good! Easy enough to add
2. We have the `max` aggregation entity property pattern, as exemplified [here](https://github.com/BrickSchema/Brick/blob/master/examples/building_meter/building_meter.ttl#L10-L15). Does this solve your issue?
3. I think this is fair; I can introduce the change in a PR and we can discuss with the broader community

gtfierro:
@danhugo there is already an `Electric_Power_Sensor` in Brick; be sure to check out the nightly build

gtfierro:
@danjhugo can you leave some feedback on https://github.com/BrickSchema/Brick/pull/425 to see if this fits your needs? Thanks!

PR #425, opened 2022-08-29
Swap "power" for "electrical" in complexity and flow shapes

gtfierro:
Addresses the third point in #419 --- the complexity and flow are related to electricity, not necessarily power (which is a broader concept)


gtfierro:
@epaulson @steveraysteveray what do you think?

steveraysteveray:
Since you asked, ElectricalComplexity (or PowerComplexity) seems a little weird as a term. I agree that we're talking about Electrical Complex Power here, not power in general, not electricity in general and not complexity in general, so why not call them BRICK.electricalComplexPower and BSH:ElectricalComplexPowerShape? Seems clearer.
My 2 cents!

Issue #439, opened 2022-10-04
Light Command vs Lighting Command


jbkoh:
I missed `Light Command` has been added recently, but can't locate the original discussion thread, so I'm starting a new one.

`Light Command` sounds more like a command to a piece of Equipment, `Light` (e.g., `Fan Command`), while `Lighting Command` sounds more related to the lighting action (e.g., `Cooling Command`). I see that they are similar to each other and probably most people won't distinguish them, but for us, Brick, it is a consistency issue.

- If we want it to be a command to a device, we'd want to have the device type to be materialized, but we don't have `Light` as a type but only `Luminaire`. (regardless of the conclusion of this discussion, `Light` as a type may be more aligned to control concepts than `Luminaire`.)
- Lighting controllers often control multiple lights altogether, while `Light_Command` seems more for a single light.

Thus, I think `Lighting_Command` makes more sense and is generalizable.

1.3.0 release is about to be done, so I don't mind reflecting this in the next version, but figured it's worth discussing now.

jbkoh:
@OskarNSE Maybe you are interested in this topic.

OskarNSE:
Yes, thanks for pinging me! I'll look into it

gtfierro:
I don't recall any conversation around `Light Command` but I'm ok with changing to `Lighting_Command`. Can you get a PR done this weekend?

jbkoh:
Closing this and continue discussing at https://github.com/BrickSchema/Brick/pull/459

PR #442, opened 2022-10-08
Light Command to Lighting Command

jbkoh:
Resolving #439


gtfierro:
Hope you don't mind but I just fixed the CSV file to have the trailing comma

jbkoh:
Not at all. Thanks!

jbkoh:
Sure, I'm fine with it.

cbupp:
@jbkoh @gtfierro - reading through this discussion and the PR, I'm tempted to propose an alternative naming. I work with a team that sends many different commands to our fixtures. Light level (or sometime called dim level) is just one of those commands. We also send color change commands, sense commands (used during commissioning), and in some cases we send a command to alter the driver's modulation.

An alternative naming could be `Light_Level_Command`?

cbupp:
Here's a related issue: [Add ColorTemperatureLightingCommand #445](https://github.com/BrickSchema/Brick/issues/445)

jbkoh:
@cbupp Thanks for brining that up! I added a comment to the other thread, but Lighting Command presumed to change the levels but I think your idea of being more explicit is better. I will create a separate PR and close this one then.

gtfierro:
@jbkoh is the plan still to close this PR and continue the development discussion?

jbkoh:
Yes, let me close this one.

PR #452, opened 2022-11-23
add ict devices

jbkoh:
Adopting ICT devices from REC v3 with some modification
- Adapted them to Brick naming convention of Pascal_Snake_Case.
- Excluded Thermostat_Equipment as we already have Thermostat.
- LeakDetectorEquipment to Leak_Detector as the name already represents it's a device.

A possibly controversial topic: Sensor_Equipment already represents all the sensing devices, so do we need all the subclasses? E.g., CO2_Sensor_Equipment, Speed_Sensor_Equipment, etc? (maybe, if we want to standardize some properties of each sensor device types.)


gtfierro:
Thanks for handling this!

One approach we could take for the `CO2_Sensor_Equipment`, etc is to make those SHACL shapes w/ required components:

```ttl
brick:CO2_Sensor_Equipment a sh:NodeShape, owl:Class ;
  refs:subClassOf brick:Sensor_Equipment ;
  sh:property [
    sh:path brick:hasPoint ;
    sh:qualifiedValueShape [ sh:class brick:CO2_Sensor ];
    sh:qualifiedMinCount 1 ;
  ] ;
.
```

How many other ICT devices are there in the REC ontology? Did you port all of them over?

@hammar what do you think of the proposed changes?

hammar:
Love that this is being done, we've had asks in REC to increase the coverage of ICT and sensor equipment that we ingest from Brick.

Ideally, we'd love if all the old equipment from REC3 (pre-Brick) were covered. See https://doc.realestatecore.io/3.3/asset.html#ICTEquipment for the top-level class, which includes subclasses for, e.g., wifi access points, server racks, sensors, ethernet switches and ports, etc. 

I'd be happy to contribute if there's anything I can do here -- but don't want to step on anyone's toes either, so only if you want me to :-)

jbkoh:
I added all the sensor equipment classes as listed [here](https://github.com/Azure/opendigitaltwins-building/tree/master/Ontology/Asset/Equipment/ICTEquipment/SensorEquipment) (except Thermostat which we already have.)

I would like to discuss the proposed shapes in a separate thread because I would like to discuss removing CO2_Sensor. As we discussed before, CO2 sensor is ambiguous to CO2_Level_Sensor while CO2_Sensor alone is not much useful other than being a placeholder of CO2-related classes, which can be represented as hasSubstance instead. 

And my reservation was about class explosion (Equipment class per substance or measurement type?), but I think we can handle case by case by checking whether such devices actually exist (e.g., whether dedicated CO2 sensors do exist or not, which exist).

Another point of discussion I forgot to bring up earlier was that sensor "devices" are more common than sensor "equipment". Is there a reason it was originally called equipment (other than to show it's a subclass of Equipment)? @hammar

hammar:
@jbkoh Yes, that name including "Equipment" was used primarily for consistency in the hierarchy, and to ensure that at-a-glance the semantics of the types were obvious from the name. I'm sure "Device" would work also.

epaulson:
At one point I thought we might have discussed having some kind of more generic 'device' that would be the 'ICT device' for sensors and then attach sensors to that - e.g. the white box that houses the CO2 sensor and boots up and gets an IP address from DHCP and has an asset inventory sticker is the 'device' in the model, and then the actual CO2 sensor (which is spitting out telemetry events that are getting logged into a historian database) is a separate node in the graph. This leads naturally to multi-sensor devices too. 

In such cases, we wouldn't have that many classes. 

In that case we still have as our main query pattern you look for the sensor, but if you care (and if the modeler cared to add it) you could find the box that is housing the sensor.  

Not sure if the world would prefer to just know to look for 'CO2_Sensor_Device' rather than a CO2_Sensor_Device/CO2_Sensor

(Also +1 to deprecating CO2_Sensor and being more specific about what kind of sensor we actually mean)

gtfierro:
One point of clarification: `CO2_Sensor` is the parent class of `CO2_Differential_Sensor` and `CO2_Level_Sensor` in addition to a few others. How would you recommend clarifying this set of concepts?

epaulson:
(Let's kick CO2 sensor hierarchy clarification to a new issue, sorry, Jason already suggested that and I cheated and left it in this thread)

jbkoh:
Leaving notes from our last week discussion. 
- Many of us preferred having a Sensor_Device as a class for all the sensor related physical devices as Points are more exact way of modeling measurement types and the type won't matter too much at device/equipment-level. And Sensor_Device can also accommodate multi-sensing devices like Erik mentioned.
    - If we care the users on REC's older version, I'd propose introducing the above detailed classes but with a deprecate flag. That way, existing users can still use them, but new users will stick to Sensor_Device.
- I had some reservation on Sensor_Device as a subclass of ICT_Device, but realized that makes sense after some googling.
- We haven't made a decision on whether we call Device or Equipment for ICT devices. I still prefer calling them Devices, so that I can call "a device" or "devices" other than "a piece" or "pieces" of equipment, but maybe it's only my thing. Please express your opinions here if you prefer one way or the other.

gtfierro:
> We haven't made a decision on whether we call Device or Equipment for ICT devices. I still prefer calling them Devices, so that I can call "a device" or "devices" other than "a piece" or "pieces" of equipment, but maybe it's only my thing. Please express your opinions here if you prefer one way or the other.

One idea: what if we have `Device` subclass for microprocessor-based entities, similar to how [Haystack does](https://project-haystack.org/doc/appendix/device). We can maybe specify it as `ICT_Device` to be more precise, but `Device` is nice because it already aligns with the existing Haystack taxonomy

jbkoh:
I'm fine with ICT_Device but not just Device as it's such a generic term. It's sad that Haystack went that route already. A snippet from wiki:
> A VAV terminal unit,[8] often called a VAV box, is the zone-level flow control **device**.

If the above items are in agreement, I will update the PR soon.

gtfierro:
What wiki?

jbkoh:
I meant Wikipedia

hammar:
A little late to the party, but "ICT_Device" works for me as well.

hammar:
See also work done by Peter Hartlev in REC, to bring more of our old REC3 concepts into a (delayed) REC4 release: https://github.com/RealEstateCore/rec/pull/229

That is based on REC needs. I would ideally prefer minimal duplication of these things between our two ontologies, so the more of those things we can get into Brick, the happier I'd be. But it depends on your timeframes, we're starting to feel a little stressed about shipping :-)

jbkoh:
I added the definition of [ICT Device](https://github.com/BrickSchema/Brick/pull/452/files#diff-4c8d8a4a5addf35ab5e64925595c0f46184f700d3dca0d180c52a30e34923b06R542) as 
> Information and Communications Technology (ICT) devices operate with a processor to process data or logic and create digital signals.

Please provide any feedback. @gtfierro  @epaulson  @hammar

epaulson:
I'm fine with the definition Jason has for 'ICT Equipment' - it feels maybe a bit abstract but I guess top-level classes might. 

I would note that the use of 'Device' is going the opposite way from what REC is doing in their 229 PR that Karl linked to earlier - they're leaning into 'Equipment'. The changes in 229 are more extensive than ours but I think do mostly line up with our longer-term vision, so I wouldn't mind trying to stay more consistent with that. (I wish their 'controllers' and 'gateway' were under the DataNetwork part of the tree) and maybe trying to take on a bigger chunk of the ICT tree sooner. If that should be a separate PR, that's fine.

I poked some around IFC  (with Dion's handy https://blenderbim.org/search-ifc-class.html ) and they've got a lot of "IT" things under 'CommunicationsAppliance'

hammar:
> I'm fine with the definition Jason has for 'ICT Equipment' - it feels maybe a bit abstract but I guess top-level classes might.
> 
> I would note that the use of 'Device' is going the opposite way from what REC is doing in their 229 PR that Karl linked to earlier - they're leaning into 'Equipment'. The changes in 229 are more extensive than ours but I think do mostly line up with our longer-term vision, so I wouldn't mind trying to stay more consistent with that. (I wish their 'controllers' and 'gateway' were under the DataNetwork part of the tree) and maybe trying to take on a bigger chunk of the ICT tree sooner. If that should be a separate PR, that's fine.

I think REC is fine with aligning on most of this re. naming. The important bit is not so much class names as domain coverage. But since we'd really like to ship soon, and need more than this PR includes, we went and copied our own older REC 3.3 design, including the names from there. Might it be an option to take that design wholesale, but adapting the names as discussed in this thread? Ping @PeteHart

jbkoh:
Hi @PeteHart and @hammar do you have any other thoughts on this? I'd love to merge this minimal PR and add more specific items in a following PR. Thanks and let me know your thoughts.

hammar:
> Hi @PeteHart and @hammar do you have any other thoughts on this? I'd love to merge this minimal PR and add more specific items in a following PR. Thanks and let me know your thoughts.

I'd like to go further as indicated, but if that is in later PRs, then I've no further comments right now. Just a slight nudging against (now merged) https://github.com/RealEstateCore/rec/pull/229

jbkoh:
Sorry, this has been sitting here too long. @epaulson  please let me know your thoughts and I will merge this.

- I reverted the change and now we use Equipment as ICT_Equipment and Sensor_Equipment. It's too compliant to the existing vocabularies and it seems I'm outvoted too.
- I added Audio_Visual_Equipment from REC 3.5 as it seems little controversial

The other parts of the REC vocabularies are worth more discussions. 
- We haven't distinguished controllers and actual equipment so far (a VAV vs its controller). I'd like to have a separate discussion on that to make a guideline. 
- Network devices are in a big category, which deserves a separate discussion too.

epaulson:
I'm fine with this as-is. It's a fine start, and we can move on to the subissues we identified in the thread.

epaulson (commented):
LGTM

gtfierro (approved):
I agree with Erik -- let's get this merged in its current form

jbkoh:
Yay!

PR #456, opened 2022-12-08
Update nightly-builds.yml

epaulson:
Nightly builds started failing a couple of days ago because our python build is out of date. We use 3.8.13 on the build script, it looks like 3.8.16 is the newest point release, but I am just matching what we use in our other tests.


gtfierro:
Thanks! Makes me wonder if we are ok just using an annotation like '3.8' instead of the full version

PR #467, opened 2023-01-20
fix/compare_versions_tool

filiphl:
The version comparison tool did not work for since the ontology URLs are formatted differently since when it was created. This fixes that.


gtfierro:
Thanks for the PR! I'll be sure to test it out to ensure it works on some development versions of Brick

gtfierro (approved):
Looks great. Thanks for the PR!

Issue #472, opened 2023-01-24
Use of bsh:RDFSSubPropertyOfRule


steveraysteveray:
This rule takes all triples involving a property and duplicates those triples for its super-property. I'm pretty uncomfortable with this, because, for example, it results in:

```
?s skos:exactMatch ?o                 yielding the following assertions:

?s skos:closeMatch ?o
?s skos:mappingRelation ?o
?s skos:semanticRelation ?o
```
...because skos:exactMatch is an rdfs:subProperty of those other properties.

Here's another:
```
?s skos:altLabel ?o                   results in:
?s rdfs:label ?o
```

I recognize that 
"A property p is a subproperty of property q exactly when:
        p(x,y) implies q(x,y)"

...so the rule isn't incorrect, but I'm bothered by explicitly asserting these new triples. 

Thoughts?

gtfierro:
I wanted to hold off on commenting until I had done a couple tests. The immediate consequence of removing the `bsh:RDFSSubPropertyOfRule` is essentially that certain Entity Properties (e.g., `area` vs `grossArea`) won't be properly added. It may be worth specializing the rule to only target those entity properties rather than arbitrary sub-properties. Then, as we add additional uses of `rdfs:subPropertyOf`, we can expand the target of that SHACL rule.

PR #473, opened 2023-01-25
Limit subproperty inference to entity properties

gtfierro:
Address #472

### DIFF
--- a/bricksrc/rules.ttl
+++ b/bricksrc/rules.ttl
@@ -83,7 +83,7 @@ CONSTRUCT {
     sh:targetSubjectsOf rdfs:range ;
 .
 
-bsh:RDFSSubPropertyOfRule
+bsh:RDFSSubPropertyOfRuleForEntityProperties
     a   sh:NodeShape ;
     sh:rule [
         a sh:SPARQLRule ;
@@ -92,6 +92,8 @@ CONSTRUCT {
     ?s ?super ?o .
 } WHERE {
     $this rdfs:subPropertyOf ?super .
+    ?super a brick:EntityProperty .
+    $this a brick:EntityProperty .
     ?s $this ?o .
 }""" ;
       sh:prefixes <https://brickschema.org/schema/1.3/Brick> ;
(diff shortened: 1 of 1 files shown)


gtfierro:
@jbkoh @epaulson do you have any issues with this fix? @steveraysteveray laid out the issues w/ sub property inference in #472. It's not an issue for Brick, but by making our inference rule more specific, Brick can be a better "ontology citizen" when integrated with other ontologies.

jbkoh:
I'm fine with the changes!

jbkoh (approved):

steveraysteveray:
As long as you don't nest rdfs:subPropertyOf more than 1 level deep this works in a single iteration. Otherwise, you could add a + to the rdfs:subPropertyOf predicate, or assume iterative reasoning.

gtfierro:
Are the defaults semantics for reasoning not fixed point? I'm pretty sure that pySHACL iterates it's rules a number of times until the graph stops changing. If the reasoning is not fixed point by default, then I agree it would be better to use the transitive closure here. Probably more performant anyway, too!
From: steveraysteveray ***@***.***>
Sent: Thursday, January 26, 2023 8:48:29 AM
Cc: Gabriel Fierro ***@***.***>; Author ***@***.***>
Subject: [External] Re: [BrickSchema/Brick] Limit subproperty inference to entity properties (PR #473)

CAUTION: This email originated from outside of the Colorado School of Mines organization. Do not click on links or open attachments unless you recognize the sender and know the content is safe.


As long as you don't nest rdfs:subPropertyOf more than 1 level deep this works in a single iteration. Otherwise, you could add a + to the rdfs:subPropertyOf predicate, or assume iterative reasoning.

steveraysteveray:
Not the default in TBC, although I always set it to iterate. The [SHACL Advanced Features spec](https://w3c.github.io/shacl/shacl-af/#rules-execution) seems to be non-committal one way or the other.

steveraysteveray:
Looks good to me!

PR #476, opened 2023-02-06
Remove Substance punning

gtfierro:
Switch to SKOS hierarchy like we have with brick Quantities


steveraysteveray:
@gtfierro, hope you don't mind I piggy-backed on your PR to fix the Supply_Air declaration, which was getting turned back into a class because of the owl:equivalentClass declaration (via the bsh:RDFSRangeRule).

gtfierro:
@steveraysteveray thanks! I added `owl:sameAs` instead to preserve the discharge==supply semantics that we've used elsewhere in Brick

PR #484, opened 2023-03-06
 Removed inclusion of unit definitions

steveraysteveray:
@gtfierro, assuming this doesn't break something else I'm unaware of, this change reduces the line count of Brick.ttl from 52,044 to 43,490 by suppressing the (redundant) definition of all the QUDT units, since they are now imported.


steveraysteveray:
It seems the checks do indeed notice the absence of the unit definitions. The question is whether we can get the system to see the imports where such things are defined.

gtfierro:
Hi @steveraysteveray -- I just pushed a fix to this branch that uses the `ontoenv` library to import the dependencies into an RDFlib Graph (see `tests/test_examples.py`). This fixes some of the tests--I'll take a look at the others soon but it should be the same fix.

One issue I'm getting is `qudt:Unit-supercededBy` is used as the object of `sh:property` but is never actually defined anywhere. The pySHACL library doesn't like this, so validation fails on graphs that import QUDT because pySHACL wants to know that `qudt:Unit-supercededBy` is actually a PropertyShape. I believe this *should* be inferred from `sh:property`, but I need to double check how the standard is worded

steveraysteveray:
Thanks! I assume the same ontoenv snippet in analyze.py will fix the submeter test failure.

Meanwhile, I'll remove the definition of qudt:Unit-supercededBy. We are using dcterms:isReplacedBy instead, and never cleaned up.

gtfierro:
> Thanks! I assume the same ontoenv snippet in analyze.py will fix the submeter test failure.

It should! If you change the definitions of any of the imported ontologies, remember to run `make` (to update the `imports` folder)

> Also, got the following warning when running "pytest analyze.py"
> /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages/brickschema/graph.py:231: DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
logger.warn(report)

This shouldn't affect any of the tests -- it's something I can address in another PR but it is really just as simple as adding `ing` onto the `warn()` method calls

steveraysteveray:
@gtfierro, I think this PR is good now. The units do not explicitly show up in Brick.ttl.

There is one minor side-effect that has come up because of some refactoring in the QUDT repo. Namely, all Currency units have been moved to a separate graph. They are still instances of qudt:Unit. However, if you do not choose to import the Currency units, then we get a validation error caused by qudqk:Currency referring to a bunch of units that are unknown. If you are OK with importing Currency units, there's no problem, but I'd like to hear your thoughts about this as a sample user of QUDT. If you really don't want to include Currency units, QUDT.org may reconsider how we package the graphs.

gtfierro:
> @gtfierro, I think this PR is good now. The units do not explicitly show up in Brick.ttl.
> 
> There is one minor side-effect that has come up because of some refactoring in the QUDT repo. Namely, all Currency units have been moved to a separate graph. They are still instances of qudt:Unit. However, if you do not choose to import the Currency units, then we get a validation error caused by qudqk:Currency referring to a bunch of units that are unknown. If you are OK with importing Currency units, there's no problem, but I'd like to hear your thoughts about this as a sample user of QUDT. If you really don't want to include Currency units, QUDT.org may reconsider how we package the graphs.

I'm not strictly against including the Currency graph -- it might be helpful for modeling tariff structures for example. It should be ok to include it for now, but it does raise the question of how to bound the size of these graphs. I wonder if there is some sort of [tree shaking](https://webpack.js.org/guides/tree-shaking/) procedure we could use to remove unused parts of the graph (esp when we take imports into account)

steveraysteveray:
OK, great. In that case, merging #495 includes the Currency graph, and includes this merge.

I'll think about "tree shaking"...

PR #488, opened 2023-03-22
Srr skos narrower

steveraysteveray:


gtfierro (approved):

gtfierro:
@gtfierro check the `SKOS.narrower` line in `generate_brick.py` works as expected

gtfierro:
@steveraysteveray the line does work, but it only applied in certain situations; I've added a couple lines to `generate_brick.py` that populate all narrower/broader pairs -- this this ok to do, or should I change the implementation so that I only populate the narrower/broader pairs for concepts defined by Brick?

steveraysteveray:
I think it would be best to populate the skos:narrower and skos:broader triples where the subject is a Brick object. The object can be anything, Brick or otherwise. i.e. Nobody expects to control who points to their objects, but they might not agree with affecting what their objects point to.
I'm saying this because the Brick ontology might be part of a workspace containing various other ontologies, and an application might think that one of the other ontologies (I'm thinking QUDT here) is asserting some relation to Brick that has not been vetted by QUDT.

gtfierro:
@steveraysteveray when you get a chance can you verify that I'm not adding rogue narrower/broader statements to URIs that Brick doesn't own?

steveraysteveray:
@gtfierro, it all looks good!

PR #494, opened 2023-03-27
 Updated QUDT imports

steveraysteveray:


steveraysteveray:
QUDT will be updated again once it is decided whether Brick wants to import the QUDT Currency vocabulary. That vocabulary has now been refactored so that it is not necessary to be imported just to avoid validation errors. 

@gtfierro, what say you?

gtfierro:
@epaulson @jbkoh any opinions on whether or not we include the QUDT currency graph? I figure we might as well include it in the distribution for the sake of completeness. It moderately increases the storage requirements and doesn't seem to affect SHACL reasoning times to any major degree

gtfierro:
> (Do we have a call in the py-brickschema package that pulls in graphs we include in the distro at construction time?)

Not yet, but that's a great idea

PR #522, opened 2023-06-21
Add automatic tint window array

connorjcantrell:
Add new concept of Automatic Tint Window Array an update Tint Status definition to support the broader use of the term.


connorjcantrell:
I will create a new branch with same changes to avoid implications with merging master into this branch

PR #523, opened 2023-06-22
Srr cleanup 223 extension

steveraysteveray:


steveraysteveray:
PR puts in place the option of extending Brick with all of the 223 standard, by importing extensions/brick_extension_s223_extension.shapes.ttl (graph https://brickschema.org/extension/s223extension) into Brick.ttl. It includes being able to partially interpret 223 data files into Brick entities, and expands some Brick inference rules and constraints to allow for 223 concepts.

steveraysteveray:
I put this back to draft until the ref: ontology settles down.

steveraysteveray:
Cleaned up the extensions folder to not rely on imports

PR #524, opened 2023-06-28
Add `Electrical_Energy_Usage_Sensor`

connorjcantrell:
Proposing the addition of `Electrical_Energy_Usage_Sensor` as a subclass of `Energy_Usage_Sensor`


gtfierro (approved):
I like this idea. We can discuss at the weekly meeting to see if there is any feedback

jbkoh:
@gtfierro I meant to comment on this earlier. The concept of "usage" needs to be clear in my opinion like the other usage sensors like Steam Usage Sensor and Gas Usage Sensor and so on. I meant to introduce "Electrical_Energy_Usage_Sensor" but not looking back, it should be just called "Electricity Usage Sensor". Along that line, I haven't seen actual "Thermal_Energy_Usage_Sensors". There is no sensor measuring only thermal energy. One can calculate it though in that case, it'd rather just be called "Thermal_Energy_Sensor".

PR #525, opened 2023-06-28
Add `Automatic_Tint_Window_Array`

connorjcantrell:
Add new concept of Automatic Tint Window Array and update Tint Status definition to support the broader use of the term.


gtfierro (commented):
@connorjcantrell are there any more generic terms that we might nest `Automatic Tint Window Array` under? I'm ok if we discover that down the line, but I figured I would ask just in case

connorjcantrell:
@gtfierro None that come to mind at the moment. This addition was influenced by a real use case, where a single controller manages an array of windows.

PR #527, opened 2023-07-12
Fix Deprecation Statements

gtfierro:
Ensure that the deprecated concepts are not always asserted to be `owl:Class`

### DIFF
--- a/generate_brick.py
+++ b/generate_brick.py
@@ -314,11 +314,11 @@ def define_classes(definitions, parent, pun_classes=False):
         assert isinstance(aliases, list)
         for alias in aliases:
             G.add((classname, OWL.equivalentClass, alias))
-      
+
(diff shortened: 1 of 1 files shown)


gtfierro:
@steveraysteveray I attempted to fix the issue with deprecated concepts where some of the substances were asserted as owl Classes. Can you double check that the generated Brick file [here](https://gist.github.com/gtfierro/d51703811e23c34326ed4e636f4126a2) doesn't exhibit those issues? Thanks! I've spot checked it myself

steveraysteveray:
Will do. Might not be till next week though because I have babysitting duties this week.

gtfierro:
> Will do. Might not be till next week though because I have babysitting duties this week.

No rush! Thanks :)

steveraysteveray:
This seems to have fixed things. All good.

PR #543, opened 2023-08-14
Rename `Solar Radiance Sensor` to `Solar Irradiance Sensor` for Accuracy

connorjcantrell:
Upon reviewing our current `Solar_Radiance_Sensor`, renaming to`Solar_Irradiance_Sensor` might be the most appropriate class name for our context. Here's why:

Solar Irradiance represents the power per unit area received from the Sun. Solar applications are most interested in the total amount of solar power received on a surface area. The unit of measurement is often given in watts per square meter (W/m²).

Considering the above distinctions and the prevalent use of W/m² as a measurement unit, I believe that Solar Irradiance Sensor would be a more accurate and suitable class name.

Thank you


gtfierro (changes requested):
Looks great and I agree with the change! Can you also add a deprecation notice for `Solar_Radiance_Sensor` so that previous uses of it receive a warning message?

gtfierro (approved):
Thanks!

PR #544, opened 2023-08-18
fix a wrong hierarchy

jbkoh:
`Hot_Water_Differential_Temperature_Sensor` should not be a subclass of leaving temperature and entering temperature even if they are related with each other.


gtfierro:
Sounds like a good change to me!

jbkoh:
Thanks. Btw, why does the unit test take forever? 4 hours unit testing is way too long for an individual contributor to iterate development

gtfierro:
That is a great question and a discussion for another time! It essentially comes down to reasoning

gtfierro:
@jbkoh is this PR ready for review?0

PR #549, opened 2023-08-30
 Refreshed 223 and QUDT files

steveraysteveray:


extensions/brick_extension_G36_SP223-v1.0.ttl
```diff
@@ -131,19 +128,18 @@ g36:TwoPositionDamper
...
 .
-
-
 g36:HotWaterCoil
   a s223:Class, rdfs:Class, sh:NodeShape ;
   rdfs:label "HotWater/ReheatCoil" ;
   rdfs:comment "A heating element made of pipe or tube that may or may not be finned and formed into helical or serpentine shape." ;
   sh:class s223:HeatingCoil ;
-  rdfs:subClassOf s223:Equipment;
+  rdfs:subClassOf s223:HeatingCoil;
```
connorjcantrell, line 136: A heating coil may utilize electric heating elements instead of a valve

connorjcantrell: Disregard, I see that Resistance Heater is used for this.

PR #553, opened 2023-09-01
Introduce Base `Tank` class and deprecate `Water_Tank`

connorjcantrell:
This PR introduces a new foundational class, `Tank`, under the 'Equipment' class. This class will serve as the parent class for specialized tanks utilized in HVAC, plumbing, and irrigation systems.

Prior to introducing substance-specific subclasses like Water, we are establishing two key base classes: `Storage_Tank`, `Thermal_Expansion_Tank`, and `Separation_Tank`. The Storage Tank class will serve as the parent for all tanks designed primarily for storing various substances. The Thermal Expansion Tank class is for tanks that are engineered to accommodate volume changes due to temperature/ pressure fluctuations. The `Separation_Tank` class tank used in conjunction with a filter to facilitate the separation of filtrate material for disposal. (I am open for suggestions on renaming Separation Tank, but it seemed necessary parent class for something like a Grease Interceptor.)

By first laying down these general categories, we can define substance-specific subclasses, such as Chilled Water Storage Tank or Hot Water Thermal Expansion Tank.

In alignment with these changes, this PR also deprecates the existing `Water_Tank` class and removes its prior categorization under the 'Space' class. Tanks serve specific functional roles in building systems, be it storage, thermal expansion, or other specialized purposes. Their functional attributes align more closely with equipment rather than spatial entities.


JoelBender:
Here at Cornell we have a [`Chilled_Water_Thermal_Storage_Tank`](https://fcs.cornell.edu/departments/energy-sustainability/utilities/cooling-home/cooling-production-home/thermal-storage-tank), so please include "thermal".

(and as an aside, the acoustics inside this before they filled it up were _amazing_)

connorjcantrell:
@JoelBender When you say "thermal storage tank", are you referring to a thermally insulated tank?

JoelBender:
No, the tank is "charged" by pumping chilled water into the bottom of the tank and draining off an identical amount of warm water off the top.  When the tank is "discharged" the warm chilled water from the campus is pumped into the top of the tank and the chilled water drained off the bottom.  The tank is charged at night from chillers when electricity is the least expensive, and used as peak load shaving during the day to supplement the chillers, perhaps by not having to run multiple chillers to keep up with the load.

There is a thermocline that drifts up and down and the layers of hot and cold water do not mix (much, there is a little).

Chilled water storage is very similar to [ice storage](https://en.wikipedia.org/wiki/Ice_storage_air_conditioning), with the exception that the medium doesn't go through a phase transition.  Ice storage "tanks" are smaller and typically serve a single building.

bricksrc/deprecations.py
```diff
@@ -1,6 +1,11 @@
 from .namespaces import BRICK, RDFS, SKOS, A
 
 deprecations = {
+    BRICK.Water_Tank: {
+        "version": "1.3.0",
+        "mitigation_message": "'Water_Tank' is deprecated as a space. 'Water_Storage_Tank' serves as it's replacement and is being moved to the equipment class.",
+        "replace_with": BRICK.Water_Storage_Tank,
+    },
```
gtfierro, line 8: Not deprecating yet. Instead, change the definition of Water Tank to emphasize it is a space and we might move/get rid of it later

connorjcantrell: The current definition is, "A space used to hold water". Would you like me to update the definition to include that it may be deprecated in the future?

gtfierro: I can do it! Thanks

PR #557, opened 2023-09-05
Fix ref-schema imports

gtfierro:
Handle `sh:prefixes` when importing the ref-schema

### DIFF
--- a/bricksrc/ontology.py
+++ b/bricksrc/ontology.py
@@ -78,11 +78,11 @@ shacl_namespace_declarations = [
     },
 ]
 
+BRICK_IRI_VERSION = URIRef(f"https://brickschema.org/schema/{BRICK_VERSION}/Brick")
 
 def define_ontology(G):
-    brick_iri_version = URIRef(f"https://brickschema.org/schema/{BRICK_VERSION}/Brick")
-    G.add((brick_iri_version, RDF.type, OWL.Ontology))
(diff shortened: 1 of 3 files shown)


steveraysteveray:
The one thing that seems to have been dropped is the ontology declaration. 
This causes (me) problems with imports, since there is no name for the graph. Easy enough to fix, right?

In Brick.ttl I see:

```
<https://brickschema.org/schema/Brick> sh:declare [ sh:namespace "https://brickschema.org/schema/Brick#"^^xsd:anyURI ;
            sh:prefix "brick" ],
        [ sh:namespace "http://www.w3.org/ns/shacl#"^^xsd:anyURI ;
            sh:prefix "sh" ],
        [ sh:namespace "http://data.ashrae.org/standard223#"^^xsd:anyURI ;
            sh:prefix "s223" ],
        [ sh:namespace "http://www.w3.org/2002/07/owl#"^^xsd:anyURI ;
            sh:prefix "owl" ],
        [ sh:namespace "https://brickschema.org/schema/Brick/ref#"^^xsd:anyURI ;
            sh:prefix "ref" ],
        [ sh:namespace "http://www.w3.org/1999/02/22-rdf-syntax-ns#"^^xsd:anyURI ;
            sh:prefix "rdf" ],
        [ sh:namespace "http://www.w3.org/2000/01/rdf-schema#"^^xsd:anyURI ;
            sh:prefix "rdfs" ] .
```

steveraysteveray (approved):

PR #560, opened 2023-09-06
Add Solar Inverter

connorjcantrell:
Proposing to add Solar Inverter, which converts direct current electricity generated by solar panels into alternating current.


gtfierro (commented):
`s/solar/photovoltaic/` and then we are good to merge!

gtfierro (approved):
Thanks!

PR #561, opened 2023-09-06
Add reheat and ability to represent dehumidification process

connorjcantrell:
Proposing to add commands, sensor, and equipment classes to support the dehumidification process in the Brick ontology. Specifically, we are introducing the following:

- Reheat Command (`Reheat_Command`)
- Reheat Supply Air Temperature (SAT) Sensor (`Reheat_SAT_Sensor`)
- Reheat Hot Water Valve (`Reheat_Hot_Water_Valve`)
- Dehumidification Command (`Dehumidification_Command`)

### Reheat Command (`Reheat_Command`)
The reheat command is needed for distinguishing the heating command needed to reheat the supply air post-dehumidification. Without a dedicated reheat command, it isn't possible to decouple the reheat operation from other heating operations within an HVAC system.

### Reheat Hot Water Valve (`Reheat_Hot_Water_Valve`)
The Reheat Hot Water Valve controls the flow of hot water to the heating elements used in the reheat process. This tag allows for control and monitoring of the hot water valve.

### Reheat SAT Sensor (`Reheat_SAT_Sensor`)
The Reheat SAT Sensor will be responsible for monitoring the temperature of the supply air post-reheat. This sensor is necessary for implementing feedback loops that can adjust the reheat command or reheat hot water valve.

### Dehumidification Command (`Dehumidification_Command`)
Dehumidification Command is used to initiate the control sequence using the introduced concepts listed above.

### DIFF
--- a/bricksrc/command.py
+++ b/bricksrc/command.py
@@ -21,6 +21,7 @@ command_definitions = {
             "Cooling_Command": {"tags": [TAG.Point, TAG.Cool, TAG.Command]},
             "Heating_Command": {"tags": [TAG.Point, TAG.Heat, TAG.Command]},
             "Preheat_Command": {"tags": [TAG.Point, TAG.Preheat, TAG.Command]},
+            "Reheat_Command": {"tags": [TAG.Point, TAG.Reheat, TAG.Command]},
             "Luminance_Command": {"tags": [TAG.Point, TAG.Luminance, TAG.Command]},
             "Bypass_Command": {"tags": [TAG.Point, TAG.Bypass, TAG.Command]},
             "Damper_Command": {
@@ -35,6 +36,10 @@ command_definitions = {
             },
             "Humidify_Command": {
                 "tags": [TAG.Point, TAG.Humidify, TAG.Command],
+                BRICK.hasQuantity: BRICK.Humidity,
+            },
+            "Dehumidification_Command": {
+                "tags": [TAG.Point, TAG.Dehumidification, TAG.Command],
                 BRICK.hasQuantity: BRICK.Humidity,
             },
             "Position_Command": {
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -728,6 +728,16 @@ valve_subclasses = {
                                     TAG.Equipment,
                                 ]
                             },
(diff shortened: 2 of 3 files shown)


connorjcantrell:
I accidentally added Reheat SAT sensor. I corrected it with "Cooling Coil Supply Air Temperature Sensor"

gtfierro:
This looks good for the most part, but I want to talk a little more about the Cooling Coil SAT Sensor. In Brick we are trying to get away from having the names of the equipment inside the point names. If the SAT sensor is immediately after the coil, then we should find a way of representing that explicitly within the model, e.g. by relating it to an instance of the cooling coil.

Doing this properly means we'll have to put a Sensor Equipment right after the Cooling Coil. Maybe the way we capture this concept is with a shape?

```ttl
bsh:Cooling_Coil_Supply_Air_Temperature_Sensor a sh:NodeShape, owl:Class ;
  sh:class brick:Supply_Air_Temperature_Sensor ;
  sh:property [
    sh:qualifiedMinCount 1 ; sh:qualifiedMaxCount 1 ;
    sh:path brick:isPointOf ;
    sh:qualifiedValueShape [ sh:node bsh:CoolingCoilDownstreamSensor ]
] 
  sh:rule [
    a sh:TripleRule ;
    sh:subject sh:this ;
    sh:predicate rdf:type ;
    sh:object brick:Supply_Air_Temperature_Sensor ;
 ]
.

bsh:CoolingCoilDownstreamSensor a sh:NodeShape, owl:Class ;
  sh:class brick:Sensor_Equipment ;
  sh:property [
    sh:qualifiedMinCount 1 ; sh:qualifiedMaxCount 1 ;
    sh:path brick:isFedBy ;
    sh:qualifiedValueShape [ sh:class brick:Cooling_Coil ]
] ;
.
```

In your model, you could instantiate it as follows:

```ttl
:ahu a brick:AHU ;
  brick:hasPart :cc, :se .

:cc a brick:Cooling_Coil ;
  brick:feeds :se .

:se a brick:Sensor_Equipment ;
  brick:hasPoint :sat .

:sat a bsh:Cooling_Coil_Supply_Air_Temperature_Sensor.
```

What do you think? The CCSAT Sensor has a rule which adds the correct Brick point class to the instance

PR #565, opened 2023-09-20
Add level command and status

connorjcantrell:
Proposing to add "Level Command" and "Level Status" to compliment Brick's "On Off Command" and "On Off Status".
Brick defines "On Off Status" as a generic binary status of a control loop, relay or equipment. In addition, I propose we support a generic analog status which indicates the operational state of a specific level within a predefined range.

For instance, consider a dimmable smart light bulb. With "Level Status," you could monitor the current brightness level of the bulb within a predefined range, say from 0% (completely off) to 100% (fully bright). "Level_Command" gives us the ability to represent a command that sets the light's brightness to 50%.


gtfierro:
Before I consider merging this we should spend some more time at the Brick meeting tomorrow talking about commands (https://github.com/BrickSchema/Brick/issues/566)

jbkoh:
This wasn't related to #566 as it's more about how to represent percentage command. Erik mentioned how to represent the semantics of the raw data could be a topic for another day, but having a dedicated class for this would be necessary regardless of how we'd model the values and the values' semantics.

jbkoh:
Still, I have to point out that there is a complication that a level command may be a numeric command or a discrete command with state texts. I believe the current PR is assuming a numeric command. In the past, Brick has been liberal on the possible data types for a specific class. If so, a level command may be used as a percentage or a state command. As I develop programming interfaces over classes, I'd prefer having a single data type per class but that'd be ugly like Level_Numeric_Command and Level_State_Command. I'm fine without those distinction at Brick side, but I think we want to have "level" as generic classes.

gtfierro (approved):

Issue #532, opened 2023-07-30
add Gauge Pressure


MatthewSteen:
The ontology appears to be missing `gauge pressure` (zero-referenced against ambient air pressure), which would be helpful for ASHRAE Guideline 36 water-side points and fault conditions. Suggest adding the following.

## Pressure_Sensor
- Gauge_Pressure_Sensor
  - Chilled_Water_Gauge_Pressure_Sensor
  - Hot_Water_Gauge_Pressure_Sensor

PR #569, opened 2023-09-27
Add gauge pressure sensors

gtfierro:
Resolves https://github.com/BrickSchema/Brick/issues/532

@MatthewSteen does gauge pressure use the same units as static pressure?


MatthewSteen:
No, static pressure (e.g. for fans) is measured in `in wc (inches water column)` and gauge pressure is measured in `psig (lb/in2 gauge)`. 

### Static Pressure
image

### Gauge Pressure
image

steveraysteveray:
Just in case you are interested, in QUDT the first would be [unit:IN_H2O](https://qudt.org/vocab/unit/IN_H2O) and the second is [unit:LB_F-PER-IN2](http://qudt.org/vocab/unit/LB_F-PER-IN2). Note that the second unit is distinct from the pound of **_mass_** per square inch, [unit:LB-PER-IN2](http://qudt.org/vocab/unit/LB-PER-IN2).

PR #572, opened 2023-09-28
Add zone damper

connorjcantrell:
zone-bypass

### Add Zone Damper
A zone damper works by opening or closing, either fully or partially, to allow or restrict airflow. When the damper is open, conditioned air flows into the zone. Zone dampers are controlled by a zone thermostat or a centralized control system. When a zone's temperature deviates from the setpoint, the corresponding zone damper will adjust to either increase or decrease the airflow until the desired temperature is achieved. 

### Add Bypass Damper
It's not always the case that all zones need conditioned air simultaneously. A potential problem arises when one or more zones are not calling for conditioned air and their zone dampers are closed. If the produced air has nowhere to go because all the zone dampers are closed or mostly closed, it can create excess pressure in the supply duct.

This is where a bypass damper comes into play in a zoning system.

A bypass damper acts as a pressure relief system. When excess air pressure builds up in the ducts (due to one or more zone dampers being closed), the bypass damper opens to allow this excess air to circulate back to the return. Once the pressure stabilizes, the bypass damper will close.


gtfierro:
I will merge the latest from master and then merge this provided the tests pass

connorjcantrell:
Zone Damper made it into a later PR

PR #578, opened 2023-10-20
Add Switchgear classes

wcrd:
Adding some additional Electrical equipment classes to cover off some major components found in switch boards/panels/etc. Originally had these classes under a `Switch` class, but have update to use the already established electrical `Switchgear` class.

Not an electrical expert, but have tried to model these classes as best I can based on schedules, schematics, manuals, and BACnet discoveries I have seen. I think a second opinion would be good here.

I have also made a change to an existing class `Disconnect Switch`:
- I have renamed this to `Building_Disconnect_Switch`
- I have subclassed this under a new `Disconnect_Switch` class, which I have given a new definition.

_**Note:**_ It looks like some auto-formatting stuff for other class definitions was also caught up in this. Let me know if OK.

## Hierarchy
- `existing` Switchgear
  - Automatic_Switch
    - Automatic_Transfer_Switch
    - Static_Transfer_Switch
  - Circuit_Breaker
    - Main_Circuit_Breaker
  - `existing` Disconnect_Switch -> `renamed & moved`: Building_Disconnect_Switch
    - Main_Disconnect_Switch
    - Building_Disconnect_Switch
  - Isolation_Switch
  - Transfer_Switch
    - Automatic_Transfer_Switch
    - Static_Transfer_Switch

### Automatic Switch
An automatic switch operates automatically in the event of some current threshold or other designed event. Criteria for automatic operation is generally a physical property of the switch.

### Automatic Transfer Switch
An automatic transfer switch (ATS) is a device that automatically transfers a power supply from its primary source to a backup source when it senses a failure or outage in the primary source.

### Circuit Breaker
A circuit breaker is a safety device to prevent damage to devices in a circuit, such as electric motors, and wiring when the current flowing through the electrical circuit supersedes its design limits. It does this by removing the current from a circuit when an unsafe condition arises. Unlike a switch, a circuit breaker automatically does this and shuts off the power immediately﻿.

### Main Circuit Breaker
All breaker panels generally have a main circuit breaker before the bus / MCBs. In some older panels there may be a Main Disconnect Switch instead. It is also possible to have a main disconnect switch, and a main circuit breaker in a panel.

### Disconnect Switch
A disconnect switch performs the task of manually cutting or reconnecting power from an electrical supply by creating or closing an air insulation gap between two conduction points. Also known as an 'Isolation Switch'

### Main Disconnect Switch
Main manual switch on a breaker at the incoming feed.

### Isolation Switch
See Disconnect Switch

### Transfer Switch
An electrical switch that switches a load between two sources

### Static Transfer Switch
Similar to ATS, but utilises no moving parts in the switch to achieve much faster response times


connorjcantrell:
Hey @wcrd, 

Before coming across your PR, I was going to propose renaming "Breaker Panel" to "Electrical Panelboard". Panelboards would allow us to effectively represent various configurations and components, such an MCB or MLO panelboard. Curious to get your thoughts on this. Thanks

gtfierro (approved):

PR #579, opened 2023-10-25
 Updated QUDT to v2.1.32

steveraysteveray:


gtfierro:
Thanks, @steveraysteveray ! I think the rest of these are mostly Warning/Info errors now, but all the Violations are gone? I want to prioritize some other PRs first but I will eventually get back to those. One issue I'm running into is the fixed point inference in the TopQuadrant SHACL implementation isn't working the way I expect, but I haven't pinned down the issue yet.

steveraysteveray:
All violations are indeed gone.

Issue #581, opened 2023-10-30
Need ability to represent VRF systems


connorjcantrell:
Opening a new issue to bring attention to a previously discussed topic: https://github.com/BrickSchema/Brick/pull/344

### Fan Coil Unit/ Indoor Unit
The term "Indoor Unit" is equivalent to Brick's existing "Fan Coil Unit". We can also improve our ability to represent the various types of Indoor Units/ Fan Coil Units by defining the following:

- Wall [Mounted] Fan Coil Unit
![Screenshot 2023-10-30 at 11 20 22 AM](https://github.com/BrickSchema/Brick/assets/33592545/513f0810-d863-42e4-be59-585d53e85269)
- Ceiling [Mounted] Fan Coil Unit
![Screenshot 2023-10-30 at 11 17 01 AM](https://github.com/BrickSchema/Brick/assets/33592545/dff6b98b-4a3b-46a9-9625-503dd1f4e007)
- Concealed/ Duct Fan Coil Unit
![Screenshot 2023-10-30 at 11 18 53 AM](https://github.com/BrickSchema/Brick/assets/33592545/86cd0f2d-95f1-471c-b9bf-88954df72334)
- Floor [Mounted] Fan Coil Unit
![Screenshot 2023-10-30 at 11 17 10 AM](https://github.com/BrickSchema/Brick/assets/33592545/fe3e9a72-5664-4f91-8fa8-9380ebc72062)

### Condensing Unit/ Outdoor Unit
The term "Outdoor Unit" is also referred to as a "Condensing Unit", which needs to be added to Brick. Condensing Unit is not to be confused with Condenser, which is a core component of a Condensing Unit along with a compressor and, in some cases, a reversing valve.

We can categorize Condensing Units as follows:

- Heat Pump Condensing Unit
- Cooling Only Condensing Unit

### Branch Selector
It is worth mentioning that up to this point, all concepts can be used interchangeably with conventional cooling and heating systems and VRF systems. The distinction lies in the presence of a "Branch Selector" between the FCUs and Condensing Units. Therefore, Brick will also need to introduce a Branch Selector Box class.

----
If "Outdoor Unit" and "Indoor Unit" must be represented, we should make these an alias to "Condensing Unit" and "FCU" respectively.  In summary, I propose introducing:

- Condensing Unit
- Branch Selector
- Wall Fan Coil Unit
- Ceiling Fan Coil Unit
- Floor Fan Coil Unit
- Duct Fan Coil Unit

[...]

PR #344, opened 2021-10-23
update vrf classes

jli113:
Adding the following entities:
https://brickschema.org/schema/Brick#Hydrocarbon_Refrigerant,"Hydrocarbon Refrigerants are natural, nontoxic refrigerants that have no ozone depleting properties and absolutely minimal global warming potential. The most efficient and environmentally safe refrigerants in the world are the five natural refrigerants which are Air, Water, Carbon Dioxide, Ammonia and Hydrocarbons.",https://www.engas.com.au/about/about-hydrocarbon-refrigerants/
https://brickschema.org/schema/Brick#Fluorocarbon_Refrigerant,"Although they can be found in nature, the gases used are industrial gases, just like HFCs & HFOs, and are produced in refineries or other industrial facilities.",
https://brickschema.org/schema/Brick#CFC,"Chlorofluorocarbons (CFCs): when derived from methane and ethane these compounds have the formulae CClmF4−m and C2ClmF6−m, where m is nonzero. More than 98% of the CFCs used in MDIs have already been phased out, and will disappear completely in 2016, a significant milestone following 30 years of concerted global action to protect the ozone layer.",
https://brickschema.org/schema/Brick#HCFC,"Hydro-chlorofluorocarbons (HCFCs): when derived from methane and ethane these compounds have the formula CClmFnH4−m−n and C2ClxFyH6−x−y, where m, n, x, and y are nonzero. New production and import of most HCFCs were phased out as of 2020. The most common HCFC in use today is HCFC-22 or R-22, a refrigerant still used in existing air conditioners and refrigeration equipment. ",
https://brickschema.org/schema/Brick#HFC,"Hydrofluorocarbons (HFCs): when derived from methane, ethane, propane, and butane, these compounds have the respective formulae CFmH4−m, C2FmH6−m, C3FmH8−m, and C4FmH10−m, where m is nonzero. An HFC phase-down under the Montreal Protocol was agreed in Kigali, Rwanda, in October 2016. Developed countries will phase-down their production and imports of HFCs by 85% between 2019 and 2036. Developing countries will also phase-down their HFC production and imports.",
https://brickschema.org/schema/Brick#HFO,"Hydrofluoroolefins (HFOs) are unsaturated organic compounds composed of hydrogen, fluorine and carbon.",https://en.wikipedia.org/wiki/Hydrofluoroolefin
[...]

PR #590, opened 2023-12-05
Modeling VRFs and Heat Pumps

connorjcantrell:
Improve representation of VRF systems and Heat Pumps, drawing from #581 and integrating concepts from previous PRs: 
- Update VRF classes #344
- Heat Pump / Chiller Plant Model #257


bricksrc/equipment.py
```diff
@@ -558,6 +565,42 @@
             },
         },
     },
+    "Heat_Pump": {
```
connorjcantrell, line 568: @gtfierro How could we model a heat pump which is a split system? These definitions work well for a packaged unit, such as the illustrated "Air_Source_Heat_Pump":
image

However, a split system would have both a Condensing Unit and an FCU. This, too, is a Heat Pump:
image

connorjcantrell: In some sense, a split system is more of a "Heat Pump System", where as a packaged unit is a "Heat Pump Equipment"

connorjcantrell: We do have "Heat_Pump_Condensing_Unit", perhaps "Heat_Pump" class could be modified to "Heat_Pump_Packaged_Unit" for clarification.

connorjcantrell:
@gtfierro @jbkoh While while "Expansion Valves" and "Metering Devices" are often used interchangeably, a clearer distinction is beneficial. Expansion valves, such as the Thermostatic Expansion Valve (TXV), encompass functionalities beyond traditional valves. A TXV, for instance, adjusts refrigerant flow automatically, guided by a sensing bulb that reacts to temperature changes. 
Screenshot 2023-12-12 at 3 11 25 PM

Similarly, capillary tubes, another form of metering devices, control refrigerant flow through a fixed orifice and do so without moving parts, relying on the pressure differential.
image

With that being said, Electronic Expansion Valves (EXVs) are closer to traditional valves. Being as though these are designed specifically for metering refrigerant and often containing much more than just a valve, I would prefer to classify this outside of the Valve class, as "Metering Devices" more accurately reflects their unique functionalities in HVAC systems.

Alternatively, we could define a Metering Device class in addition to Expansion Valves. Expansion Valve could be a subclass to both Valve and Metering Device. Capillary Tube could exclusively be a subclass to Metering Device, whereas TXV and EXV can be a subclass of both Metering Device and Expansion Valve.

bricksrc/relationships.py
```diff
@@ -5,6 +5,10 @@
 Defining Brick relationships
 """
 relationships = {
+    "connectedTo": {
```
connorjcantrell, line 8: @gtfierro Is it necessary to specify that 'connectedTo' is a superclass to 'feeds' and 'isFedBy'?

Something like:
```
rdfs:subPropertyOf :feeds, :isFedBy ;
```

bricksrc/equipment.py
```diff
@@ -742,19 +880,41 @@
         },
         "parents": [BRICK.Water_Heater],
     },
-    "Metering_Device": {
-        "tags": [TAG.Refrigerant, TAG.Gas, TAG.Liquid, TAG.Metering, TAG.Expansion, TAG.Valve],
+    "Throttling_Device": {
```
connorjcantrell, line 883: Though I do prefer to use Metering Device, I have renamed "Metering Device" to "Throttling Device" to avoid confusion with Brick's [Meter ](https://brickschema.org/ontology/1.3/classes/Meter/) class. Throttling device can refer to any component that restricts or regulates the flow of a fluid. It's not specific to refrigeration systems and can apply to various types of mechanical and fluid systems. Open to feedback on this, as I would prefer to use Metering Device if we didn't have a naming conflict.

bricksrc/relationships.py
gtfierro: In my intended design, `feeds` and `isFedBy` would both be subproperties of `connectedTo`, thinking about it more. I'm not sure this is what we actually want to do. If we were to follow this design, then any use of the `feeds` relationship would also imply the existence of a `connectedTo` relationship.

This raises the essential question, which is "does `connectedTo` communicate a bidirectional relationship, or a relationship where the directionality is not known? If its the former, then `connectedTo` should be a sibling of `feeds` and `isFedBy` because it is altogether different. If its the latter, then `connectedTo` should be a superproperty of `feeds` and `isFedBy`

bricksrc/equipment.py
gtfierro: I'm not familiar with the area, but this substitution seems fine for me. We could also go "Metering Device" and then add shackle rules to either Meter class or this class that detect when it's being used incorrectly.

MatthewSteen: I'm not either, but in general I always recommend/want terms to be rooted in a standard (ASHRAE, ISO, etc.). Does `Throttling Device` come from somewhere? 

I see `Throttling Valve` in ASHRAE terminology (although it's not a standard nor are their standards required to use these). 

[Throttling Valve](https://terminology.ashrae.org/?term=throttling)
> valve used to restrict (throttle) the flow of fluid.

MatthewSteen:
A general comment here. Does it make sense for this PR to focus on just VRFs rather than trying to complete the larger/more complex Heat Pump issue?

bricksrc/equipment.py
MatthewSteen: Is the existing `Condenser` not adequate? Perhaps we need a `Condenser_Unit` class and a corresponding `Evaporator_Unit` class.

connorjcantrell:
Thanks @MatthewSteen,

>  Does it make sense for this PR to focus on just VRFs rather than trying to complete the larger/more complex Heat Pump issue?
Yes, I am addressing both VRFs and Heat Pumps in this PR.

Regarding the alignment with Haystack's VRF concepts, here are the equivalent terms from Haystack the current proposal.

vrf-outdoorUnit -> [Condensing_Unit](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L759)
vrf-outdoorUnit-coolingOnly -> [Cooling_Only_Condensing_Unit](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L762)
vrf-outdoorUnit-heatPump -> [Heat_Pump_Condensing_Unit](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L771C14-L771C39)
vrf-outdoorUnit-heatRecovery -> [Heat_Recovery_Condensing_Unit](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L781C14-L781C43)
vrf-indoorUnit -> [Fan_Coil_Unit](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L388C14-L388C27)
vrf-indoorUnit-fcu -> also [Fan_Coil_Unit](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L388C14-L388C27)
branchSelector -> [Branch_Selector](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L371)
vrf-refrig-plant: Haystacks defines this as "Multiple outdoor units can be connected, and the refrigerant plant covers all the connected outdoor units". In Brick, this can be achieved by connecting multiple Condensing Unit's under the same [VRF_System](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/collections.py#L58)

Within the [Haystack thread](https://project-haystack.org/forum/topic/916), Leroy Simms stated the following:
> I am in favor of all the general concept presented, my only concern would be the indoorUnit and outdoorUnit tags. In my opinion the outdoorUnit tag could be misleading in certain situations.

I, [along with others](https://github.com/BrickSchema/Brick/issues/581#issuecomment-1787589621),  agree with this perspective, and have chosen not to introduce Outdoor and Indoor for this reason.

Please let me know if you have any concerns with the implementation/ alignment. Thanks!

bricksrc/equipment.py
connorjcantrell: @gtfierro Ok, great to hear. In that case, I would prefer to call it "Metering Device". @MatthewSteen I was careful not to use the name "valve", due to the various metering devices used in practice, though I do agree that Throttling Device is a bit broad. If you're curious to learn more about metering devices, this is an informative article:
https://www.skillcatapp.com/post/a-complete-guide-to-metering-devices-fixed-modulating.

With the addition of metering device, Brick now has the ability to represent a complete refrigeration cycle using the following classes:
- Compressor
- Reversing Valve (if Heat Pump)
- Metering Device
- Evaporator Coil
- Condenser Coil

refrigerant-flow copy

bricksrc/equipment.py
connorjcantrell: @MatthewSteen In the split-system example above, this would be achieved by using the [Heat_Pump_Condensing_Unit](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L771C14-L771C39) > feeds > [ FCU](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L388)

In this case, do you agree that [Heat_Pump](https://github.com/mapped/Brick/blob/8e73f532db499395462747edf05fbd209895f378/bricksrc/equipment.py#L672) should be "Packaged Heat Pump"?

https://www.trane.com/residential/en/products/packaged-systems/heat-pump-packaged-systems/

bricksrc/equipment.py
MatthewSteen: I strongly suggest basing the terminology on existing definitions/terms that are as close to canonical as possible so we're not "inventing" our own or basing them on a website. There are a lot of interchangeable terms used in industry but I think Brick should be using "authoritative" sources when it comes to its class names whenever possible. 

I may have missed it, but what's the problem with using `Throttling_Valve` per ASHRAE terms instead of `Metering_Device`, which seems a bit general for this application (VRF). Seems like the following hierarchy might work...

Valve
- Reversing Valve
- Throttling Valve

bricksrc/equipment.py
connorjcantrell: Renamed "Heat Pump" to "Packaged_Heat_Pump" and added air/water/ground source to all condensing unit classes. In doing so, we can represent the same level of detail for both Packaged and Split System variations.
https://github.com/BrickSchema/Brick/pull/590/commits/cbcc5d47c49d9110478339c72ea59c3630ad008a

bricksrc/equipment.py
connorjcantrell: Agreed that reusing ASHRAE terms is ideal. Between the two, I would prefer "Refrigerant Metering Device".

Throttling_Valve is also a good alternative. However, Throttling Valve would only be able to represent [modulating metering devices](https://www.skillcatapp.com/post/a-complete-guide-to-metering-devices-fixed-modulating#viewer-3cgvk) and would not be able to represent [fixed metering devices](https://www.skillcatapp.com/post/a-complete-guide-to-metering-devices-fixed-modulating#viewer-vdi1).

MatthewSteen: I think we should be basing Brick's class names on an authoritative/canonical source (such as AHRI, ANSI, ASHRAE, ISO, IEE, etc.), *not* https://www.skillcatapp.com/.

connorjcantrell: 100% agreed, I shared the article with you to explain the differences between fixed and modulating metering devices and how throttling valve does not encompass all metering devices.

I have renamed the class to refrigerant metering device per ASHRAE's definition:
https://terminology.ashrae.org/?entry=refrigerant%20metering%20device

bricksrc/relationships.py
gtfierro: We probably want both:
- "feedsAndFedBy" (**we need a better name**) for bidirectional. We want this to work for equipment as well as for spaces/locations
- "connectedTo" for a super property of feeds, isFedby, and feedsAndFedBy that means "we don't know the direction"

PR #597, opened 2023-12-13
Add meter as subclass to sensor equipment

connorjcantrell:
Meter is a type of sensor equipment, adding Meter as a subclass to Sensor Equipment.

### DIFF
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -19,6 +19,7 @@ equipment_subclasses = {
             },
             "Sensor_Equipment": {
                 "tags": [TAG.Sensor, TAG.Equipment, TAG.ICT],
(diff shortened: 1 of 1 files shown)


gtfierro:
There was pushback on this during the meeting, based on the widely-held interpretation that Meters do _not_ fall under ICT equipment, i.e. are not managed as part of the building's ICT infrastructure. One could argue that Meters are Sensing Equipment, though this feels awkward.

One question this exposes is whether we want to differentiate between sensing equipment that is part of the ICT infra (e.g. things that have an IP, participate in DHCP, etc) and sensing equipment that is part of the OT infra (e.g. the air flow station in an AHU). Currently, our SensorEquip class is "just" the ICT flavor, which means we don't have an answer for sensing equipment that is part of OT.

PR #602, opened 2023-12-18
Add booster and circulating pump

connorjcantrell:
Proposing the inclusion of booster and circulator pumps in the pump topology to clearly differentiate and highlight their specific functions and roles.


connorjcantrell:
It may also be worth considering adding "condensate pump" and "sump pump".

gtfierro (approved):

gtfierro:
Feel free to add them in a future PR!

PR #603, opened 2023-12-18
Fix IndexError in `add_definitions` by checking list length

connorjcantrell:
Identified and fixed a test failure caused by an `IndexError` during the graph population process involving the 'seeAlso' definition. Additionally, to resolve `flake8` formatting conflicts with certain SPARQL queries and warning messages, I've implemented `# noqa` comments to selectively bypass these specific checks.

### DIFF
--- a/generate_brick.py
+++ b/generate_brick.py
@@ -682,12 +682,15 @@ def add_definitions(graph=G):
     with open("./bricksrc/definitions.csv", encoding="utf-8") as dictionary_file:
         dictionary = csv.reader(dictionary_file)
 
-        # skip the header
-        next(dictionary)
+        header = next(dictionary)
 
         # add definitions, citations to the graph
         for definition in dictionary:
             term = URIRef(definition[0])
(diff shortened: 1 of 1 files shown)


generate_brick.py
```diff
@@ -682,12 +682,15 @@ def add_definitions(graph=G):
...
         dictionary = csv.reader(dictionary_file)
 
-        # skip the header
-        next(dictionary)
+        header = next(dictionary)
 
         # add definitions, citations to the graph
         for definition in dictionary:
             term = URIRef(definition[0])
+            if len(definition) > len(header):
```
connorjcantrell, line 690: It appears some row's definitions are not wrapped in parentheses and using commas within the definition, which is causing parsing issues. This new condition should catch most of them.

gtfierro: There was one bad row which you can fix in definitions.csv (I think it is the Tablet row)

gtfierro (approved):
Thanks!

PR #613, opened 2024-01-17
V1.4.0 Release Candidate 1

gtfierro:


steveraysteveray:
@gtfierro, I was going to test this branch, but found some errors running generate_brick.py (I believe during the validation phase):
```
2024-01-18:10:09:46,199 ERROR   [__init__.py:350] Could not resolve https://brickschema.org/schema/1.3/Brick (No definition for https://brickschema.org/schema/1.3/Brick: https://brickschema.org/schema/1.3/Brick:1:0: not well-formed (invalid token))
```
This seems odd because the base uri is https://brickschema.org/schema/1.4-rc1/Brick. Not sure how this reference crept in.

Moving on, it did create brick.ttl, so I loaded it in TBC. However, TBC would not execute the reasoner because it complained that rdf:type had an undefined prefix (presumably inside a SPARQLRule, but I checked them all manually and they seem good).

I should note that the master branch does run in TBC, albeit it creates many new triples in the bacnet namespace... all due to this rule:
```
bsh:RDFSRangeRule a sh:NodeShape ;
    sh:rule [ a sh:SPARQLRule ;
            sh:construct """
CONSTRUCT {
    ?val a ?shape .
} WHERE {
    $this rdfs:range ?shape .
    ?shape a sh:NodeShape .
    ?ent $this ?val .
}""" ;
            sh:prefixes <https://brickschema.org/schema/1.4-rc1/Brick> ] ;
    sh:targetSubjectsOf rdfs:range .
```

It has been a while...

gtfierro:
Thanks! I'll take a closer look at that rule and see if I can determine why that other issue with the namespace is showing up

PR #631, opened 2024-04-12
Srr qudt 2 1 37

steveraysteveray:
@gtfierro, here is what the QUDT imports graph should look like after the changes I suggested in email are done on your end.

image


steveraysteveray:
Interestingly, this implementation surfaced a couple of minor errors in the QUDT schema, which have already been fixed for the next QUDT release. I'm still puzzling why they weren't flagged in other environments. The error messages are generated by tosh via the brick->rec->dash->tosh import, but even explicitly importing tosh into either the 223 workspace or the QUDT native workspace doesn't surface the errors (even though they are valid errors that needed correcting).

image

Regardless, I'm grateful that the validation in Brick is finding them!

gtfierro:
Interesting! It seems like there are so many paths for ontologies to be included and for rules to be run that it can be hard to get reproducible results.

Issue #608, opened 2024-01-05
Redefining Setpoint Classifications: Target and Threshold


connorjcantrell:
How do we define setpoint? I personally like [Wikipedia's](https://en.wikipedia.org/wiki/Setpoint_(control_system)) definition:
> The desired or target value for an essential variable, or process value (PV) of a control system which may differ from the actual measured value of the variable.

Let's call this a "Target Based Setpoint"

### Target-Based
These are values that is the desired target for a process variable (PV). Examples include:
- _Temperature Setpoint_: A value specifying the desired temperature to be maintained.
- _Humidity Setpoint_: A value specifying the desired humidity to be maintained.
- _Differential Setpoint_: A value specifying the desired difference between two points in a system.

____

However, there are many terms used to describe different types of setpoints, such as:
- Cutoff Setpoint
- Reset Setpoint
- Enable Setpoint

These setpoints are an attempt to describe something else.. and it doesn't fit into the definition mentioned above. So how do we distinguish these types of setpoints? I would like to propose a term, "Threshold Setpoint".

### Threshold-Based
These values initiates a specific action when a process variable crosses a certain threshold. They can be further categorized into upper and lower thresholds:
#### Upper Thresholds
Initiates a specific action when the process variable exceeds specified value.
- _Cooling Temperature Setpoint_: Activates cooling when temperature exceeds this value.
- _Dehumidification Setpoint_: Activates dehumidification when relative humidity exceeds this value.
- _CO2 setpoint_: Activates ventilation or other CO2 mitigation actions when CO2 concentration exceeds this value.
#### Lower Thresholds
Initiates a specific action when the process variable falls below specified value.
- _Heating Temperature Setpoint_: Activates heating when temperature falls below this value.
- _Humidification Setpoint_: Activates humidification when relative humidity falls below this value.

___

### Example
![237245173-4f704b44-a4a1-4de6-8125-2941bbf88b4e](https://github.com/BrickSchema/Brick/assets/33592545/702baf55-55e9-449d-8089-553192251a50)

Let's compare "Effective Cooling/Heating Setpoint" to "Effective Target Setpoint" as an example:

- **Effective Cooling Setpoint**: Serves as an "Upper Threshold", which initiates the cooling process
- **Effective Heating Setpoint**: Serves as a "Lower Threshold", which initiates the heating control process 
- **Effective Target Setpoint**: Serves as a Target Setpoint, serving as a desired temperature to be maintained.

### Key Takeaway
[...]

PR #637, opened 2024-05-30
Add target and threshold setpoints

connorjcantrell:
This PR is based on [Option 1](https://github.com/BrickSchema/Brick/issues/608#issuecomment-2109032894) of issue  https://github.com/BrickSchema/Brick/issues/608.

In addition to existing class hierarchies, setpoints are now a subclass of either Target Setpoint or Upper/Lower Threshold Setpoint.

### DIFF
--- a/bricksrc/setpoint.py
+++ b/bricksrc/setpoint.py
@@ -8,6 +8,13 @@ setpoint_definitions = {
         ),
         "tags": [TAG.Point, TAG.Setpoint],
         "subclasses": {
+            "Target_Setpoint": {},
+            "Threshold_Setpoint": {
+                "subclasses": {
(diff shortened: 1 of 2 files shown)


bricksrc/setpoint.py
```diff
@@ -27,10 +34,16 @@
             "Enthalpy_Setpoint": {
```
connorjcantrell, line 34: @gtfierro Is Enthalpy_Setpoint intended to be used as an enthalpy to be maintained (target), or an enthalpy to enable something such as an economizer (lower/upper threshold)?

gtfierro: I think I have mostly seen it as a target

PR #641, opened 2024-06-18
Move fume hood under exhaust fan class

connorjcantrell:
Proposing to classify `brick:Fume_Hood` as a type of `brick:Exhaust_Fan`. 

A fume hood is a type of local exhaust ventilation device designed to protect users from exposure to hazardous fumes, vapors, and dust. It is typically mounted over a workspace, table, or shelf to capture and conduct unwanted gases away from the enclosed area.


connorjcantrell:
ASHRAE's definition implies the exhaust function in the phrase "_serving to conduct unwanted gases away from the area enclosed_", though it does not explicitly state that it exhausts air from an enclosed space.:
> a fume-collection device mounted over a work space, table, or shelf and serving to conduct unwanted gases away from the area enclosed. The least effective fume hood is a canopy hood, open on four sides. The most effective fume hood effective is an enclosed hood with operable front sash and fixed sides and back

gtfierro (approved):
Looks good to me! I'm changing the base branch to v1.4.1 so this can be part of that release

Issue #673, opened 2024-10-24
rec:ICTEquipment vs brick:ICT_Equipment, deprecate rec:ICTEquipment?


PeteHart:
@gtfierro 

As Brick has cloned the ICT equipment super class with its subclasses it'd logical to deprecate it on the REC side. However, it appears that the properties of the REC definitions are missing!

Best Peter

gtfierro:
Thanks for letting us know! I will make a PR to add these soon

PeteHart:
Thanks @gtfierro

PR #681, opened 2025-01-17
Add missing properties from rec.ttl

metesaka:
This commit targets to solve #673


gtfierro:
Looks like a great start! Please do the same for *all* of the ICT equipment classes in REC (you should be able to find them at https://dev.realestatecore.io/ontology/), and make sure the properties themselves are also defined in the recpatches.ttl file

metesaka:
I added missing properties for all children (and grandchildren) of ict_equipment and also fixed some deprecation messages. However I did not understand what you mean by defining the properties, they're not explicitly defined under rec.ttl either.

gtfierro (approved):
Looks great, thank you!

Issue #682, opened 2025-01-24
Adding resolution as an Entity Property


lazlop:
Resolution is a common property of sensors and thermostats, that can have different values (e.g. 0.1°F, 0.5°F, 1°F). I can see it being useful to relate to equipment (primarily thermostats) as well as points.

gtfierro:
@metesaka You want to take a stab at this? Look at `bricksrc/entity_properties.py`

gtfierro:
@lazlop @metesaka does this want to be on equipment? [SSN/SOSA's related concept](https://www.w3.org/TR/vocab-ssn/#SSNSYSTEMResolution) places this on the Property, which would suggest the Point should own this rather than Equipment. What do you think?

gtfierro:
This will be available  in the next nightly release!

PR #683, opened 2025-01-30
add resolution property

metesaka:
This commit targets to solve #682


gtfierro:
Can you rewrite the condition on the "comment artifact" step to be `      if: ${{ github.event_name == 'pull_request' && matrix.python-version == env.RELEASE_PYTHON_VERSION && github.event.pull_request.head.repo.full_name == 'BrickSchema/Brick' }}`? This should sidestep the permissions issue

gtfierro (approved):

PR #703, opened 2025-06-25
Fix line endings

ektrah:
This changes `.gitattributes` so that all line endings in textual files are converted to the platform-specific line ending (CRLF on Windows, LF on Linux and macOS) on checkout and to LF on commit. The only affected file is `bricksrc/definitions.csv`, which was committed with CRLF line endings. (No changes are made to this file except for line endings changes.)

This should fix the huge diff in #702 once that is rebased.


gtfierro:
Awesome! I'll merge this as soon as the tests pass

PR #717, opened 2025-08-21
Update tooling along with new API usage

gtfierro:
From #710 changes (breaking this up)

- Updated GitHub Actions workflow to use updated checkout and uv setup actions.
- Simplified Python setup, replacing dependency installation with the `uv` tool.
- use updated compile./dependency managemetn from ontoenv and py-brickschema


gtfierro:
This fails, but only the tests that wil lbe fixed by #718

PR #721, opened 2025-08-29
feat: introduce specific relationships for Controller: controls, concerns and hosts

fennibay:
Currently [brick:Controller](https://ontology.brickschema.org/brick/Controller.html) doesn't have any specific relationships defined, it just inherits its relationships from brick:Equipment.

I propose introducing some Controller-specific relationships that a Building Automation Controller would have, because of the control logic it contains:

- controls: A controller can reside outside of the equipment (e.g. control cabinet). Even when it is part of an equipment, it is not straightforward to come to the conclusion which part of the equipment it is controlling. So a controls relationship from brick:Controller to brick:Equipment would be useful.
- concerns: A controller controls an equipment that influences (feeds) a rec:Space (we will use rec:Zone specifically). It would be useful to shorten this path and introduce a concerns relationship from brick:Controller to rec:Space (alternatively rec:Zone).
- hosts: A collection of entities can be hosted on a Controller (e.g. sensors, zones, light groups etc.), the hosts relationship from brick:Controller to brick:Collection would allow this. This is also a preparation for future introduction of functions.

I am already providing this draft-PR to demonstrate the idea concretely. I'm open for any other proposals.

/cc @ektrah

### DIFF
--- a/bricksrc/relationships.py
+++ b/bricksrc/relationships.py
@@ -162,6 +162,48 @@ relationships = {
         "domain": BRICK.Point,
         RDFS.label: Literal("Has unit", lang="en"),
     },
+    "controls": {
+        A: [OWL.ObjectProperty, OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+        OWL.inverseOf: BRICK["isControlledBy"],
+        "range": BRICK.Equipment,
+        "domain": BRICK.Controller,
+        RDFS.label: Literal("Controls", lang="en"),
+    },
+    "isControlledBy": {
+        A: [OWL.ObjectProperty, OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+        OWL.inverseOf: BRICK["controls"],
+        "range": BRICK.Controller,
+        "domain": BRICK.Equipment,
+        RDFS.label: Literal("Is controlled by", lang="en"),
+    },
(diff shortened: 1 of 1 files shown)


gtfierro:
arch-2

Thanks for opening the issue! We had a discussion today and I captured some of this above in the diagram.

The `controls` relationship makes sense, and there was no push back from the group on this.

The `concerns` relationship also makes sense. We wondered if it could be inferred to simplify modeling. For example, does the following always hold?
```
?controller controls ?equip .
?equip feeds ?zone .
=> implies ?controller concerns ?zone
```

We had the most discussion on the `hosts` relationship. We imagined two ways of doing this modeling: first, `Controller hosts Point` could directly relate a point to a controller; I like the simplicity of the model, but it doesn't capture nested point groups. Alternatively, as you've suggested here, we could create a Collection of Points which is hosted by a controller. I imagine these could be recursively nested too!

```
:C1 hosts :pointgroup1, :pointgroup2 .
:pointgroup1 hasMember :pointgroup3 .
:pointgroup2 hasMember :SAT1, :RAT1 .
:pointgroup3 hasMember :MixedAirTemp1 .
```

As an aside, this also raises the question if we want to re-adopt the `hasMember` relation from REC to handle collection membership.

Is this what you had in mind?

Issue #722, opened 2025-08-29
Collections as part of other entities


fennibay:
Today [rec:Collection](https://ontology.brickschema.org/rec/Collection.html) (its subclass [brick:Collection](https://ontology.brickschema.org/brick/Collection.html)) can only be `isPartOf` Collections and no other entities (e.g. rec:Space, brick:Equipment).

We have certain scenarios where we would like to logically group entities in Collections, where:

- they do not represent a meaningful domain entity such as an equipment, zone etc.; but just a grouping of underlying entities
- they are not at top-level, but within a given context (rec:Space, brick:Equipment).

Some examples:

- CdnMsgCol: Is a collection of condensation message points within a [Supply Air VAV](https://sid.siemens.com/r/A6V13561255/30648920203__V6.0_en-US_7206694027).
- Separating different disciplines within a room: HVAC, Lighting, Shading. See RHvacCoo, RLgtCoo, RShdCoo at <https://nas.desigo.tw/NGA/Help/en/en-US/10454554635.html>
- Organizational "folders" in building automation projects for example for separating room automation from primary automation plants within a building.

So, I propose to extend the properties
- hasPart: allow Collection in the range
- isPartOf: allow Collection in the domain.

This might be better addressed at <https://github.com/RealEstateCore/rec>. Please advise.

See also: #671 

/cc @ektrah

PR #724, opened 2025-09-25
Controller and Point Group model

gtfierro:
Building on #721

### DIFF
--- a/bricksrc/equipment.py
+++ b/bricksrc/equipment.py
@@ -57,6 +57,7 @@ equipment_subclasses = {
             },
             "Controller": {
                 "tags": [TAG.Equipment, TAG.ICT, TAG.Controller],
+                BRICK.hosts: [BRICK.Point_Group],
                 "subclasses": {
                     "BACnet_Controller": {
                         "tags": [TAG.Equipment, TAG.BACnet, TAG.Controller],
--- a/bricksrc/relationships.py
+++ b/bricksrc/relationships.py
@@ -179,7 +179,7 @@ relationships = {
     "hosts": {
         A: [OWL.ObjectProperty, OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
         OWL.inverseOf: BRICK["isHostedBy"],
-        "range": REC.Collection,
+        "range": BRICK.Point_Group,
         "domain": BRICK.Controller,
         RDFS.label: Literal("Hosts", lang="en"),
     },
@@ -187,7 +187,7 @@ relationships = {
         A: [OWL.ObjectProperty, OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
         OWL.inverseOf: BRICK["hosts"],
         "range": BRICK.Controller,
-        "domain": REC.Collection,
+        "domain": BRICK.Point_Group,
         RDFS.label: Literal("Is hosted by", lang="en"),
     },
     "concerns": {
--- a/bricksrc/collections.py
+++ b/bricksrc/collections.py
@@ -1,4 +1,4 @@
-from .namespaces import TAG, OWL, BRICK
+from .namespaces import TAG, REC, BRICK
 
 system_subclasses = {
     "Automatic_Tint_Window_Array": {
@@ -158,6 +158,10 @@ collection_classes = {
     "Portfolio": {
         "tags": [TAG.Collection, TAG.Portfolio],
         "constraints": {BRICK.hasPart: [BRICK.Site]},
+    },
+    "Point_Group": {
+        "tags": [TAG.Collection, TAG.Point_Group],
+        "constraints": {REC.includes: [BRICK.Point]},
     },
     "System": {
         "tags": [TAG.Collection, TAG.System],
--- a/bricksrc/rules.ttl
+++ b/bricksrc/rules.ttl
@@ -556,3 +556,23 @@ bsh:InheritEVSEChargerDirection a sh:NodeShape ;
         """ ;
     ] ;
 .
+
+# check that controllers host the points of the equipment they control
+bsh:CheckControllerHostsPointGroups a sh:NodeShape ;
+    sh:targetClass brick:Controller ;
+    sh:sparql [
+        a sh:SPARQLConstraint ;
+        sh:message "Controllers should host the points of the equipment they control" ;
+        sh:prefixes <https://brickschema.org/schema/1.4/Brick> ;
+        sh:select """
+        SELECT $this ?pt ?eq WHERE {
+            $this brick:controls ?eq .
+            ?eq brick:hasPoint ?pt .
+            FILTER NOT EXISTS {
+                $this brick:hosts ?pg .
+                ?pg rec:includes ?pt .
+            }
+        }
+        """ ;
+    ]
+.
--- a/examples/controller/controller.ttl
+++ b/examples/controller/controller.ttl
@@ -1,0 +1,40 @@
+@prefix bldg: <http://example.com/controller#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix rec: <https://w3id.org/rec#> .
+@prefix brick: <https://brickschema.org/schema/Brick#> .
+@prefix bacnet: <http://data.ashrae.org/bacnet/2020#> .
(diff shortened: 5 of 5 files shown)


fennibay:
Thank you @gtfierro.

`concerns` and `controls` look good.

Now I see we missed one point about `hosts`. These hosted Collections need to contain not only Points and PointGroups, but also equipments. See #721: "A collection of entities can be hosted on a Controller (e.g. sensors, zones, light groups etc.), the hosts relationship from brick:Controller to brick:Collection would allow this. " #722 is also related to this.

Below some first ideas on how to address this, we shall discuss tomorrow:

1. Extend PointGroup to also contain Equipment?
2. Go back to Controller hosts Collection, no PointGroup?
3. Merge this PR without `hosts` and look at it with #722?
4. Merge this PR AS IS, and look at the equipment in Collection with #722?

Below another example assuming option 2 and #722 are in, just to give some concrete context.

```ttl
@prefix bldg: <http://example.com/controller2#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@prefix rec: <https://w3id.org/rec#> .
@prefix bacnet: <http://data.ashrae.org/bacnet/2020#> .
@prefix unit: <http://qudt.org/vocab/unit/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix ref: <https://brickschema.org/schema/Brick/ref#> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .

<http://example.com/controller2> a owl:Ontology ;
    owl:imports <https://brickschema.org/schema/1.4/Brick> .


bldg:Controller_1 a brick:Controller ;
    rdfs:label "AS20" ;
    brick:hosts bldg:HVAC1 ;
    brick:controls bldg:VavSu1 ;
    brick:concerns bldg:R009 ;
.

bldg:HVAC1 a brick:Collection ;
    brick:hasPart bldg:VavSu1 ;
    brick:hasPart bldg:HCl1 ;
.

bldg:VavSu1 a brick:Variable_Air_Volume_Box ;
    rdfs:label "VAV Supply Unit 1" ;
    brick:hasPart bldg:CdnMsgCol1 ;
.

bldg:CdnMsgCol1 a brick:Collection ;
    rdfs:label "Condensation Message Collection 1" ;
    brick:hasPart bldg:CdnMsgRs1 ;
.

bldg:HCl1 a brick:Heating_Coil ;
    rdfs:label "Heating Coil 1" ;
    brick:hasPoint bldg:Pos1 ;
.

bldg:CdnMsgRs1 a brick:Status ;
    rdfs:label "Condensation Message Result 1" ;
.

bldg:Pos1 a brick:Valve_PositionCommand ;
    rdfs:label "Heating Coil Position 1" ;
.

bldg:R009 a rec:Zone ;
    rdfs:label "Room 009 Zone" ;
    brick:hasPoint bldg:TR1 ;
    brick:hasPart bldg:SpTRDTr1 ;
.

bldg:TR1 a brick:Air_Temperature_Sensor ;
    rdfs:label "Room Temperature 1" ;
.

bldg:SpTRDTr1 a brick:Collection ;
    rdfs:label "Room Temperature Setpoint Determination 1" ;
    brick:hasPart bldg:SpTR1 ;
.

bldg:SpTR1 a brick:Air_Temperature_Setpoint ;
    rdfs:label "Room Temperature Setpoint 1" ;
.

```

bricksrc/relationships.py
```diff
@@ -162,6 +162,48 @@
...
+        A: [OWL.ObjectProperty, OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+        OWL.inverseOf: BRICK["controls"],
+        "range": BRICK.Controller,
+        "domain": BRICK.Equipment,
+        RDFS.label: Literal("Is controlled by", lang="en"),
+    },
+    "hosts": {
+        A: [OWL.ObjectProperty, OWL.AsymmetricProperty, OWL.IrreflexiveProperty],
+        OWL.inverseOf: BRICK["isHostedBy"],
+        "range": BRICK.Point_Group,
```
jbkoh: Have we decided to avoid union of point group and point here?

gtfierro:
I'm trying to merge this in, but I'm still not 100% happy with the "concerns" relationship name. How do you all feel about `manages` for the Controller/Zone relationship?

fennibay:
@gtfierro thank you, it looks good. Matches what we discussed yesterday as the first step.

gtfierro:
One heads up; to get the tests to pass, I had to make `brick:hasPart` a subPropertyof `rec:includes`. Because Point Group is a rec:Collection, it *needs* at least 1 thing in the collection using `rec:includes` (this is the existing REC validation rule). By making hasPart a subproperty of includes, we get this automatically. @ektrah @jbkoh @fennibay this is what is necessary to use `brick:hasPart` instead of `rec:includes`

fennibay:
@jbkoh thanks for elaborating further.

> * Does a cloud server that is hosting all the copies of points hostsPoint the Points along with the source devices (e.g., a controller)? Are the copies of the Points considered different Points? Maybe, and that's why I was adding "represents the source" in my attempt of the definition.
> * If a Point originates from a certain device, but if the Point is only visible through a proxy, is the proxy hosting the Point or the origin device or both? Two representative examples are
>   * A thermostat is not directly exposed through the network but only through a VAV controller. Are Points of the thermostat hosted by the thermostat or the vav controller?
>   * Often, BMSes do not expose the field controllers but only as the BMS's own points. In that case, are the Points hosted by the BMS or the field controllers?

I think BMSs in general (including cloud) have quite some complexity. They have sometimes intermediaries/proxies to bring the point to their system, also there can be archiving systems (historians) and analytics systems that store copies of the points. Also, for BMS-local reasons, they might assign multiple IDs to the same point. I see the purpose of Brick+REC to make the automation system understandable. We don't need to understand how the BMS works. So I would leave all this out-of-scope.

I can support the thermostat example. For instance [QAP22](https://sid.siemens.com/v/u/9922) cannot be accessed in the network, only via a Controller; but [QMX6.P30](https://sid.siemens.com/v/u/A6V15997757) would be accessible if the client is connected to the network and speaks the KNX protocol, otherwise again accessed via controller.

> Maybe our homework is to precisely define what "hosting" actually mean. To me, I define it as following
> 
> > A physical device stores a Point and represents the Point's source.

You're absolutely right. I would remove the "stores" and "physical", because maybe it's not storing, but just making the Point accessible; and it might be a virtual device as well (like SW controller).

So my proposal would be: "Equipment hosts the Point, by making it available for access."

Options I see:

- Option 1: Define `hostsPoint` it only for Controller.
  - Rationale: for sensors and actuators either it is not hosted at all, or `hostsPoint` will be for the same Equipment and Point like `hasPoint`. So we avoid unnecessary triples.
- Option 2: Define `hostsPoint` for Equipment. All Equipments that make Points available for access could host it.
  - Rationale: make the hosting explicit for all types of Equipment.
- Option 3: Do no define `hostsPoint` in Brick. This can be addressed via different means.
  - Rationale: keep it simple

In Option 1 and 3, we need an additional hint on how to access the point.
- We use Web of Thing's hasInteractionAffordance to show the hosting relationship, which also allows us to define a Form for accessing it.
- 223 uses hasExternalReference
- Brick also allows BACnet-specific `isObjectOf`

The starting point for this (#721) was to host Collections as first step towards modelling Control Programs running on the Controller, hence we landed in Option 1. Now that it has evolved into hosting points, we could also wait for the eventual introduction of Functions.

gtfierro:
Just pushed the `Controller` -> `ICT Equipment` change for hostspoint/ishostedby. Do we want a similar change for the `controls`/`isControlledBy` relation?

@jbkoh I would love a better definition of controller! I just threw something in there

fennibay:
I would also keep controls on the Controller.

Many thanks for driving this forward!

bricksrc/equipment.py
```diff
@@ -57,6 +57,7 @@
             },
             "Controller": {
                 "tags": [TAG.Equipment, TAG.ICT, TAG.Controller],
+                BRICK.hostsPoint: [BRICK.Point],
```
fennibay: I think this should go to ICT_Equipment now,

Issue #726, opened 2025-10-02
Deprecated classes are in descriptions.csv


metesaka:
There are 107 deprecated classes with definitions in definitions.csv. We don't want to lose them but these should move to deprecations.py file.

PR #704, opened 2025-06-26
Descriptions of canonical and alias classes

metesaka:
Pull Request to Fix #699

This PR addresses missing and inconsistent class descriptions in the definitions.csv file:

Changes Made
- Removed all descriptions from "Non-canonical" classes.

- Added 8 "canonical" classes that were missing from the definitions.csv file, each with a corresponding definition.

- Populated descriptions for "canonical" classes that were previously missing them but have equivalent classes with valid descriptions.

- links are moved to canonical classes

Important Note:
- 116 "canonical" classes still lack descriptions in the CSV file. These classes either have no equivalent classes or their equivalents also lack descriptions.

- 274 "canonical" classes are still missing entirely from the CSV file.

Note: This fix only considers subclasses of brick:Entity, as outlined in [this document](https://docs.brickschema.org/brick/aliases.html).

PR #727, opened 2025-10-02
Deprecated classes definitions

metesaka:
This aims to solve #726, continuation of the work in #704, check that PR.

Current solution involves removing deprecated classes from definitions.csv, add to a new csv that generate brick code will handle.


generate_brick.py
```diff
@@ -822,6 +822,35 @@ def handle_deprecations():
...
             G.add((deprecated_term, BRICK.isReplacedBy, md["replace_with"]))
 
+    """
+    The following part adds definitions for deprecated Brick subclasses through SKOS.definitions.
+
+    This parses the definitions from ./bricksrc/definitions_of_deprecated.csv and
+    adds it to the graph. If available, adds the source information of
+    through RDFS.seeAlso.
+    """
+    with open("./bricksrc/definitions_of_deprecated.csv", encoding="utf-8") as dictionary_file:
```
gtfierro, line 832: This should use the python `pathlib` module to handle cross platform paths

generate_brick.py
```diff
@@ -822,6 +822,35 @@ def handle_deprecations():
...
+
+        # add definitions, citations to the graph
+        for definition in dictionary:
+            term = URIRef(definition[0])
+            if len(definition) > len(header):
+                raise ValueError(
+                    f"The term '{term}' has more elements than expected. Please check the format."
+                )
+            if len(definition[1]):
+                graph.add((term, SKOS.definition, Literal(definition[1], lang="en")))
```
gtfierro, line 845: I think this needs to use the global `G` variable to add the triples to the Brick ontology, but please double check

generate_brick.py
```diff
@@ -822,6 +822,35 @@ def handle_deprecations():
...
+            if len(definition) > 2 and definition[2]:
+                try:
+                    graph.add((term, RDFS.seeAlso, URIRef(definition[2])))
+                except Exception as e:
```
gtfierro, line 849: This should just raise the exception. Replace the `print` with a `raise Excaption`

gtfierro: What is the case where this would fail? What does `len(definition) > 2 and definition[2]` mean? Please describe the failure condition here

gtfierro (changes requested):
Thanks @metesaka ! I left a couple comments on the code. Let me know if you have any questions

gtfierro:
Thank you!

metesaka:
@gtfierro I noticed, you've changed the deprecated_definions.csv however, that csv is called in generate_brick, so now the build fails. 

I'll open a new PR to fix it and merge that

PR #735, opened 2025-11-11
Added Chilled Water Bypass and Isolation Valves and Plate Heat Exchanger

sappling:
Added three equipment classes:

- Chilled_Water_Bypass_Valve
- Chilled_Water_Isolation_Valve
- Plate_Heat_Exchanger


gtfierro:
Thanks @sappling ! This should be available in the next nightly release by tomorrow morning

PR #747, opened 2025-12-18
Adding automation groups

gtfierro:
Addressing #722


fennibay:
This is the current validation error:

```
   Severity: sh:Violation
  Focus Node: <urn:example/collection#Ahu>
  Message: Value must have at least one of the following shapes: _:7ed7414772658c5d3453cd82915382c7, _:e07496e981cc2a35110fc89f8f65c0eb
  Path: brick:hasPart
  Value: <urn:example/collection#FrostDetection>
  Source Constraint: sh:OrConstraintComponent
  Source Shape: _:ne1f5e65888c64b61a1d5168bc478d3ddb3
```

My interpretation is that we're now running against the following constraint

https://github.com/BrickSchema/Brick/blob/f7f0db97f1aecd998623c3aec2492cf3e91c16b3/bricksrc/root_class_shapes.ttl#L49-L54

But <urn:example/collection#FrostDetection> is of type brick:AutomationGroup. So I don't understand what is going wrong here.

Could it be that the SHACL engine is using a cached version or a different format of this ttl file?

fennibay:
@gtfierro thank you for preparing this.

I think this is now good for the equipment part. The second aspect in #722 was collections within Zones, and I'd like address that part, as well.

A typical example is how we structure our room application. The points are organized in folders because they have some shared context, e.g.:

Z: Zone:
- C: HVAC
  - E: Radiant ceiling
- C: Room temperature setpoint determination
  - P: Room temperature setpoint
  - P: Room temperature setpoint shift
- C: Air volume flow tracking
  - P: Room supply air volume flow
  - P: Present setpoint for room extract air volume flow
- C: Ventilation control
  - P: Present ventilation setpoint
  - P: Setpoint room air quality for comfort
  - P: Setpoint room air quality for pre-comfort
  - P: Setpoint room air quality for economy

(Z: Zone, C: Collection, P: Point, E: Equipment)

Ref: <https://sid.siemens.com/r/A6V11176325/6498277643__6.0_en-US_6419929099>

So, can we apply the same principle to rec:Zone or rec:Space where an Automation_Collection can be included within them?

/cc @ektrah

gtfierro:
I just implemented it, and added a quick example of your Room application: 

https://github.com/BrickSchema/Brick/blob/79fab96de19cdf0d0d442a85ca22f754d9b2ac4c/examples/collection/room_application.ttl

Let me know what you think!

examples/collection/room_application.ttl
```diff
@@ -0,0 +1,82 @@
...
+@prefix rdfs:  <http://www.w3.org/2000/01/rdf-schema#> .
+
+<http://example.com/room-application> a owl:Ontology ;
+    owl:imports <https://brickschema.org/schema/1.4/Brick> .
+
+bldg:Zone a rec:Zone ;
+  rec:includes bldg:Hvac .
+
+bldg:Room a rec:Space ;
+  rec:isPartOf bldg:Zone .
```
fennibay: This is not model-relevant at all, but for me: Room would be of type rec:Room and would include the zone and not vice-versa.

```suggestion
bldg:Room a rec:Room;
  rec:hasPart bldg:Zone .
```

bricksrc/equipment.py
```diff
@@ -99,7 +99,15 @@
...
         "tags": [TAG.HVAC, TAG.Equipment],
         "constraints": {
-            BRICK.hasPart: [BRICK.HVAC_Equipment, BRICK.Valve],
+            BRICK.hasPart: [
+                BRICK.HVAC_Equipment,
+                BRICK.Valve,
+            ],
+            REC.includes: [
+                BRICK.HVAC_Equipment,
+                BRICK.Valve,
```
fennibay: Do we need to allow includes between an equipment and its subparts? We already have hasPart for this.

For me it would be enough to just allow includes for Automation_Collection.

fennibay (changes requested):
Thank you @gtfierro, looks great for me. I have one important and one non-important comment.

bricksrc/root_class_shapes.ttl
```diff
@@ -86,6 +91,14 @@ brick:Equipment a sh:NodeShape ;
     ] ;
     .
 
+brick:Automation_Collection a sh:NodeShape ;
+    sh:property [
+        sh:path brick:hasPart ;
+        sh:maxCount 0 ;
+        sh:message "Automation Collections must use rec:includes instead of brick:hasPart."
+    ] ;
```
fennibay, line 99: I think this is colliding with the following rule: https://github.com/BrickSchema/Brick/blob/b61d72ab0b8af5ad6ff3de1320e16b2248375cb7/bricksrc/rules.ttl#L576-L598

That rule inserts a brick:hasPart relationship when there is a rec:includes relationship of a Point_Collection, and here we say that Automation_Collections may not have brick:hasPart relationships.

I would revisit rules.ttl. I think we should keep includes and hasPart separate.

bricksrc/equipment.py
```diff
@@ -99,7 +99,13 @@
     "HVAC_Equipment": {
         "tags": [TAG.HVAC, TAG.Equipment],
         "constraints": {
-            BRICK.hasPart: [BRICK.HVAC_Equipment, BRICK.Valve],
+            BRICK.hasPart: [
+                BRICK.HVAC_Equipment,
+                BRICK.Valve,
+            ],
+            REC.includes: [
+                BRICK.Automation_Collection,
```
fennibay, line 107: And this is colliding with <https://github.com/BrickSchema/Brick/blob/b61d72ab0b8af5ad6ff3de1320e16b2248375cb7/bricksrc/rules.ttl#L500-L526>

because that inserts rec:includes where there is a brick:hasPart and here we constrain rec:includes to brick:Automation_Collection only.

I would revisit rules.ttl

fennibay (commented):
Thanks for updating this. 

I put some analysis on the last validation errors.

bricksrc/equipment.py
gtfierro: I believe I fixed this! Thanks for the analysis

gtfierro:
@fennibay I amended the documentation for this feature at https://docs.brickschema.org/modeling/point-collections.html

Please take a look and let me know  if you have any feedback!

fennibay:
> @fennibay I amended the documentation for this feature at https://docs.brickschema.org/modeling/point-collections.html
> 
> Please take a look and let me know if you have any feedback!

Thank you @gtfierro, looks great to me. Hopefully we can discuss this today and merge it.

Issue #743, opened 2025-12-01
Entering/Leaving Chilled_Water_Temperature definitions.


sappling:
Currently these two classes are defined as:
Leaving_Chilled_Water_Temperature_Sensor : Measures the temperature of chilled water that is supplied from a chiller
Entering_Chilled_Water_Temperature_Sensor : Measures the temperature of chilled water that is **enteringed to a cooling tower**

I have several concerns and would like to get consensus before creating a pull request.

1. Entering_Chilled_Water_Temperature_Sensor is not the temperature of the water entering the cooling tower.  That would be condenser water.
2. I don't know that these definitions should refer to the chiller.  I think that perhaps they should be like Entering_Chilled_Water_Flow_Sensor that doesn't refer to the chiller, but just refers to "chilled entering water" (although I would prefer "entering chilled water").
3. How would these be used outside of a chiller?  If a water cooled AHU has a point measuring the temperature of it's incoming chilled water, it would seem to be the Entering_Chilled_Water_Temperature_Sensor, but by the above definitions, it should probably be the Leaving one since that is defined as what is supplied by the chiller.  I think that having the chiller as part of the definition is confusing.
4. Lastly - I don't think that there is a clear way of distinguishing the chilled water leaving a specific chiller from the chilled water leaving the whole chiller plant.  This was brought up in [Issue 628](https://github.com/BrickSchema/Brick/issues/628), but doesn't seem to be resolved.

I am proposing:
Leaving_Chilled_Water_Temperature_Sensor : Measures the temperature of chilled water that is leaving an equipment
Entering_Chilled_Water_Temperature_Sensor : Measures the temperature of chilled water that is entering an equipment

To me this would mean that the a point in an AHU measuring the temperature of it's incoming chilled water would be an Entering_Chilled_Water_Temperature_Sensor.

I'm not clear how to model the chilled water for the whole chiller plant.  Would this be a Leaving_Chilled_Water_Temperature_Sensor on a [Chilled_Water_System](https://brickschema.org/schema/Brick#Chilled_Water_System)?

gtfierro:
I'm commenting the Project Haystack model here (https://project-haystack.org/doc/docHaystack/chilled-water-plant.svg) so I can refer to it when I get to working on this. This suggests we want the substance (chiller vs condenser water) to encode which loop we are talking about

gtfierro:
[...]

Issue #677, opened 2024-10-30
Need for additional Point classes, as well as "un deprecation" of some.


PeteHart:
@gtfierro 

A number of large Swedish property owners have undertaken a joint effort to classify existing HVAC systems with Brick 1.4.1

To get consistency we've realized that some Point classes are missing and some fairly recent deprecations are needed.

The background for the requests are meticulously described in attached. In summary we suggest

Non existing classes to Add:
• Return_Air_Static_Pressure_Sensor
• Return_Air_Static_Pressure_Setpoint
• Return_Air_Flow_Setpoint
• Room_Air_Temperature_Sensor
• Zone_CO2_Level_Setpoint
• Zone_CO_Level_Sensor
• Zone_CO_Level_Setpoint
• Return_Air_CO_Setpoint
• Domestic_Hot_Water_Return_Temperature_Sensor
• Domestic_Hot_Water_Return_Temperature_Setpoint
• Water_Volume_Sensor

Deprecated classes to retain:
• Zone_Air_Temperature_Setpoint
• Hot_Water_Supply_Temperature_Sensor
• Supply_Hot_Water_Temperature_Setpoint
• Chilled_Water_Supply_Temperature_Sensor
• Supply_Chilled_Water_Temperature_Setpoint
• Hot_Water_Return_ Temperature_sensor
• Return_Hot_Water_Temperature_Setpoint
• Chilled_Water_Return_Temperature_Sensor
• Return_Chilled_Water_Temperature_Setpoint
• Hot_Water_Supply_Flow_Sensor
• Hot_Water_Return_Flow_Sensor
• Chilled_Water_Supply_Flow_Sensor
• Chilled_Water_Return_Flow_Sensor
• Domestic_Hot_Water_Supply_Temperature_Sensor
• Domestic_Hot_Water_Supply_Temperature_Setpoint

Additional requests:
• Add system: Domestic_Water_System or Domestic_Cold_Water_System
• Rename the following classes as follows:
o Hot_Water_Supply_Temperature_Sensor >> Supply_Hot_Water_Temperature_Sensor
o Chilled_Water_Supply_Temperature_Sensor >> Supply_Chilled_Water_Temperature_Sensor
o Hot_Water_Return_ Temperature_sensor >> Return_Hot_Water_Temperature_Sensor
o Chilled_Water_Return_Temperature_Sensor >> Return_Chilled_Water_Temperature_Sensor
[...]

PR #752, opened 2026-01-01
Fix issue 743, adding chilled water system points back

gtfierro:
Closing #677, #743 at least


gtfierro:
- make sure the definitions of substances are correct, and not overly specific on what kind of equipment they belong to. Remember: the modeling choice we're making is that the substance encodes which "loop" you are in, e.g. the condenser water loop, without saying if you're on the condenser or on the chiller -- that comes from hasPoint.
- double check the point rdfs:label in the example. Produce a picture.

PR #765, opened 2026-03-21
csv name fixed

metesaka:
This is to fix a small complication of #727.

The new csv file in bricksrc doesn't match the open statement in generate_brick.py


gtfierro:
Thanks!

PR #793, opened 2026-07-21
Add building types from Energy Star, OmniClass, and BREEAM USA

### LINKED ISSUE #784: Expanding on Building Types
Chatting with @gtfierro and @ektrah yesterday, there is a need to expand on the kinds of building types that are supported within Brick. Currently, we rely on Real Estate Core (REC) under https://ontology.brickschema.org/1.4/rec/Building.html . Naturally there are many other building types so one of the challenges is finding an appropriate limit. One reference point in our HVAC and plumbing market in the US is Energy Star. Energy Star defines a set of property types under https://www.energystar.gov/buildings/benchmark/understand-metrics/property-types . All `Property Types` have definitions in their [glossary](https://portfoliomanager.energystar.gov/pm/glossary).

I don't think Brick necessary *needs* to have the same hierarchy, but they are good placeholders/categories when a property type does not neatly fall into one of the child classes. Further, this is going to be a matter of proposing these classes to REC, afterwhich would get referenced in Brick.

There are a few places where we can find inspiration. A quick Claude prompt says to look to BREEAM and CSI's OmniClass for inspiration.

Here are a couple tables that reference BREEAM USA, OmniClass, and Energy Star.

[Building Types.xlsx](https://github.com/user-attachments/files/28522219/Building.Types.xlsx)

gtfierro:
Adds 171 rec:Building subclasses covering the full Energy Star Portfolio Manager property-type taxonomy plus additional types from CSI OmniClass Table 11 and BREEAM USA that Energy Star doesn't cover on its own, each with a definition, a source link where available, and cross-references to the corresponding term in the other classification systems.

Reuses existing REC classes (Hospital, School, Stadium, ShoppingMall) where an exact match exists, and renames a few new classes (OfficeBuilding, LibraryBuilding, LaboratoryBuilding, ParkingFacility) to avoid colliding with existing Room/Zone-level classes of the same name.

These are defined in bricksrc/recpatches.ttl as temporary local additions, following the existing pattern for concepts not yet proposed to REC.

### DIFF
--- a/bricksrc/recpatches.ttl
+++ b/bricksrc/recpatches.ttl
@@ -11,6 +11,7 @@
 @prefix sh: <http://www.w3.org/ns/shacl#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 @prefix bsh: <https://brickschema.org/schema/BrickShape#> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
 
 # allowing brick:feeds to be used to connect Spaces to "feed" Equipment, e.g. for the return/discharge path
 rec:Space sh:property [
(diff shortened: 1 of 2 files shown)


gtfierro:
~~One note: consider making these annotations so we can apply uses to architecture instead of specifically building.~~

Keep these as subclasses for consistency with existing REC building types. Edit/clean up the definitions.

PR #797, opened 2026-08-11
Add Brick → bSDD dictionary export

gtfierro:
Adds an offline export of Brick as a [buildingSMART Data Dictionary](https://technical.buildingsmart.org/services/bsdd/) import document.

The exporter reads the built `Brick+imports.ttl` and writes one generated JSON
dictionary. The output is never hand-edited.

```bash
make Brick.ttl
uv run python tools/bsdd/generate_bsdd.py
uv run pytest tests/test_bsdd_export.py
```

The default Brick build is unchanged; `make bsdd` remains a separate target and
the generated `bsdd/` directory is ignored.

## Mapping scope

- **Brick classes** become bSDD Classes. Labels and definitions come from the
  built graph. bSDD permits one `ParentClassCode`; the most-specific Brick
  parent is used in the tree and additional parents are emitted as `IsChildOf`
  relations.
- **Quantity kinds** become numeric bSDD Properties, with units and dimensions
  derived from QUDT and translated through the checked-in bSDD unit map. Brick
  quantities that explicitly reference a QUDT quantity kind use that quantity
  kind as their canonical property.
- **EntityProperties** become `entityProperty<Code>` Properties, keeping them
  distinct from physical quantity kinds. For example, `brick:volume` maps to
  `entityPropertyVolume`, while the physical quantity is `Volume`.
- **Structured EntityProperties** become bSDD `Complex` Properties whose scalar
  children use `parent.field` codes, such as
  `entityPropertyCoordinates.latitude`.
- **ClassProperties** are emitted from `brick:hasQuantity`,
  `brick:hasSubstance`, and EntityProperty SHACL attachments.
- **IFC mappings** come directly from `tools/bsdd/mappings/ifc.csv`. Each row
  maps one exact Brick class code to `RelatedIfcEntityNamesList`; mappings do
  not inherit through the class hierarchy.

## Deliberate boundaries

- Only live, Brick-owned classes are published. Imported vocabularies,
  including RealEstateCore location classes, are not republished under Brick
  URIs.
- The export has no network dependency. `mappings/units.csv` and `ifc.csv` are
  checked-in, reviewed inputs.
- The exporter does not map general Brick relationships such as `feeds`,
  `hasPoint`, or `controls`, because they do not have an equivalent bSDD class
  relation.
- Brick substances are emitted as allowed values of an observation property;
  they are not bSDD materials.

The dictionary is emitted with `Status: Preview`. Publishing and activation are
deliberate manual steps outside this PR.


gtfierro:
[brick-bsdd.json](https://github.com/user-attachments/files/31278634/brick-bsdd.json)

Here is the current JSON export

PR #800, opened 2026-09-09
Move deprecated class definitions from CSV into deprecations.py

metesaka:
Definitions (and rdfs:seeAlso) for deprecated terms now live in bricksrc/deprecations.py: inline SKOS.definition for entries in the `deprecations` dict, and a new `deprecated_definitions` dict for terms whose deprecation is declared in bricksrc/recpatches.ttl. generate_brick.py no longer reads bricksrc/deprecated_definitions.csv, which is removed from the repository.


metesaka:
@gtfierro I didn't want to touch recpatches.ttl so removed them into new dicts in deprecations.py. I think this might create issues as now there's two sources to maintain related to deprecations. If you don't like this approach, I can instead remove these definitions directly into recpatches.ttl, or move all the deprecated classes from recpatches.ttl to deprecations.py. Let me know if you prefer either of these choices!

gtfierro (approved):
Thanks!

PR #803, opened 2026-09-10
Make sure our collection classes are subclass of brick entity

gtfierro:
They fell out of the Brick tree at some point when we deprecated Brick:collection. This re-adds them so they play nicer with the ontology documentation.


ektrah:
Would it make sense to revive `brick:Collection` to the extent that e.g. `brick:System rdfs:subClassOf brick:Collection` and `brick:Collection rdfs:subClassOf brick:Entity` as it was in Brick 1.4.4 (i.e., just mark `brick:Collection` as deprecated without removing it from the subclass hierarchy)?

gtfierro:
Great idea! It ended up being pretty clean to do this.

